PR #5, opened 2016-10-22
TypeScript Code generation

rricard:
This is a work in progress targeting TS type annotation generation. For now, only basic structures will be generated.

### DIFF
--- a/test/typescript/codeGeneration.js
+++ b/test/typescript/codeGeneration.js
@@ -75,7 +75,47 @@ describe('TypeScript code generation', function() {
         export interface HeroNameQuery {
           hero: {
             name: string,
-          };
+          } | null;
+        }
+      `);
+    });
+
+    it(`should generate simple nested query operations including input variables`, function() {
+      const context = this.compileFromSource(`
+        query HeroAndFriendsNames($episode: Episode) {
+          hero(episode: $episode) {
+            name
+            friends {
+              name
+            }
+          }
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
+
+        // The episodes in the Star Wars trilogy
+        export type Episode =
+          "NEWHOPE" | // Star Wars Episode IV: A New Hope, released in 1977.
+          "EMPIRE" | // Star Wars Episode V: The Empire Strikes Back, released in 1980.
+          "JEDI"; // Star Wars Episode VI: Return of the Jedi, released in 1983.
+
+
+        export interface HeroAndFriendsNamesQueryVariables {
+          episode: Episode | null;
+        }
+
+        export interface HeroAndFriendsNamesQuery {
+          hero: {
+            name: string,
+
+            friends: Array< {
+              name: string,
+            } > | null,
+          } | null;
         }
       `);
     });
--- a/test/typescript/codeGeneration.js
+++ b/test/typescript/codeGeneration.js
@@ -47,6 +47,28 @@ describe('TypeScript code generation', function() {
   });
 
   describe('#generateSource()', function() {
+    it(`should generate simple query operations`, function() {
+      const context = this.compileFromSource(`
+        query HeroName {
+          hero {
+            name
+          }
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
+
+        export interface HeroNameQuery {
+          hero: {
+            name: string,
+          } | null;
+        }
+      `);
+    });
+
     it(`should generate simple query operations including input variables`, function() {
       const context = this.compileFromSource(`
         query HeroName($episode: Episode) {
--- a/test/typescript/codeGeneration.js
+++ b/test/typescript/codeGeneration.js
@@ -215,7 +215,7 @@ describe('TypeScript code generation', function() {
 
     it(`should generate mutation operations with complex input types`, function() {
       const context = this.compileFromSource(`
-        mutation ReviewMovie($episode: Episode, $review: Review) {
+        mutation ReviewMovie($episode: Episode, $review: ReviewInput) {
           createReview(episode: $episode, review: $review) {
             stars
             commentary
@@ -228,21 +228,38 @@ describe('TypeScript code generation', function() {
       expect(source).to.include(stripIndent`
         //  This file was automatically generated and should not be edited.
 
+        // The episodes in the Star Wars trilogy
+        export type Episode =
+          "NEWHOPE" | // Star Wars Episode IV: A New Hope, released in 1977.
+          "EMPIRE" | // Star Wars Episode V: The Empire Strikes Back, released in 1980.
+          "JEDI"; // Star Wars Episode VI: Return of the Jedi, released in 1983.
+
+
         export interface ReviewInput {
+          // 0-5 stars
           stars: number;
+          // Comment about the movie, optional
           commentary: string | null;
-          favoriteColor: {
-            red: number,
-            green: number,
-            blue: number,
-          } | null;
+          // Favorite color, optional
+          favoriteColor: ColorInput | null;
+        }
+
+        export interface ColorInput {
+          red: number;
+          green: number;
+          blue: number;
+        }
+
+        export interface ReviewMovieMutationVariables {
+          episode: Episode | null;
+          review: ReviewInput | null;
         }
 
         export interface ReviewMovieMutation {
           createReview: {
             stars: number,
             commentary: string | null,
-          };
+          } | null;
         }
       `);
     });
--- a/test/typescript/codeGeneration.js
+++ b/test/typescript/codeGeneration.js
@@ -1,0 +1,83 @@
+import { expect } from 'chai';
+
+import { stripIndent } from 'common-tags';
+
+import {
+  parse,
+  isType,
+  GraphQLID,
+  GraphQLString,
+  GraphQLInt,
+  GraphQLList,
+  GraphQLNonNull
+} from 'graphql';
+
+import {
+  generateSource
+} from '../../src/typescript/codeGeneration';
+
+import { loadSchema } from '../../src/loading';
+const schema = loadSchema(require.resolve('../starwars/schema.json'));
+
+import CodeGenerator from '../../src/utilities/CodeGenerator';
+
+import { compileToIR, printIR } from '../../src/compilation';
+
+describe('TypeScript code generation', function() {
+  beforeEach(function() {
+    const context = {
+      schema: schema,
+      operations: {},
+      fragments: {},
+      typesUsed: {}
+    }
+
+    this.generator = new CodeGenerator(context);
+
+    this.compileFromSource = (source) => {
+      const document = parse(source);
+      const context = compileToIR(schema, document);
+      this.generator.context = context;
+      return context;
+    };
+
+    this.addFragment = (fragment) => {
+      this.generator.context.fragments[fragment.fragmentName] = fragment;
+    };
+  });
+
+  describe('#generateSource()', function() {
+    it(`should generate simple query operations including input variables`, function() {
+      const context = this.compileFromSource(`
+        query HeroName($episode: Episode) {
+          hero(episode: $episode) {
+            name
+          }
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
+
+        // The episodes in the Star Wars trilogy
+        export type Episode =
+          "NEWHOPE", // Star Wars Episode IV: A New Hope, released in 1977.
+          "EMPIRE", // Star Wars Episode V: The Empire Strikes Back, released in 1980.
+          "JEDI"; // Star Wars Episode VI: Return of the Jedi, released in 1983.
+
+
+        export interface HeroNameQueryVariables {
+          episode: Episode | null;
+        }
+
+        export interface HeroNameQuery {
+          hero: {
+            name: string,
+          };
+        }
+      `);
+    });
+  });
+});
--- a/test/typescript/codeGeneration.js
+++ b/test/typescript/codeGeneration.js
@@ -133,7 +133,6 @@ describe('TypeScript code generation', function() {
         export interface HeroAndFriendsNamesQuery {
           hero: {
             name: string,
-
             friends: Array< {
               name: string,
             } > | null,
@@ -173,6 +172,43 @@ describe('TypeScript code generation', function() {
           friends: Array< {
             name: string,
           } > | null;
+        }
+      `);
+    });
+
+    it(`should generate query operations with inline fragments`, function() {
+      const context = this.compileFromSource(`
+        query HeroAndDetails {
+          hero {
+            name
+            ...HeroDetails
+          }
+        }
+
+        fragment HeroDetails on Character {
+          ... on Droid {
+            primaryFunction
+          }
+          ... on Human {
+            height
+          }
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
+
+        export interface HeroAndDetailsQuery {
+          hero: HeroDetailsFragment & {
+            name: string,
+          } | null;
+        }
+
+        export interface HeroDetailsFragment {
+          primaryFunction: string | null;
+          height: number | null;
         }
       `);
     });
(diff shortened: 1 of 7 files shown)


stubailo:
I think it will be easier to see what it generates with some tests!

rricard:
@stubailo I intend to write a lot of them, I'm just starting on this one. I just opened the PR in case you want to provide some early feedback.

stubailo:
Yeah, my main interest would be to see some example TS code generated from this approach, and one way would have been to look at some tests! But I can wait :]

stubailo:
Yeah, I think for the iOS generation we took a different approach which was to design the generated result first, then implement the generation. But either way is fine!

rricard:
@stubailo I'm trying to restart with the implementation structure for swift actually. What helps is that I clearly know what I want to get but I should formalize it indeed. I'll try to do that before continuing.

rricard:
@stubailo found a basic query passing with the generator I currently have 😉 : https://github.com/apollostack/apollo-codegen/blob/a9ef9c1d1b351701426209edb17d215848194cb8/test/typescript/codeGeneration.js#L62:L79

rricard:
and ... actually writing it in a test makes me realize it's not valid TS 😉. Thank you very much for the advice! Will continue on TDD now, I usually do that but I don't know why, I did it freestyle this time!

martijnwalraven:
@rricard: Seems like you're on a roll! Let me know if you have any questions about the codebase, or if anything needs to change about the compiler IR output.

rricard:
here is what I'll have for today: https://github.com/apollostack/apollo-codegen/blob/8bda617e7b69ffecfc2cf716cba7247260f7a958/test/typescript/codeGeneration.js

rricard:
@martijnwalraven seems good for now, I'll definitely need to refactor at some point, the swift code structure was a good start but TS doesn't need that much stuff (I'm just generating type annotations) so I'll try to see how to make things lighter ... But, yea, the generator is definitely a good idea, especially for indents and blocks!

martijnwalraven:
Good to hear! I'll probably try to split up the Swift `codeGeneration.js` at some point as well, because it is indeed getting a little too heavy.

rricard:
Fearing to do the same in TS...

rricard:
@stubailo would love some feedback on what I'm generating (see tests 🎉 !). Once you feel it's what we would expect from this generator, I'll refactor it!

rricard:
I need to test that against the kind of complex queries I have in my app and see if that doesn't break (we have TONS of fragments!)

test/typescript/codeGeneration.js
```diff
@@ -0,0 +1,216 @@
...
+          hero {
+            name
+          }
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
```
stubailo: I think it would be best to write tests so that they don't all have to hardcode this boilerplate at the top.

rricard:
Ok, just tested excepted small edge cases, I think this works pretty well!

Here is what I see that is left to cover:

- [ ] Nullable arrays of nullable elements (for now, doing something ugly such as: `Item | null[] | null`, should do: `(Item | null)[] | null`)
- [ ] Input object fields (check if they work correctly)

test/typescript/codeGeneration.js
```diff
@@ -0,0 +1,216 @@
...
+
+        // The episodes in the Star Wars trilogy
+        export type Episode =
```
stubailo: If we compile multiple queries with this enum, do we generate the type multiple times, or just once? Sorry, I'm not really in depth on the codegen, although I'm trying to learn more about it!

test/typescript/codeGeneration.js
```diff
@@ -0,0 +1,216 @@
...
+        export interface HeroAndFriendsNamesQueryVariables {
+          episode: Episode | null;
+        }
+
+        export interface HeroAndFriendsNamesQuery {
+          hero: {
+            name: string,
+            friends: Array< {
+              name: string,
+            } > | null,
```
stubailo: Personally I'd prefer `{ ... }[]` over `Array<{ ... }>`, just because of the JSX thing.

test/typescript/codeGeneration.js
```diff
@@ -0,0 +1,216 @@
...
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
+
+        export interface HeroAndFriendsNamesQuery {
+          hero: HeroFriendsFragment & {
```
stubailo: Wow, TypeScript is great!

test/typescript/codeGeneration.js
```diff
@@ -0,0 +1,216 @@
...
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
+
+        export interface HeroAndDetailsQuery {
+          hero: HeroDetailsFragment & {
+            name: string,
+          } | null;
+        }
+
+        export interface HeroDetailsFragment {
```
stubailo: Would it be more accurate to do something like:

```ts
{ primaryFunction: string | null } | { height: number | null }
```

So that it encodes that it's exactly one of the two combinations?

stubailo:
@jbaxleyiii take a look at this as well, it could be a good base for your Flow Type generator!

rricard:
@stubailo & @jbaxleyiii I actually need a flow generator as well and I plan to make this one run either in flow or ts mode once I hit the next PR 😉 . Since both syntaxes are similar, I think we could exploit the same thing to generate them!

test/typescript/codeGeneration.js
rricard: This concern is actually handled "generally" in the common parts of the codegen system. @martijnwalraven may be more able to tell you how he did that but in my codebase for instance I'm reusing the same enum type at multiple places and it's generated on top only once!

test/typescript/codeGeneration.js
rricard: @stubailo actually cheating here to get something closer to what I would do in flow ...

stubailo: That's a fine justification!

test/typescript/codeGeneration.js
stubailo: Sounds great to me! I was suspecting Martijn's code already did the right thing ;]

test/typescript/codeGeneration.js
```diff
@@ -0,0 +1,216 @@
...
+          }
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
+
+        // The episodes in the Star Wars trilogy
```
rricard: @stubailo interesting, I would love to get doc comments as well ... Do you have any doc on that subject so I can add it correctly ?

test/typescript/codeGeneration.js
rricard: @martijnwalraven I'm amazed on the amount of stuff I don't have to do! Thank you, that's some quality code you wrote!

test/typescript/codeGeneration.js
rricard: agreed, will try to see that during the refactor!

test/typescript/codeGeneration.js
stubailo: OK, I just tested it out. It looks like the comment has to be of the form `/** content */` for VSCode to recognize it and display as a tooltip.

stubailo: It even supports markdown!!

```ts
/**
 * Generate an executable schema
 *
 * - This
 * - is
 * - A list
 */
function makeExecutableSchema({
  typeDefs,
  resolvers = {},
  connectors,
  logger,
  allowUndefinedInResolve = true,
  resolverValidationOptions = {},
}: IExecutableSchemaDefinition) {
```

rricard: gotcha! will do that as much as I can, I think I'll do that on fields as well. This is super useful!

rricard: @stubailo GraphQL descriptions are actual markdown: this will be fun! I just need to create an helper to create multiline doc comments!

test/typescript/codeGeneration.js
stubailo: Yeah, I don't think it's critical. Although I think Martijn is doing something like this in the iOS code, so maybe there is an easy way?

rricard: I'll keep that in mind while refactoring, I still need to get more familiar with all of this!

test/typescript/codeGeneration.js
rricard: Urrgh, doesn't seem to work with attributes... Feels like only functions/methods will work...

stubailo: Dang. Well, we can iterate on it later.

pspeter3:
One question I have is whether or not `null` will be explicitly sent back from the server. If not, the conditional property operator may be better.

rricard:
@pspeter3 usually I see that the server sends a null back, that's how the graphql execution engine works in the spec at least

rricard:
@pspeter3 btw, if you want to work on tests, you should fork the repo and derive your own branch from this one (basically, you will end up doing a PR into this PR :wink:), that way, I'll be able to see through the diff more easily and guide you more easily!

rricard:
Here is a list I rapidly put together of what's left to do:

- more tests on working features (take examples from the swift tests)
- tests on non-working features (you will have to figure out the syntax we want - so don't hesitate to ask me, either here or on slack - my username is @rricard)
- I may work on making those tests pass but if you find yourself a lot of time to contribute, you can do it as well!

Any question: ping me here or on slack. Just keep in mind that I'm in Europe so, well, I may sleep while you are working!

pspeter3:
I have my own fork already and was planning to do the PR of a PR. Also one question, why decided to do type aliases over interfaces? Was it to intersect fragments? My belief was that type aliases don't show up in compilation errors so we may have a better developer experience if we use interfaces.

rricard:
@pspeter3 most of my decisions are based on doing something similar to flow (I am using that primarily, but I wanted to do TS first to see later how much I could centralize under the same syntax actually). So, in the test, do what you feel is the more logical for TS and I'll figure out how to make that common with flow and TS later. My first objective will be to make your tests pass. So, please write some idiomatic TS, and I'll work around that!

rricard:
That's actually great to have you give your input on this. I think we could end up with a better implementation in the end!

dotansimha:
Tried some use cases for the generator, it looks good, but there are some issues I noticed:

- When using custom inline fragment on an object that of type List (defined as `[MyObject]`) in the scheme definition, the generated type does not handle the fragment as array. 
This is a unit test that fails, to demonstrate the issue:
```
    it(`should generate correct list with custom fragment`, function() {
      const context = this.compileFromSource(`
        fragment Friend on Character {
          name
        }
        
        query HeroAndFriendsNames($episode: Episode) {
          hero(episode: $episode) {
            name
            friends {
              ...Friend
            }
          }
        }
      `);

      const source = generateSource(context);

      expect(source).to.include(stripIndent`
        
        // The episodes in the Star Wars trilogy
        export type Episode =
          "NEWHOPE" | // Star Wars Episode IV: A New Hope, released in 1977.
          "EMPIRE" | // Star Wars Episode V: The Empire Strikes Back, released in 1980.
          "JEDI"; // Star Wars Episode VI: Return of the Jedi, released in 1983.
        
        
        export interface HeroAndFriendsNamesQueryVariables {
          episode: Episode | null;
        }
        
        export interface HeroAndFriendsNamesQuery {
          hero: {
            name: string,
            friends: Array<FriendFragment>,
          } | null;
        }
        
        export interface FriendFragment {
          name: string;
        }
      `);
    });
```

rricard:
@dotansimha that's exactly the kind of feedback I needed! thanks! I'm back working on this this week! 😉

dotansimha:
Found a solution, need to replace the following line:
```ts
https://github.com/apollostack/apollo-codegen/blob/e32dd8983ec8733f76c7843ea66c26f8ba1ecc69/src/typescript/language.js#L38
```

to:
```ts
    generator.printOnNewline(`${propertyName}: ${isArray ? 'Array<' : ''}${fragmentSpreads.map(n => `${n}Fragment`).join(' & ')}${isArray ? '>' : ''}`);
```

Will create a PR into this PR in a few minutes.

@rricard , any chance to add a TODO list or a list of known issues to this PR? :)

rricard:
@dotansimha Nice! Please add a test for your case if you write a PR. I don't have any TODO right now but you are right it would be useful, I'll start by doing that this afternoon!

rricard:
Okay nice, let's finish this PR then! Just going to eat something and I'm back at work!

dotansimha:
Also noticed that the generator changed the name of the fields (and it should not do it in any case - it's fine for the name of the operation or interface, but not for the field name).
For example, a field named `full_name` will generate a type with a field called `fullName`.

dotansimha:
Also, @rricard , the concept of generating one big interface is a mistake. 

The actual usage will be something like:
```ts
    this.subscriptionSub = this.apollo.subscribe({
      query: subscriptionQuery,
      variables: {repoFullName: repoName},
    }).subscribe({
      next: (data: OnCommentAddedSubscription) => {
        const newComment: Comment = data.commentAdded;
        // ... more code ...
      }
    });
```

`OnCommentAddedSubscription` is generated, and looks like that:
```ts
export interface OnCommentAddedSubscription {
  commentAdded: {
    id: number,
    postedBy: {
      login: string,
      html_url: string,
    },
    createdAt: number,
    content: string,
  } | null;
}
```

The problem is that `newComment` it still remains as `any` because we can't use the inner type of `OnCommentAddedSubscription` as interface, the actual generated should looks like that:

```ts
export interface OnCommentAddedSubscriptionCommentAdded {
{
    id: number,
    postedBy: {
      login: string,
      html_url: string,
    },
    createdAt: number,
    content: string,
  }
}

export interface OnCommentAddedSubscription {
  commentAdded: OnCommentAddedSubscriptionCommentAdded | null;
}
```

So when you want to use an inner type in your code - it's an easy usage. 

The level of generated code can depend on the implementation, but the first level must get it's own interface so you will be able to extract inner types from the GraphQL response.

rricard:
@dotansimha I prefer the module notation. For now, I don't think we should change this. I want to ship that as fast as possible and actually derivate this implementation for Flow ASAP (which in its current state won't take much time).

In your example, I don't see the issue though, just let TS infer your type right? Or did I spend too much time with Flow and its inference engine that I think it's a completely normal thing?

Anyway, I see the point of what you want to do, maybe just let's do that on the next iteration for TS support!

rricard:
I think a "first version" is quite ready now, thanks to you (for handling those edge cases that I missed), I'll let @stubailo or @martijnwalraven give there feedback and then, let's get this merged.

Thanks for your help, ping me on slack to discuss what's next on this (same as on github: @rricard)!

rricard:
@dotansimha ping me on slack if you see anything else that seems important to you!

I'll work on refactoring that PR a bit once we agree it covers a "minimal" TS support...

dotansimha:
TS with WebStorm does not auto complete if you don't declare the type, so when using an inner property of an object - you must specify it's type, otherwise it treated as `any` and the auto-complete for fields does not work :(

rricard:
Humm, well, I'll see what I can do during the refactor then, do you see anything else we missed before I start?

pspeter3:
In general, we should prefer `namespace` over `module`.

rricard:
Okay, I added input object types. I will followup tomorrow with a proper refactor!

rricard:
@martijnwalraven & @stubailo maybe we should merge that now, the feature is properly isolated and this would allow the following:

- other community contributors will be able to propose PRs against master
- I'll be able to start the flow implementation (and refactor things while doing it)
- I think that proposing the feature to try as early as possible would help us spot bugs or things wanted by the users more easily

rricard:
(in fact starting with flow ASAP may be beneficial since it would push me faster towards refactor)

martijnwalraven:
@rricard: Happy to merge this, looks great!

stubailo:
Wow, excited to see so much progress today on this!

rricard:
@stubailo you can thank @dotansimha that spotted things I couldn't think of! Hope he'll stay around to help us get a better TS support!

martijnwalraven:
@rricard: I think it would be great if you could write an article about this for the Apollo blog!

rricard:
@martijnwalraven Good idea! I need to setup a sample project first though to be able to try out things correctly!

csillag:
+1

rricard:
@martijnwalraven can we have a release on npm with typescript support so I can use it in a sample repo? Thanks!

martijnwalraven:
@rricard: Just published apollo-codegen 0.9.5 on npm

rricard:
@martijnwalraven :+1: eating something and then I'm back on it!

PR #9, opened 2016-10-31
keep the data field when storing and loading the schema

### LINKED ISSUE #8: Keep top-level data field when storing the schema.json
Most (if not all) of the `schema.json` files provided online (such as [Relay](https://github.com/facebook/relay/blob/master/scripts/jest/testschema.json) for example) have the `data` field as the top-level field.

When downloading a schema through an introspection query using apollo-codegen, it only stores the `__schema` object in the json file.

IMO, keeping the data field should provide more flexibility for users, particularly in cases when someone doesn't go through the `download-schema` process.

This would also require minor modification to the checks in `loading.js` to support both cases.

I can push a fix if that's of interest.

marwanad:
Simply, modifies downloadSchema to keep the `data` field when storing the schema file to disk. An error is now thrown if `data` isn't the top-level field in the response.

As far as loading the schema for code-generation goes, an error is thrown when neither `data` nor `schema` exist as the top-level field. The reasoning is so as not break the old code-gen behaviour but I can modify that to only check for `data`.

Closes #8.

### DIFF
--- a/src/downloadSchema.js
+++ b/src/downloadSchema.js
@@ -37,8 +37,8 @@ export default async function downloadSchema(url, outputPath, additionalHeaders)
     throw new ToolError(`Errors in introspection query result: ${result.errors}`);
   }
 
-  const schemaData = result.data;
-  if (!schemaData) {
+  const schemaData = result;
+  if (!schemaData.data) {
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Thanks!

PR #16, opened 2016-12-02
Add empty public inits for Swift classes 

chrisvis:
These are needed for access from outside of Swift module


chrisvis:
Fixes for the tests are coming up

martijnwalraven:
Great, thanks!

PR #18, opened 2016-12-05
[TypeScript] Fixed bug with camelCased propertyName

dotansimha:
@rricard 

Can't add quick unit tests because the star wars schema does not have non-camelcased properties that I can add quick test for. 
Verified it against the GitHunt schema and it works.

https://github.com/apollostack/apollo-codegen/pull/5


rricard:
Yep, ok, completely makes sense, nice catch!

PR #24, opened 2016-12-10
Basic flow support

### LINKED ISSUE #2: Flow codegen
We could generate Flow types for the GraphQL queries in the project in parallel of something like [graphql-document-collector](https://github.com/rricard/graphql-document-collector) to get the query docs globally.

I'll work on this in a folder parallel to the `swift/` folder.

rricard:
Mainly copy-pasta from typescript, only the nullables changes...

Fixes #2

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -73,7 +73,7 @@ yargs
       target: {
         demand: false,
         describe: 'Code generation target language',
-        choices: ['swift', 'json', 'ts', 'typescript', 'flow', 'flowtype'],
+        choices: ['swift', 'json', 'ts', 'typescript', 'flow'],
         default: 'swift'
       },
       "passthrough-custom-scalars": {
--- a/src/cli.js
+++ b/src/cli.js
@@ -73,7 +73,7 @@ yargs
       target: {
         demand: false,
         describe: 'Code generation target language',
-        choices: ['swift', 'json', 'ts', 'typescript'],
+        choices: ['swift', 'json', 'ts', 'typescript', 'flow', 'flowtype'],
         default: 'swift'
       },
       "passthrough-custom-scalars": {
--- a/test/flow/codeGeneration.js
+++ b/test/flow/codeGeneration.js
@@ -59,6 +59,7 @@ describe('Flow code generation', function() {
       const source = generateSource(context);
 
       expect(source).to.include(stripIndent`
+        /* @flow */
         //  This file was automatically generated and should not be edited.
 
         export interface HeroNameQuery {
@@ -81,6 +82,7 @@ describe('Flow code generation', function() {
       const source = generateSource(context);
 
       expect(source).to.include(stripIndent`
+        /* @flow */
         //  This file was automatically generated and should not be edited.
 
         // The episodes in the Star Wars trilogy
@@ -117,6 +119,7 @@ describe('Flow code generation', function() {
       const source = generateSource(context);
 
       expect(source).to.include(stripIndent`
+        /* @flow */
         //  This file was automatically generated and should not be edited.
 
         // The episodes in the Star Wars trilogy
@@ -160,6 +163,7 @@ describe('Flow code generation', function() {
       const source = generateSource(context);
 
       expect(source).to.include(stripIndent`
+        /* @flow */
         //  This file was automatically generated and should not be edited.
 
         export interface HeroAndFriendsNamesQuery {
@@ -198,6 +202,7 @@ describe('Flow code generation', function() {
       const source = generateSource(context);
 
       expect(source).to.include(stripIndent`
+        /* @flow */
         //  This file was automatically generated and should not be edited.
 
         export interface HeroAndDetailsQuery {
@@ -226,6 +231,7 @@ describe('Flow code generation', function() {
       const source = generateSource(context);
 
       expect(source).to.include(stripIndent`
+        /* @flow */
         //  This file was automatically generated and should not be edited.
 
         // The episodes in the Star Wars trilogy
@@ -283,6 +289,7 @@ describe('Flow code generation', function() {
       const source = generateSource(context);
 
       expect(source).to.include(stripIndent`
+        /* @flow */
         //  This file was automatically generated and should not be edited.
         
         // The episodes in the Star Wars trilogy
--- a/test/flow/codeGeneration.js
+++ b/test/flow/codeGeneration.js
@@ -62,10 +62,10 @@ describe('Flow code generation', function() {
         /* @flow */
         //  This file was automatically generated and should not be edited.
 
-        export interface HeroNameQuery {
+        export type HeroNameQuery = {
           hero: ? {
             name: string,
-          };
+          },
         }
       `);
     });
@@ -92,14 +92,14 @@ describe('Flow code generation', function() {
           "JEDI"; // Star Wars Episode VI: Return of the Jedi, released in 1983.
 
 
-        export interface HeroNameQueryVariables {
-          episode: ?Episode;
+        export type HeroNameQueryVariables = {
+          episode: ?Episode,
         }
 
-        export interface HeroNameQuery {
+        export type HeroNameQuery = {
           hero: ? {
             name: string,
-          };
+          },
         }
       `);
     });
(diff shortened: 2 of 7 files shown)


rricard:
Okay, I still have to try it in real world situations first but this already works!

rricard:
I copy-pasted it because I saw a lot of push on the TS side to do things more specifically to TS (namespaces, the way arrays are expressed and so on...), I have to admit I kept the TS impl. close to what I wanted to output for flow. And I don't want the TS version to be limited to what flow does. So that's why I'm thinking both implementations shouldn't share code, otherwise, things may get confusing at some point when too much differences will appear between implementations...

rricard:
So the only shared code will be code between ts, flow, swift and whatever we may think of...

chollier:
This is really neat!

I wonder how it would integrate into our workflow. Do you know of any tools that would help extracting the queries and fragments from a Relay project? (We don't use `.graphql` files)

rricard:
@chollier I see your issue, I have yet to discuss with others how this could go read js files and extract `gql\`` or `Relay.QL\`` template strings...

test/flow/codeGeneration.js
```diff
@@ -0,0 +1,312 @@
...
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
+
+        export interface HeroAndDetailsQuery {
+          hero: ? HeroDetailsFragment & {
```
erydo: I've run into issues before with Flow's intersection types, particularly when there are overlapping fields (which is totally valid between fragments in GraphQL). Is there any way to add actual flow tests here that the generated code type checks?

Otherwise I imagine that these tests are going to be maintained by simply copying the generated output as it evolves, but ideally the crux of the testing would be that the generated types are correctly constrained.

test/flow/codeGeneration.js
```diff
@@ -0,0 +1,312 @@
...
+          hero {
+            name
+          }
+        }
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        //  This file was automatically generated and should not be edited.
```
erydo: Should these generated files start with the `/* @flow */` pragma?

rricard:
@erydo thanks for the feedback, this is some awesome comments. Let me finish with create-react-app's flow support (that uses a lot of `flow-bin` under the hood btw) and I'll try to add that to tests. From what I tested "by hand", this worked but you're right, let's add some more checking!

test/flow/codeGeneration.js
```diff
@@ -0,0 +1,312 @@
...
+
+        export interface HeroNameQuery {
+          hero: ? {
```
rricard: @erydo I know, as I said, this first implementation is kind of a hack, I have to rewrite some parts of how I output that to get that right!

test/flow/codeGeneration.js
rricard: Wow! thanks! forgot about that, I can try to run flow against it as much as I want, he won't see it!

erydo:
@rricard Awesome, sounds good! Since I was tinkering a bit, a couple things I discovered that might help with the potential `flow-bin` tests:
  - Putting a `.flowconfig` in an empty directory with an "ignore everything" directive helps with startup times for unit tests, and thankfully these are all standalone definitions.
  - Then you can use `flow check-contents --quiet --json the-empty-root-directory/` and pipe the tests scripts into that. The stdout of that command will be JSON with any type-checking errors.
  - A helpful pattern for pure typechecking tests is: `let x: Foo = (undefined: any);` in order to test properties on the type `Foo`. For example, this test typechecks successfully:
  ```js
  const q: HeroAndFriendsNamesQuery = (undefined: any);
  if (q.hero) {
    const name: string = q.hero.name;
  }
  ```
  And this test correctly fails:
  ```js
  const q: HeroAndFriendsNamesQuery = (undefined: any);
  const name: string = q.hero.name; // Property cannot be accessed on possibly undefined value
  ```

test/flow/codeGeneration.js
erydo: No worries! Just commenting on everything I see :) This is really good work.

rricard:
@erydo this is some great ideas (that we should also port to TS). Are you interested in trying to contribute to this? I can't work on that until tomorrow, so you can try to work on that until then if interested and I can join on your work tomorrow.

erydo:
@rricard I'm interested, but unfortunately can't guarantee much time this week. Since this is really useful I wanted to at least get some comments in there while I had the chance to look through it.

But I have a flight on Friday, so if you've gotten stalled by then, I'd be happy to use those couple hours of quiet to contribute. Otherwise, please don't wait up for me :)

rricard:
@erydo that's okay, you already greatly contributed with your comments indeed! You can check here anytime to see where I am.

stefanorg:
:+1:

rricard:
@erydo I don't think I'll add `flow-bin` right away:

1. It's not needed if we commit in checking each test file we put in the tests. That should be the case, I'm going to do that now.
2. It will slow down significantly the test suite which is not ideal...
3. I'm still open to add it (to cover this: https://github.com/apollostack/apollo-codegen/pull/24#issuecomment-266525944) but it can definitely wait for a next pr

So let's ship this one first, and then we'll refine! As said before the more we are to refine, the better so your contribs will be greatly appreciated!

rricard:
@erydo checked manually each file we're putting in the tests (as well as checking we have the right checks when using the types). It's pretty enjoyable to use as it is. As I said before, adding new ways to check can be interesting but I think we can ship without it. I ensured annoying cases pass correctly, see my next comments:

test/flow/codeGeneration.js
```diff
@@ -0,0 +1,319 @@
...
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        /* @flow */
+        //  This file was automatically generated and should not be edited.
+
+        export type HeroNameQuery = {
+          hero: ? {
```
rricard: The way code generation works doesn't let me (yet?) have those (`?` & `{`) correctly concatened. However, it is compliant with what we would do in other places!

test/flow/codeGeneration.js
```diff
@@ -0,0 +1,319 @@
...
+      `);
+
+      const source = generateSource(context);
+
+      expect(source).to.include(stripIndent`
+        /* @flow */
+        //  This file was automatically generated and should not be edited.
+
+        export type HeroAndFriendsNamesQuery = {
+          hero: ?(HeroFriendsFragment & {
```
rricard: This nullable is now properly handled. `?` being a priority annotation before `&`, I had to add paren wrapping. Now the type is correct. Either it's null or either it's the union!

test/flow/codeGeneration.js
```diff
@@ -0,0 +1,319 @@
...
+        //  This file was automatically generated and should not be edited.
+
+        // The episodes in the Star Wars trilogy
+        export type Episode =
+          "NEWHOPE" | // Star Wars Episode IV: A New Hope, released in 1977.
+          "EMPIRE" | // Star Wars Episode V: The Empire Strikes Back, released in 1980.
+          "JEDI"; // Star Wars Episode VI: Return of the Jedi, released in 1983.
+
+
+        export type ReviewInput = {
```
rricard: Everything is now a type, especially inputs. They are easier to exploit than interfaces and maybe more idiomatic in flow.

rricard:
@martijnwalraven this PR now seems pretty ready and usable to me. If you can come in for a review that would be awesome!

src/cli.js
```diff
@@ -73,7 +73,7 @@ yargs
       target: {
         demand: false,
         describe: 'Code generation target language',
-        choices: ['swift', 'json', 'ts', 'typescript'],
+        choices: ['swift', 'json', 'ts', 'typescript', 'flow', 'flowtype'],
```
martijnwalraven: Do we need both `flow` and `flowtype` here? I would say we just pick just one (`flow`?) to avoid confusion.

test/flow/codeGeneration.js
martijnwalraven: Could you give an example of how you'd want this to be formatted?

martijnwalraven:
This looks great! I don't know too much about Flow, so I can't really comment on the mapping, but we can always iterate on it.

I'm also looking at making some changes to the IR to support some additional features, and having this code on master makes it easier to add corresponding changes to the code generation.

test/flow/codeGeneration.js
rricard: `hero: ?{`

The issue is, opening a scope creates a space... This is not a big issue though and this all seem pretty reasonable to me as it is...

src/cli.js
rricard: Let's go with `flow` then!

rricard:
@martijnwalraven merging since you already approved changes and I addressed your remark.

rricard:
I'm also going to change slightly the README, stay tuned for an another PR!

rricard:
See #30.

PR #27, opened 2016-12-20
[WIP] TypeScript Rewrite

jtmthf:
This is something I wanted to show off and gauge some interest in. I started a rewrite of the apollo-codegen codebase to TypeScript for personal reasons. I was interested in adding new features to this code base, but I found it difficult to follow due to the complex data structures used throughout. So I started this fork and got a minimal viable product to compile. I have it set to compile TypeScript to es2017 and then have Babel chained to compile to es5. Essentially the TypeScript compiler only strips types in this scenario while Babel does the rest of the work. This allows the generated code to remain nearly identical. However, this conversion is still a work in progress and there's still a bit of work left in order to get it to a high enough quality to merge.

The task list of what to do is:

- [x] Convert build process to run through TypeScript and then Babel
- [x] Convert core to TypeScript
- [x] Convert utilities to TypeScript
- [ ] Convert TypeScript generator to TypeScript
- [ ] Convert Swift generator to TypeScript
- [ ] Write quality interfaces for all methods that have arguments or return types that are objects
- [ ] Define and export interface for code generator that would require generators to conform to such interface and allow for easier modularizing of generators to other repositories.
- [ ] Remove many of the explicit any's
- [ ] Verify that types are correct
- [ ] Add tslint
- [ ] Remove force null unwrapping
- [ ] Convert tests to TypeScript
- [ ] General quality clean ups

If the maintainers are interested, this could be a meaningful addition. This would bring this project more in line with both apollo-client and apollo-server that are also written in TypeScript. Additionally due to the complex data structures in the code base, it would make it easier for newcomers to follow the code and to make contributions. Lastly this change would allow for a generator interface that generators would conform to allowing this project to eventually move to a more modular structure.

Please share your thoughts and opinions on this. I could definitely use some help verifying and improving the type signatures throughout the code base, especially in compilation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 lib
+temp
--- a/package.json
+++ b/package.json
@@ -5,9 +5,9 @@
   "main": "./lib/index.js",
   "bin": "./lib/cli.js",
   "scripts": {
-    "compile": "babel src --out-dir lib --source-maps",
-    "watch": "babel src --out-dir lib --watch --source-maps",
-    "prepublish": "rm -rf lib && npm run compile",
+    "compile": "tsc && babel temp --out-dir lib --source-maps && rm -rf temp",
+    "watch": "concurrently \"tsc --watch\" \"babel temp --out-dir lib --watch --source-maps\"",
+    "prepublish": "rm -rf lib temp && npm run compile",
     "test": "mocha --recursive --compilers js:babel-register"
   },
   "repository": {
@@ -17,6 +17,10 @@
   "author": "Martijn Walraven <martijn@martijnwalraven.com>",
   "license": "MIT",
   "devDependencies": {
+    "@types/graphql": "^0.8.2",
+    "@types/node": "^6.0.52",
+    "@types/node-fetch": "^1.6.6",
+    "@types/yargs": "^6.3.3",
     "babel-cli": "^6.11.4",
     "babel-plugin-transform-runtime": "^6.12.0",
     "babel-preset-es2015": "^6.13.2",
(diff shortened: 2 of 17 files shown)


martijnwalraven:
@jtmthf: Thanks a lot for doing this! I definitely agree that the codebase could benefit from typing. I've been holding off on this because TypeScript was missing some ES6 features until recently, but this seems like a good time to make the jump.

I've also been wanting to make some simplifications (to the compiler especially), so types might help in making the logic more explicit, and ease refactoring.

I won't have time to look into the code until tomorrow because I'm traveling, but I'm really happy you started this, and it would be great to work on it together.

Not sure we need the additional step of going through Babel. What do you see as the benefit of that?

jtmthf:
@martijnwalraven Thanks for the input, the biggest reason I saw for keeping babel for now was the transform runtime plugin. It's nice to have polyfills automatically inserted instead of manually including what's needed from core-js. Not necessary at all, but a nice convenience. Perhaps though not worth the added complexity in the build process.

martijnwalraven:
@jtmthf: Yeah, I can see your point, but I don't think it is worth the added complexity. It would be nice to be able to rely on just one toolchain, which is also what we do with `apollo-client`. And because we only have to run on Node, it seems we may not need too many polyfills after all.

jtmthf:
@martijnwalraven That's fair. Looking again their's only a handful of polyfills needed. I'll change up the build process before continuing further.

lewisf:
@jtmthf would you have time to make more progress on this? it seems like the sooner we merge this in the less merge conflicts we'd have to deal with. let me know how/if I can help out

jtmthf:
@lewisf should be able to do so shortly. I opened this PR over my winter break, but now I'm back in school which has been time consuming. I'm graduating in a few weeks so I'll have a lot more time to do open source.

Also TypeScript 2.2 now has better support for `Object.create(null)` which is used a lot here and should clean up the typings.

lewisf:
@jtmthf awesome. let me know if there's anything I can do to help

lewisf:
This is looking to be really out of date -- I'm going to close this PR out for now. When you do refresh it, feel free to open a new one

PR #29, opened 2016-12-22
Non-nullable lists should be an Array type. Fixes #28

### LINKED ISSUE #28: Typescript output for non-nullable list is incorrect
Assume schema/type:
```
type Query {
  episodes: [Episode]! //non-null list
}
```

Typescript output for query:
```
query LoadEpisodes {
  episodes {
     ...EpisodeInfo
  }
}

fragment EpisodeInfo on Episode {
  name
}
```

Results in the following output:

```
export interface LoadEpisodesQuery {
  episodes: EpisodeInfoFragment;
}
```

Expected:

```
export interface LoadEpisodesQuery {
  episodes: Array<EpisodeInfoFragment>;
}
```

armstrjare:


armstrjare:
Not sure what you want to do on introducing tests here... It will require a different schema than the existing starwars/schema.json as it doesn't use non-nullable list object types.

armstrjare:
ping @martijnwalraven

rricard:
For the tests, indeed different schemas should be a good idea, but in general, I'd like to find a better way to test these cases without the need of a complete schema...

rricard:
@martijnwalraven I think we should cut a new release as well.

@armstrjare I think the same mistake exists on the Flow codegen, can you take a look to check and submit a small PR as well if necessary and possible? Thank you for spotting this one already! 😉

PR #32, opened 2016-12-28
Fix variables map to use field name instead of camel cased name / Use Double for GraphQLFloat.

davidwu226:
Hi,

I've found that if the GraphQL field names are not camel cased then the variables mapping is wrong:

Current code (wrong):

```
  public var variables: GraphQLMap? {
    return ["userId": userId]
  }
```

Fixed code (correct):

```
  public var variables: GraphQLMap? {
    return ["user_id": userId]
  }
```

This patch fixes the issue.

Also, Javascript floats are double precision, so I believe the correct mapping to Swift should be a Double instead of a Float. An update to the mapping fixes this.

### DIFF
--- a/src/swift/codeGeneration.js
+++ b/src/swift/codeGeneration.js
@@ -111,7 +111,7 @@ export function classDeclarationForOperation(
         const propertyName = camelCase(name);
         const typeName = typeNameFromGraphQLType(generator.context, type);
         const isOptional = !(type instanceof GraphQLNonNull || type.ofType instanceof GraphQLNonNull);
-        return { propertyName, type, typeName, isOptional };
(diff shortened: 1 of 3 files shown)


davidwu226:
BTW, related to this issue -- is there a reason why we're forcing camel case on the Swift side? Seems more trouble than its worth and prone to bugs such as this.

martijnwalraven:
Thanks for fixing these two issues. Not a big deal, but next time it might be clearer to split unrelated commits up into separate pull requests.

The reason for forcing camel case on the Swift side is to make the generated model types blend in with language conventions. I agree it increases the risk of bugs, but it seems worth it to me. If people feel otherwise, we could consider making it a configuration option.

PR #36, opened 2017-01-13
Add option to allow namespace export/declare for Typescript

armstrjare:
I find it useful to be able to declare a namespace and reference a global `query.schema.d.ts` file in my project, instead of having to import interfaces everywhere.

This PR adds the ability to wrap the typescript generated file in a namespace, and the option of whether to use the format `declare namespace` (for `*.d.ts` files), or `export namespace` for usage as a importable module.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -80,11 +80,27 @@ yargs
         demand: false,
         describe: "Don't attempt to map custom scalars [temporary option]",
         default: false
+      },
+      "ts-namespace": {
+        demand: false,
+        describe: "Typescript: An [optional] namespace name for generated types.",
+        default: null
+      },
+      "ts-namespace-type": {
+        demand: false,
+        describe: "Typescript: How the namespace should be generated. Either `export` or `declare`." +
+                  "Defaults to `export`, unless the `output` path ends in `.d.ts`, in which case it deafults to `declare`.",
+        default: null
       }
     },
     argv => {
       const inputPaths = argv.input.map(input => path.resolve(input));
-      const options = { passthroughCustomScalars: argv["passthrough-custom-scalars"] };
+
+      const options = {
+        passthroughCustomScalars: argv["passthrough-custom-scalars"],
+        tsNamespace: argv["ts-namespace"],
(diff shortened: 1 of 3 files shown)


armstrjare:
@martijnwalraven any chance to review this? Currently blocked on a custom fork since I need this

martijnwalraven:
@armstrjare: I don't know much about the TypeScript target, but maybe @rricard could take a look.

rricard:
Hey @armstrjare sorry for the delay... I wasn't available for contributing for a long time. I'm back now!

At the moment some time has passed. We will need to rebase that before we can do anything.

I think this feature make sense if we do the same with Flow: Flow can also declare modules. See: https://flow.org/en/docs/libdefs/creation/#declaring-a-module-a-classtoc-idtoc-declaring-a-module-hreftoc-declaring-a-modulea

We don't want to surcharge the CLI so just having a `--namespace <name>` option to cover TS & Flow would be much better. Obviously, I can help on the Flow side!

lewisf:
Sorry, I might be missing something, do you mind clarifying the intent of this PR? How do you want to use apollo-codegen's generated interfaces in practice?

Did you want all the types available globally so instead of importing them, you could just use them?

armstrjare:
@lewisf That's right - I don't want manually to import the interfaces, I want them in a d.ts file available globally.

rricard:
@lewisf And you can have the same with flow by adding a file into the libs in the `.flowconfig`.

lewisf:
Yep sounds reasonable to me. I wish the terminology for these concepts in Typescript/Flow were more similar :( Makes naming difficult. In the future we might want to consider splitting out the possible options for each target. We can target as a different subcommand so we can do something like `apollo-codegen generate typescript --namespace=... --definition-type=...` and separately `apollo-codegen generate flow --module=... --definition-type` so the commands are named after terms with which each community is familiar.

For now though, maybe we can go with `--namespace` and `--output-definition-type` or something similar?

I can help out with the Flow implementation of this as well after this lands

jbaxleyiii:

PR #44, opened 2017-01-28
skip validation of NoUnusedFragments and NoUnusedVariables

turadg:
specifiedRules` validates queries but codegen may run on just fragments of an actual query to the resolver.

### DIFF
--- a/src/validation.js
+++ b/src/validation.js
@@ -6,9 +6,11 @@ import {
 
 import { ToolError, logError } from './errors'
 
-export function validateQueryDocument(schema, document) {
-  const rules = [NoAnonymousQueries, NoExplicitTypename, NoTypenameAlias].concat(specifiedRules);
+// `specifiedRules` validates queries but codegen may run on just fragments of an actual query to the resolver.
+const baseRules = specifiedRules.filter(r => !r.name.startsWith('NoUnused'));
+const rules = [NoAnonymousQueries, NoExplicitTypename, NoTypenameAlias].concat(baseRules);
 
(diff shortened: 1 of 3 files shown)


martijnwalraven:
I'm probably missing a use case here, but when does codegen run just on fragments? As far as I know, all current targets only generate code code for operations and the fragments they depend on.

turadg:
I could have worded it better. My issue is that if the input file selection includes fragments that aren't used, the whole thing stops.

The docs show,
```
apollo-codegen generate **/*.graphql --schema schema.json --target flow --output schema.flow.js
```
(Incidentally, glob doesn't work for me so I'm doing `find . -name '*.gql' | xargs apollo-codegen …`.)

So some files are selected for input that aren't necessarily being used. I'm using codegen for Flow types so creating extra defs isn't a problem at all. I don't know if it is for other targets.

Ideally I could include the `.jsx` files that are using those fragments and the tool would pull the GraphQL queries out of those, but I thought this would be an easy stop-gap.

martijnwalraven:
Thanks for the explanation. I think there is value in knowing whether fragments (and certainly variables) are unused, so I'm a little wary of removing these validations.

Do I understand correctly that the underlying issue is that some of the actual queries are defined in `.jsx` files, while the fragments they use are defined in `.graphql` files?

I think the ability to extract queries from `jsx` files would be a great addition, and seems like a better way to solve this because you will get typings for those queries as well.

turadg:
Okay. If I find time I'll PR that instead.

PR #49, opened 2017-02-03
Add an option to allow ‘insecure’ requests to the server

ligaz:
Currently the tools is failing to download the schema if the server is
running on https with self-signed certificate. This changes adds a new
options (`skipSsl` alias `K` as in curl) that ignores invalid server
certificates.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -47,11 +47,16 @@ yargs
           return additionalHeaders;
         }
       },
+      skipSsl: {
+        alias: 'K',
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Thanks for the PR! The code looks good, but I think `skipSsl` is a bit misleading because we're not really skipping SSL but only skipping certificate checking. Maybe change it to `--insecure` like the curl option?

ligaz:
Good point. I'm pushing a commit that will address it.

martijnwalraven:
Thanks!

PR #53, opened 2017-02-08
Flag to define a prefix for custom scalars

fwal:
Builds on temporary `passthrough-custom-scalars`

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -80,11 +80,17 @@ yargs
         demand: false,
         describe: "Don't attempt to map custom scalars [temporary option]",
         default: false
+      },
+      "custom-scalars-prefix": {
+        demand: false,
+        describe: "Prefix for custom scalars. (Implies that passthrough-custom-scalars is true if set)",
+        default: '',
+        normalize: true
       }
     },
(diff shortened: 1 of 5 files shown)


martijnwalraven:
@fwal: Hmmm, I can see how type aliases can be a workaround, but I feel real mappings are a lot cleaner. With prefixes, you would still have type names like `MyAppURL` or `MyAppDate` show up everywhere in your code.

For per-type mappings, the main difficulty is finding a way to specify them I think. For now, the easiest way would probably be an additional CLI option. (With the ability to specify multiple elements, like `--header` for `download-schema`.)

In the future, we may want to see if we can use a configuration file for this, preferably some extension on a common `.graphqlrc` (see https://github.com/graphcool/graphql-config/issues/16).

What do you think?

fwal:
@martijnwalraven, I agree that per-type mappings is the proper way to go but it should be implemented properly, most likely with a config file as you describe for it to be practical.

As of now this flag should be considered a patch to the existing `passthrough-custom-scalars` option and should definitely be remove once a more solid solution is in place.

PR #54, opened 2017-02-08
Use glob to expand paths if shell fails to do so

### LINKED ISSUE #34: Windows: apollo-codegen generate does not work.
I've tested on Mac, Windows, and Bash on Ubuntu on Windows. It works great on Mac and Windows Bash, but when I run in a command prompt or powershell:

`apollo-codegen generate src/**/*.graphql --schema graphql.schema.json --target typescript --output src/schema.d.ts`

I get errors like the following:

`Error: ENOENT: no such file or directory, open 'C:\Projects\src\**\*.graphql'`

It's interpreting the wildcard path as a string

ligaz:
On some environments (Windows or Mac) if `apollo-codegen` is invoked as
npm script it fails to expand the glob input pattern. The fix is to use
`glob` module and expand the paths in code.

You [this blog post](https://medium.com/@jakubsynowiec/you-should-always-quote-your-globs-in-npm-scripts-621887a2a784) for more info.

Closes #34

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,11 +27,11 @@
     "chai": "^3.5.0",
     "chai-subset": "^1.3.0",
     "common-tags": "^1.3.0",
-    "glob": "^7.0.5",
     "mocha": "^3.1.2"
   },
   "dependencies": {
     "babel-runtime": "^6.20.0",
+    "glob": "^7.0.5",
     "source-map-support": "^0.4.2",
     "change-case": "^3.0.0",
     "graphql": "^0.9.1",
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Thanks! This makes a lot of sense.

clayne11:
When is this going to get launched?

martijnwalraven:
@clayne11: Just published a new version 0.10.7 with this PR in it.

clayne11:
Thanks for the quick response!

PR #56, opened 2017-02-10
Add a new command to generate introspection JSON from local schema

ligaz:
Introduce a new command `introspect-schema` that will generate an
introspection JSON file based on a local GraphQL schema file.

This is useful if you need to generate `schema.json` without depending
on a running server for inspection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,12 @@ apollo-codegen download-schema http://localhost:8080/graphql --output schema.jso
 
 You can use the `header` option to add additional HTTP headers to the request. For example, to include an authentication token, use `--header "Authorization: Bearer <token>"`.
 
+To generate a GraphQL schema introspection JSON from a local GraphQL schema:
+
+```sh
+apollo-codegen introspect-schema schema.graphql --output schema.json
+```
+
 This tool will generate Swift code by default from a set of query definitions in `.graphql` files:
 
 ```sh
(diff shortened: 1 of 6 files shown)


martijnwalraven:
Thanks, this looks great! Especially happy that you've also updated the README and tests.

Wondering a little bit about the `introspect-schema` command name however, because technically `download-schema` is also a way of introspecting a schema.

Alternatively, we could change this to a single `introspect-schema` command and either send an introspection query to a server or execute it on a local schema file, depending on the command argument (a file path vs. a URL). What do you think?

ligaz:
I was wondering the same and decided on a different command because `download-schema` has other arguments like headers or the insecure flag.

Ultimately I think it will be simpler for the users if they have a single command that does the introspection. We can ignore some of the options if a local path is passed as an argument.

ligaz:
Hey @martijnwalraven what should be our next step? Should I try to merge the commands into a single one?

martijnwalraven:
Yep, I'm leaning towards a single `introspect-schema` command. (Although we'd want to keep `download-schema` as an alias for now.) Do you agree?

ligaz:
@martijnwalraven I have applied the necessary changes. Please take a look.

martijnwalraven:
Thanks a lot for working on this and creating such a complete PR!

PR #66, opened 2017-02-24
Added possibleTypes of fragments to IR

### LINKED ISSUE #65: IR doesn't provide possible types required for fragments that extends interface
Original issue described here https://github.com/apollographql/apollo-android/issues/222

Query:

```
query getAllTrips {
  allTrips {
    ...commonFields
  }
}

fragment commonFields on Plan {
  id
  title
}
```

Where `Plan` extends `Trip`, response:
```
{
  "data": {
    "allTrips": [
      {
        "__typename": "Trip",
        "id": "2c494055-78bc-430c-9ab7-19817f3fc060",
        "title": "Vacation!"
      }
    ]
  }
}
```

Codegen generates IR with only one `typeCondition =Plan`.

```
"fragments": [
		{
			"fragmentName": "commonFields",
			"source": "fragment commonFields on Plan {\n  __typename\n  id\n  title\n}",
			"typeCondition": "Plan",
			"fields": [
				{
					"responseName": "id",
					"fieldName": "id",
					"type": "ID!"
				},
				{
					"responseName": "title",
					"fieldName": "title",
					"type": "String!"
				}
			],
		}
	],
```

So to be able correctly parse response on Android, we need to get access to all `possibleTypes = ["Tree", "Plan"]`

chrisvis:
Fixes #65 
One remark is that it will always add the `possibleTypes` field to fragments, also when the `typeCondition` is not an abstract type. This adds some duplications in the IR representation.

### DIFF
--- a/src/compilation.js
+++ b/src/compilation.js
@@ -149,6 +149,7 @@ export class Compiler {
     const source = print(withTypenameFieldAddedWhereNeeded(this.schema, fragmentDefinition));
 
     const typeCondition = typeFromAST(this.schema, fragmentDefinition.typeCondition);
+    const possibleTypes = this.possibleTypesForType(typeCondition)
 
     const groupedVisitedFragmentSet = new Map();
     const groupedFieldSet = this.collectFields(typeCondition, fragmentDefinition.selectionSet, undefined, groupedVisitedFragmentSet);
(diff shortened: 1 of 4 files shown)


martijnwalraven:
Sorry, I somehow completely missed this. Thanks, looks great!

Issue #73, opened 2017-03-09
Duplicate fields are generated into Fragment interface


gergely-ujvari:
I'm using version `0.10.9` and ran into the problem that some properties are generated more than one time into a Fragment interface.

I've assembled a very trivial example for this.
Schema: 
*zoo.graphql*
```
interface Feedable {
    _id: ID!,
    kind: String!
    feedWith: String!
}


interface Reptile {
    scaleSkin: String!
}

type Crocodile implements Feedable, Reptile {
    _id: ID!,
    kind: String!
    feedWith: String!
    scaleSkin: String!

    victims: Int!
}

type Iguana implements Feedable, Reptile {
    _id: ID!,
    kind: String!
    feedWith: String!
    scaleSkin: String!

    age: Int!
}
```

*root.graphql*
```
type Query {
    getZoo: [Feedable!]!
}

schema {
    query: Query
}
```

And now I setup a very simple query with some fragments:
*zooQuery.graphql*
```
query getZoo {
    getZoo {
        ...FeedableFragment
    }
}
```

*FeedableFragment*
```
fragment FeedableFragment on Feedable {
    _id
    kind,
    feedWith

    ... on Reptile {
        scaleSkin
    }

    ... on Crocodile {
        victims
    }

    ... on Iguana {
        age
    }
}
```

And now I check the generated interfaces and I have this:
```
export interface FeedableFragment {
  _id: string;
  kind: string;
  feedWith: string;
  _id: string;
  kind: string;
  feedWith: string;
  scaleSkin: string;
  victims: number;
  _id: string;
  kind: string;
  feedWith: string;
  scaleSkin: string;
  age: number;
}
```

As you can see, the `_id`, `kind`, `feedWith` fields are repeated three times, instead of one.
What's more, I think that the `scaleSkin` and `age` properties should be optional properties.

gergely-ujvari:
I have edited the description, because of a copy-paste typo some parts were copied twice into the description.

martijnwalraven:
I don't know too much about the TypeScript target, but maybe @rricard has a better idea what might be going on here.

clayne11:
I don't think it should be optional. It should be a [discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html) using `__typename` as the discriminator.

lewisf:
[...]

PR #75, opened 2017-03-09
Remove duplicate props from the generated interfaces.

### LINKED ISSUE #74: Duplicate fields on generated interfaces
In some cases, the generated interfaces can contain duplicate fields,
coming from multiple fragments.

This will cause TS errors.

csillag:
It checks that the name and the type are the same, but ignores
the rest. (This should still cover most cases.)

Fixes #74.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,12 +30,14 @@
     "mocha": "^3.1.2"
   },
   "dependencies": {
+    "@types/lodash.uniqwith": "^4.5.1",
     "babel-runtime": "^6.20.0",
     "glob": "^7.0.5",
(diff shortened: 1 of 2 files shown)


csillag:
This also fixes #73.

csillag:
> This also fixes #73.

Actually, it doesn't completely fix it, only improves the situation.
This bit is not covered:

> What's more, I think that the `scaleSkin` and `age` properties should be optional properties.

lewisf:
thanks! but #143 should fix this completely. closing this in favor of that

PR #83, opened 2017-03-21
Emit flow typedefs with exact brackets, fragment spreading

korbin:
This PR adds the Flow object spread + exact type definitions. 

https://github.com/apollographql/apollo-codegen/issues/62 is solved by this PR.

Fragment type intersections are replaced with multiple spreads within an exact object type - unions didn't seem to work with these.

Due to the use of the new object spread syntax, emitted code will break in any older version of flow.


martijnwalraven:
@korbin: Sorry, I thought I'd merge some older PRs first, but there seem to be a few conflicts now. Any chance you can fix those?

korbin:
@martijnwalraven should be good to go now - caught a bug and added an extra test case as well

martijnwalraven:
Great, thanks!

PR #86, opened 2017-04-09
Use disjoint unions on fragments w/ interfaces

lewisf:
Starting this off to address #73 by seeing if we can use disjoint/discriminated unions in flow and typescript. This will be a work in progress until the checklist is complete.

For reference on disjoint unions in Flow: https://flow.org/en/docs/types/unions/#toc-disjoint-unions
For reference on discriminated unions in Typescript, search on this page: https://www.typescriptlang.org/docs/handbook/advanced-types.html

Something that I'm still curious about is whether or not this works without an explicit discriminant. I suspect we might only be able to make this work if we use __typeName as a discriminant, which currently is not outputted in type definitions.

- [x] Test Cases
- [x] Flow Implementation
- [x] Typescript Implementation
- [x] Allow explicit __typename use on inline fragment interfaces
- [x] Flag of automatically adding __typename to everything or only when explicitly requested

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,13 @@
 {
-  "presets": ["full-node4"],
-  "plugins": ["transform-runtime"]
+  "presets": [
+    ["env", {
+      "targets": {
+        node: 4
+      }
+    }]
+  ],
+  "plugins": [
+    "transform-runtime",
+    "transform-object-rest-spread"
+  ]
 }
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # Apollo GraphQL code generator
 
-[![GitHub license](https://img.shields.io/badge/license-MIT-lightgrey.svg?maxAge=2592000)](https://raw.githubusercontent.com/apollostack/apollo-ios/master/LICENSE) [![npm](https://img.shields.io/npm/v/apollo-codegen.svg)](https://www.npmjs.com/package/apollo-codegen) [![Get on Slack](https://img.shields.io/badge/slack-join-orange.svg)](http://www.apollostack.com/#slack)
+[![GitHub license](https://img.shields.io/badge/license-MIT-lightgrey.svg?maxAge=2592000)](https://raw.githubusercontent.com/apollographql/apollo-ios/master/LICENSE) [![npm](https://img.shields.io/npm/v/apollo-codegen.svg)](https://www.npmjs.com/package/apollo-codegen) [![Get on Slack](https://img.shields.io/badge/slack-join-orange.svg)](http://www.apollostack.com/#slack)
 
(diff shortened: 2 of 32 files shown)


martijnwalraven:
It looks like disjoint / discriminated unions are the way to go here, they seem pretty powerful. We can assume `__typename` is always available, so if it isn't already we should add it to the generated definitions.

clayne11:
Very excited for this to land.

lewisf:
so the `__typename` change is a bigger change that will affect multiple pieces of apollo-codegen core so I'll do that in a separate PR. This at least enables the creation of types which means the generated Flow output is no longer invalid

clayne11:
Tests are passing! Can we merge this and release it?

clayne11:
The branch is actually not working for me:

```
yarn run v0.23.3
$ apollo-codegen download-schema http://localhost:3000/graphql --output graphql-schema.json
module.js:327
    throw err;
    ^

Error: Cannot find module 'babel-runtime/regenerator'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/Users/clayne/Library/Caches/Yarn/v1/npm-apollo-codegen-0.10.13-337c5ac58c588d3cf6f38a50a7ec6cf824f2dc98/lib/cli.js:4:20)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command
```

lewisf:
@clayne11 hold on just a little bit more! Just trying to coordinate this with #104 and there is still some work to do there.

lewisf:
Noticing there's still some issues with __typename and fragment spreads, oh boy.

lewisf:
Okay to make this easier on me, I'm going to do this in 2 parts.

Part 1: Make sure __typename is surfaced in all the right places
Part 2: Make sure __typename gets the appropriate string literal as it's type. Since __typename fields should already be in the right places, this should only further refine the generated types to allow for proper use of disjoint unions without having to typecast.

Part 2 will happen in another PR.

lewisf:
Okay this is ready -- and does change the types that are generated in some cases (mainly fragment spreads and inline fragments). this could possibly be a breaking change, but for the better.

This doesn't have the work to fully allow for disjoint unions in some cases, but that will be tackled in the next PR and since that PR is just further refining the types for `__typename` from 'string' to 'Droid' or 'Human' (whatever the type is), there won't be a further breaking change

Issue #93, opened 2017-04-23
Move to Jest + Snapshot testing


lewisf:
The last change that was committed to add description to IR broke a bunch of tests because of the `description` field is now missing from a lot of the assertions.

I think it makes sense to take the chance to move to Jest + Snapshot testing so that when we change things that do change a bunch of test output, we can just tell Jest to update all the snapshots rather than us having to manually update all the assertions.

The nice thing about doing this is that in the future -- if we are to update something that updates a bunch of the output at any step of codegen, Jest will tell us exactly what changed where so that we can confirm that our changes are sane.

There are other Snapshot libraries for Mocha that are popping up but none of them look very mature yet. I've done the transition from Mocha to Jest on a much larger project before (all of Coursera's frontend) so this transition for `apollo-codegen` should be pretty quick :)

Thoughts @martijnwalraven ?

martijnwalraven:
This sounds like a great idea! Having to update all these tests after every little change is a big pain, so that is why I got lazy with that last commit :)

Glad to hear you have experience with transitioning to Jest, really happy you want to take this on!

PR #94, opened 2017-04-23
Use Jest test runner

lewisf:
Jest snapshots seems perfectly suited for the kinds of assertions apollo-codegen
does in it's tests. Issue described in more detail in #93.

It also has a more intuitive watch mode / interface that's pleasant to use.

Surprisingly after setting up a basic configuration, everything runs out of the box, yay.

- [x] Remove mocha
- [x] Switch configuraton to use Jest
- [x] Get current failing tests passing
- [x] Switch assertions/expectations to snapshots and sanity check to make sure things are correct
- [x] Remove chai
- [x] Use Jest recommended syntax (instead of mocha/jasmine syntax even though its supported currently)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,12 +2,12 @@
   "presets": [
     ["env", {
       "targets": {
-        node: "current"
+        node: 4
       }
     }]
   ],
   "plugins": [
     "transform-runtime",
-    ["transform-object-rest-spread", { "useBuiltIns": true }]
+    "transform-object-rest-spread"
   ]
 }
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,13 @@
 {
-  "presets": ["full-node4"],
-  "plugins": ["transform-runtime"]
+  "presets": [
+    ["env", {
+      "targets": {
+        node: "current"
+      }
+    }]
+  ],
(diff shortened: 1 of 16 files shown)


lewisf:
Hey @martijnwalraven, I made enough changes to get all the current test failures passing so that you have an idea of what this looks like. Let me know what your thoughts are / if you have any questions before I finish this up

martijnwalraven:
That's indeed a lot cleaner! I haven't used Jest, but the ability to diff and update snapshots will make our lives a lot easier. Let me know if you want to discuss further or are blocked on anything, but your plan looks great.

lewisf:
Okay this is done, once the tests pass I'm going to merge it in @martijnwalraven.

One thing to be careful about is that Jest does not officially support using "this" in the test context. While it currently does work in `beforeEach/afterEach` as expected, it does not work in `beforeAll/afterAll` and has no guarantees of working in the future.

Instead, Jest recommends using closure scoping.

lewisf:
Running into an issue with travis on the node4 target :/ not sure why that's happening, but will try to figure it out

martijnwalraven:
Thanks you so much! 👍

PR #95, opened 2017-04-24
Fix incorrect rejection of queries with nested fragments.

erydo:
When a fragment is defined on a supertype, and it has inline fragments on subtypes, and that fragment is used in a context where the effective type does not match all of those subtypes, the compiler would incorrectly attempt to include fields from all of those spreads, and would fail.

This PR includes a test which crashes at first, and a follow-on commit which fixes the tests.

The following query illustrates the scenario:
```graphql
fragment CharacterFragment on Character {
  ... on Droid {
    primaryFunction
  }
  ... on Human {
    height
  }
}

query HumanAndDroid {
  human(id: "human") {
    ...CharacterFragment
  }
  droid(id: "droid") {
    ...CharacterFragment
  }
}
```

The compiler would previously incorrectly fail with the message:
```
Cannot query field "primaryFunction" on type "Human"
```

### DIFF
--- a/src/compilation.js
+++ b/src/compilation.js
@@ -200,6 +200,9 @@ export class Compiler {
             parentType;
 
           const effectiveType = parentType instanceof GraphQLObjectType ? parentType : inlineFragmentType;
+          if (inlineFragmentType !== effectiveType && !isTypeProperSuperTypeOf(this.schema, inlineFragmentType, effectiveType)) {
+            break;
+          }
 
           this.collectFields(
             effectiveType,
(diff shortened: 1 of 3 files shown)


martijnwalraven:
Thanks! Good to see you've been able to both identify and fix the issue, because I realize the compiler internals are a bit hairy and could use some cleanup. At least the tests are a lot nicer after the Jest conversion.

erydo:
The test coverage is definitely helpful. The internals are a bit complex and would be a little easier to follow with type annotations, but having a pretty well-defined failure case it made it much easier to track down.

I'm still not sure I entirely grok the compilation process, but this was the change needed to get our current schema+queries compiling from `master` as-is, so a good first step.

PR #96, opened 2017-04-24
Add support to `apollo-codegen generate` for providing an optional na…

kompfner:
…mespace in Swift.


martijnwalraven:
Using namespaces in Swift should become easier once the package manager lands and we can write modules without creating separate frameworks on iOS. Until then, the `enum` trick seems to a good workaround, so let's just merge this and revisit later.

PR #101, opened 2017-04-25
TS: Output `string` not `String` for fallback scalar type.

tekacs:
At the moment, apollo-codegen outputs the type `String` as a fallback for custom scalar types (the JS `String` type, very different from TS `string`).

After this patch, it correctly outputs `string`.

If you take a quick look at the patch, I'm fairly certain this was simply unintended behaviour (and lucky that it worked at all), caused by an implicit `toString` on a value of the wrong type.

### DIFF
--- a/src/typescript/types.js
+++ b/src/typescript/types.js
@@ -36,7 +36,7 @@ export function typeNameFromGraphQLType(context, type, bareTypeName, nullable = 
   if (type instanceof GraphQLList) {
     typeName = `Array< ${typeNameFromGraphQLType(context, type.ofType, bareTypeName, true)} >`;
   } else if (type instanceof GraphQLScalarType) {
-    typeName = builtInScalarMap[type.name] || (context.passthroughCustomScalars ? context.customScalarsPrefix + type.name: GraphQLString);
+    typeName = builtInScalarMap[type.name] || (context.passthroughCustomScalars ? context.customScalarsPrefix + type.name: builtInScalarMap[GraphQLString.name]);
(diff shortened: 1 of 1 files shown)


martijnwalraven:
Done, thanks for the review!

vjcagay:
Can the code from this PR be also added on `0.11.x`? I'm not sure why `0.10.x` and `0.11.x` have overlapping releases but I can't get `0.10.13` to work in my TypeScript project (the generated `schema.ts` file have errors) while `0.11.13` is just fine.

Sorry I forgot to mention that I tested the code on this PR on `0.11.13`.

@tekacs what version are you using?

martijnwalraven:
That is really unexpected. The `0.11.x` releases are made from the [`new-swift-target`](https://github.com/apollographql/apollo-codegen/tree/new-swift-target) branch, and are only needed when running `apollo-ios` from `master`. There should be no need to run these for TypeScript/Flow, and recent `0.10.x` releases in fact should be more up to date for these targets.

Can you post the error you get on `0.10.13` so we can look into it?

vjcagay:
Actually it doesn't really output any errors in the console at all. It's just that the resulting `schema.ts` file has errors when `apollo-codegen` tries to generate it from instrospection with multiline comments. It makes the multiline comments part of the code (that is obviously should not) when they should just remain as comments.

martijnwalraven:
Can you post an example of the generated code from `0.10.x` vs. `0.11.x` so we can look into this?

PR #103, opened 2017-04-29
Split on \n as newline character in swift codegen

qimingfang:
Otherwise, apollo-codegen results in errors where multiline comments are not wrapped correctly.


martijnwalraven:
Nice catch!

PR #104, opened 2017-05-02
Remove explicit __typename validation for non-Swift targets

lewisf:
This allows for explicit `__typename` in queries where we want to use discriminated/disjoint unions (typescript and flow) as discussed in #73. This PR naturally pairs with #86.

Once this and #73 are landed, clients will be able to use the standardized switch based type guard
```
switch (data.__typename) {
  case 'Human':
    // do human stuff
  case 'Droid':
    // do droid stuff
}
```

Or use if/else based type guards

```
if (data.__typename === 'Human') {
  // do human stuff
} else if (data.__typename === 'Droid') {
  // do droid stuff
}
```

### DIFF
--- a/src/compilation.js
+++ b/src/compilation.js
@@ -301,13 +301,11 @@ export class Compiler {
         field.isConditional = true;
       }
 
-      const isIntrospection = parentType.getFields()[fieldName] === undefined;
-      const description = isIntrospection ?
-        null :
-        parentType.getFields()[fieldName].description;
-
-      if (description) {
-        field.description = description
+      // short circuit typenames for now
+      if (fieldName === '__typename') {
+        field.type = `"${parentType}"`
+      } else {
+        field.description = parentType.getFields()[fieldName].description
       }
 
       const bareType = getNamedType(type);
--- a/src/compilation.js
+++ b/src/compilation.js
@@ -93,7 +93,7 @@ export class Compiler {
 
   addTypeUsed(type) {
     if (this.typesUsedSet.has(type)) return;
-    
+
     if (type instanceof GraphQLEnumType ||
         type instanceof GraphQLInputObjectType ||
         (type instanceof GraphQLScalarType && !isBuiltInScalarType(type))) {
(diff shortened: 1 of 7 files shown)


martijnwalraven:
Thanks for taking this on! Allowing explicit `__typename` for other targets works, but it does mean you have to add it everywhere it is needed yourself, both to switch on typename manually and for fragment matching in Apollo Client JS. 

We also have the `addTypename` option in Apollo Client JS that adds `__typename` automatically, so I’m wondering if we could take advantage of that instead, maybe by adding a similar switch to `apollo-codegen` and applying the same transform when generating types?

lewisf:
@martijnwalraven woah, thanks for the quick look!

I believe this doesn't conflict with apollo-client js because apollo-client js will only add `__typename` where needed: https://github.com/apollographql/apollo-client/blob/38b97edeffc338ff13b269faa7ef28107d8d2abb/src/queries/queryTransform.ts#L29-L37 If the developer includes it explicitly, it won't interfere with fragment matching

But you're right -- I think I need to generate `__typename` manually actually because instead of the `string` type as I currently have indicated, I need to give it the proper `string literal`. Namely, instead of 

```
export type CharacterFragmentOnHuman = {|
  name: string,
  __typename: string,
  homePlanet: ?string,
|};

export type CharacterFragmentOnDroid = {|
  name: string,
  __typename: string,
  primaryFunction: ?string,
|};
```

I need to generate

```
export type CharacterFragmentOnHuman = {|
  name: string,
  __typename: 'Human',
  homePlanet: ?string,
|};

export type CharacterFragmentOnDroid = {|
  name: string,
  __typename: 'Droid',
  primaryFunction: ?string,
|};
```

With that said -- it looks like I will have more work to do on this to pull the right __typenames. I can add a flag basically toggles

1. Add __typename types for all types
2. Add __typename types only when a selection set specifies __typename

I still think 2 is the least surprising for javascript clients and should be the default, but we can switch the default around if we need to in the future? How does that sound.

martijnwalraven:
Yes, I like the idea of making this a switch so we can reproduce the behavior of Apollo Client JS if people need it.

lewisf:
after working on this more, I don't think it makes sense anymore as a PR standalone. I'm going to merge it in with #86

PR #106, opened 2017-05-04
Add namespacing to JSONDecodable & JSONEncodable extensions in Swift

enhorn:
This adds namespacing to the generated `JSONDecodable`and `JSONEncodable` in Swift.
Consumers of the generated code can have identical symbol names, so the generated code should include the package name in the protocol reference.


martijnwalraven:
Thanks, that makes sense!

Issue #25, opened 2016-12-10
[JS/TS/Flow] Read GraphQL documents inside template strings


rricard:
Followup to https://github.com/apollostack/apollo-codegen/pull/24#issuecomment-266221363

We need to support extracting queries from template strings.

- [x] Extract from `gql` tags
- [ ] Extract and infer a fragment name for `Relay.QL` tags

rricard:
@chollier `Relay.QL` will require more work than just finding template strings since Relay dynamically generates the queries. We'll maybe need your contribution for that!

chollier:
Indeed I was thinking the same, we don't name fragments in Relay. 
I'm thinking we could maybe infer the fragment name from the filename?

I also know that with Relay2 it's planned that their would be a way to statically traverse the code, extract the Relay Queries and eventually store them as persisted queries (https://www.youtube.com/watch?v=OEfUBN9dAI8)

Also, they recently introduced babel-plugin-graphql ( https://github.com/facebook/relay/commit/c51c351de44619d1e5a05143dafad97d341d8469 ), it gives clues on how fragments will be named in the future.

jtmthf:
I was wondering for this implementation how a parser would be chosen? Would it be inferred for example that maybe [ntypescript](https://github.com/TypeStrong/ntypescript) would be used for typescript generation and [babylon](https://github.com/babel/babylon) would be used for flow generation? Or would setting the specific parser and perhaps it's setting be configurable from the CLI.

By having multiple parsers, they could each conform to an implementation interface in which they would return an array of plain text graphql source and optional paths. A modified version of `loadAndMergeQueryDocuments` would then create the merged AST. The path of execution would then look like:

CLI -> Language Specific Parser -> Standard Compiler -> Language Specific Generator -> Output

For this approach to work, the extraction of source from literal .graphql files would have to be moved into it's own "parser", but that implementation would be trivial.

stubailo:
I don't think we need a parser for this at all, a simple regex seems to work just fine: https://github.com/Poincare/persistgraphql/blob/master/src/extractFromJS.ts

stubailo:
(BTW that's the codebase for a query persisting tool for Apollo we're planning on announcing very soon)

Grmiade:
Any updates on this?

lewisf:
@Grmiade there has been a lot of talk around this topic but no progress yet because we've been trying to figure out how to create a generalized utility around extracting gql.

itajaja:
[...]

PR #113, opened 2017-05-11
[feedback] add functionality to extract gql document from gql template

itajaja:
addresses #25 

this is clearly very hap hazard, because regex is not the best tool for doing this job, but this should work for most simple cases, like:

```js
const Query = gql`
  query Foo($id: String!) {
    foo(fooId: $id) {
      ...BarFragment
    }
  }
  ${barFragment}
`
```

The alternative would be to use a different parser for ts/js/swift, but that's a much bigger effort

### DIFF
--- a/src/loading.js
+++ b/src/loading.js
@@ -23,12 +23,36 @@ export function loadSchema(schemaPath) {
   return buildClientSchema((schemaData.data) ? schemaData.data : schemaData);
 }
 
+function extractDocumentFromJavascript(content) {
+  const re = /gql`([^`]*)`/g;
+  let match
+  const matches = []
+
+  while(match = re.exec(content)) {
+    const doc = match[1]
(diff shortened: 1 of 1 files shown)


lewisf:
I think this actually will work for all cases since it works in persistgraphql. I like going with this for simplicity for now.

Are there any cases that wouldn't work?

itajaja:
I have been playing with this snippet for a while and it seems to handle all the use cases I threw at it, like multiple `gql` templates in one file, multiple fragments, etc.

itajaja:
If this looks ok, I can add docs + tests over the we

lewisf:
Yeah, this is totally reasonable, I'll merge this in now and it'll be ready with the next release. Adding docs/tests would be great :)

itajaja:
great! thanks for the speedy review 👍

PR #120, opened 2017-05-19
feat: make the tag name configurable (fixes #119)

### LINKED ISSUE #119: Make the tag name configurable
I've opted to use `graphql` tag instead of `gql`. It would be useful if `apollo-codegen` provided tag-name as an option.

Happy to raise a PR.

gajus:


rricard:
Hey @gajus, I'm sorry, we weren't really around here lately. I think we can definitely merge that if you rebase it! If you don't have time to do it, I can also do it.

rricard:
Also, will fix #120

lewisf:
Because the tag-name is only a Flow/Typescript thing -- let's add that to the help string.

lewisf:
continued this work in #148 (rebase and add the thing I requested originally). thanks for the contribution @gajus !

PR #123, opened 2017-05-23
Improve handling of fragments for Flow/Typescript targets (WIP)

lewisf:
This actually reverts a lot of the changes I made for fragments before and instead values a more generalized approach to the problem.

Essentially, handling inline fragments and fragment spreads can be better handled than the previous strategy by taking a look at their type conditions / possible types and always generating a union based on the properties specified for each type condition.

More details will be filled out here soon.

- [x] Handle merging inline fragments for Flow
- [x] Handle merging fragment spreads for Flow
- [x] Merge in IR adjustments from #128
- [ ] Adjust tests for Flow
- [ ] Copy over to Typescript


rricard:
Closing and continuing work on #143. Please do not delete the branch here.

Issue #82, opened 2017-03-20
Errors should include filename


tekacs:
apollo-codegen takes all of a project's GraphQL files at once to produce a schema, in the form **/*.graphql (or similar).

Whilst validation isn't this tool's primary job (I assume), when it displays errors, it currently doesn't display from which files those errors are sourced.

Whether embedded in something else or used directly, it seems like the resulting validation errors would be more useful if they displayed the file in which they occurred?

hzsweers:
And line number for that matter I think. Even in one file, I've had trouble understanding which block it's complaining about

martijnwalraven:
`apollo-codegen` was originally written to support Apollo iOS, and it does prefix validation errors with the  filename and line number when running from Xcode, so errors can be shown inline. See [here](https://github.com/apollographql/apollo-codegen/blob/master/src/errors.js#L34-L47) for the code.

We could definitely add the same information when running stand alone. What format would you prefer we use for that? @lewisf: I think you also mentioned wanting to add this.

hzsweers:
Something that Thrifty does nicely is when it errors out during parsing, it prints the exact line and column with it. So you'd see something like `<some message> 39:8`

lewisf:
yup, I just got back from vacation so I can take a look at this soon after I finish up a couple of pending tasks.

the information to do this in the errors is already available in the IR so it's just a matter of making sure the errors show that information if anyone wants to take a look at it -- otherwise I can get to it soon.

lewisf:
Sorry for taking so long to take a look at this! I just opened up a PR for it

tekacs:
Thanks! Sorry for not addressing it ourselves. :tada:

rricard:
Seems fixed now. Closing it.

PR #124, opened 2017-05-23
Add filenames to errors when running from CLI

lewisf:
Addresses #82. Line number information is actually incorrect in the case of non `.graphql`
source files so I don't want to expose that yet until it is fixed.

There is a lot of similarity between what is happening in this addition and the output for
xcode. It is worth combining @martijnwalraven?


lewisf:
sorry, messed this PR up. I opened up a new one

PR #126, opened 2017-05-25
Emit /* tslint:disable */ prior to comment

Sheeo:
This makes tslint configurations which use the "require-header" rule not complain about the generated file.

Would be nice with an option to add a header from CLI.


lewisf:
hey @Sheeo, thanks for this change! do you mind also updated the jest snapshots by running `jest -u` so the tests pass, thanks!

Sheeo:
Ah yes, of course -- done :)

lewisf:
Thanks for the contribution!

PR #132, opened 2017-06-08
Add typescript progressively to this codebase

rricard:
The immediate goal of this PR is to get as much help from the IDE as we can as fast as possible. For instance, by importing the graphql module we can already get some help.

However, this PR still requires babel after the typescript compilation since the typescript compiler seems to differ from babel while transpiling ES6 JS code.


martijnwalraven:
@rricard: How do you feel about merging this? Everything seems to work now without Babel, and we can convert files to TypeScript progressively as we work on new features.

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 lib
+es2015lib
```
rricard: Is that still needed if we don't use babel (a two-step compilation)?

PR #133, opened 2017-06-09
fix(flow): output nullable fragment properties as optional

yenbekbay:


yenbekbay:
@rricard Updated the snapshots.

lewisf:
looks good to me!

PR #136, opened 2017-06-10
fix(typescript) null check fragmentSpreads and inlineFragments

lanemt:
In typescript/codeGeneration.js:

fragmentSpreads and inlineFragments may be null.

This is the case for a query with a simple return type that does not return an object.

Example Query:
```
query basicQuery {
  hello: world
}
```

Example schema:
```
type Query {
  world: String
}
```

I ran into this case when querying for a `currentUserLoginEmail` that returns a string type:
```
query currentUser {
  currentUser {
    id
    contact: contactByContactId {
      id
      firstName
      lastName
    }
  }
  loginEmail: currentUserLoginEmail
}
```

### DIFF
--- a/src/flow/codeGeneration.js
+++ b/src/flow/codeGeneration.js
@@ -160,13 +160,13 @@ export function typeDeclarationForOperation(
   });
 
   properties.forEach(({ fragmentSpreads, inlineFragments, bareTypeName }) => {
-    if (fragmentSpreads.length > 0) {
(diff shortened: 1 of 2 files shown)


lanemt:
Signed the Contributor Agreement.

lanemt:
Thanks @rricard -

I took a look and just added a new commit to fix the issue for flow codeGeneration.

PR #138, opened 2017-06-12
Type the common parts (entry points)

rricard:
The goal behind this work is to provide IDE features while working on a code generator. We're not going to cover everything here but we'll at least provide enough so we can get help while writing features.


rricard:
@martijnwalraven I'd like to merge that since it works and adds a few helps without even touching the IR and the generator. I'm creating an umbrella issue to track the whole conversion to TS.

martijnwalraven:
@rricard: I've given you write access!

PR #143, opened 2017-06-15
Improve handling of fragments for Flow/Typescript targets

rricard:
Rebased version of @lewisf 's PR #123:

> This actually reverts a lot of the changes I made for fragments before and instead values a more generalized approach to the problem.
> 
> Essentially, handling inline fragments and fragment spreads can be better handled than the previous strategy by taking a look at their type conditions / possible types and always generating a union based on the properties specified for each type condition.

> More details will be filled out here soon.

> - [x] Handle merging inline fragments for Flow
> - [x] Handle merging fragment spreads for Flow
> - [x] Merge in IR adjustments from #128
> - [ ] Adjust tests for Flow
> - [ ] Copy over to Typescript

We're keeping the original PR & branch for reference in case something went bad in this rebase.


rricard:
The tests failed so keeping the old branch was not such a bad idea...

src/compilation.js
```diff
@@ -96,7 +96,7 @@ export class Compiler {
 
   addTypeUsed(type) {
     if (this.typesUsedSet.has(type)) return;
-
+
```
rricard: @rricard: clean commit needed

src/flow/codeGeneration.js
```diff
@@ -83,7 +78,7 @@ function enumerationDeclaration(generator, type) {
 function structDeclarationForInputObjectType(
   generator,
   type
-) {
+  ) {
```
rricard: @rricard: clean commit needed

src/flow/codeGeneration.js
```diff
@@ -141,104 +136,102 @@ export function typeDeclarationForOperation(
...
+  typeDeclaration(generator, {
+    interfaceName,
+    noBrackets: isAbstractType(typeCondition)
+  }, () => {
+    if (isAbstractType(typeCondition)) {
+      const propertySets = fragment.possibleTypes
+        .map(type => {
+          // NOTE: inlineFragment currently consists of the merged fields
+          // from both inline fragments and fragment spreads.
+          // TODO: Rename inlineFragments in the IR.
```
rricard: @martijnwalraven Here is an IR change we can add for your next refactor. We'll not do it here because the PR is already huge.

lewisf:
Let's get this merged :)

I do want to update docs accordingly (like why `__typename` is required for unions and interfaces) How do we want to do docs specific to targets? Would leaving a section in the README.md for Flow/Typescript be sufficient?

rricard:
That would be ok I think. At some point we have to show that targets don't exactly work the same. And that's normal I think since here we just want to output annotations while the swift target actually generates code.

lewisf:
I opened up a PR for docs on this in #146.

@martijnwalraven do you mind taking a quick look at the parts that are not flow/typescript specific? I'd love to merge this in before addressing other issues because it's a pretty big change to the code generation.

martijnwalraven:
@lewisf: I don't really know what to look for, because most changes seem specific to the Flow/TypeScript targets. Is there anything specific you want my opinion about?

I agree we should merge this as soon as possible, so we can work on other changes (such as updates to the compiler/IR) separately.

lewisf:
@martijnwalraven ah -- you're right. the other changes are fairly inconsequential. i think this is good to go so i'll go ahead and merge this in soon

PR #153, opened 2017-06-27
fix(): fix import of `fs` module

blaugold:


martijnwalraven:
Thanks for the fix! We missed this one in the TypeScript conversion.

Issue #134, opened 2017-06-09
flow: Nullable properties in query and mutation variables should be output as optional


yenbekbay:
The generated flow type
```js
export type SomeQueryVariables = {
  fieldName: ?string,
};
```

should preferably be 
```js
export type SomeQueryVariables = {
  fieldName?: ?string,
};
```

since `fieldName` is not required.

I could fix this by monkey patching https://github.com/apollographql/apollo-codegen/blob/master/src/flow/language.js#L63 as
```js
const isNullableVariableProperty =
  generator.scopeStack[0].typeName.endsWith('Variables') &&
  typeName.startsWith('?');

generator.printOnNewline(
  `${propertyName}${isNullableVariableProperty ? '?' : ''}: ${typeName}`,
);
```

but I don't think that's a good solution.

rricard:
@yenbekbay Thanks for the report. I think we can do something simpler. Can you try to put your monkey patch into a PR so we can check it out and define what to do.

Issue #81, opened 2017-03-17
Make nullable arguments optional in their typescript interface


gtod:
The graphql spec at http://facebook.github.io/graphql/#sec-Required-Non-Null-Arguments says

> If it is not non‐null, the argument is optional.

And I have been happily omitting them in my JavaScript/react-apollo projects for weeks.  But in transitioning to TypeScript I want to avoid explicitly supplying each nullable argument, with a null value.

So the following is just a hacky excursion to make such nullable arguments optional in their typescript interface declarations:

For example:

```typescript
export interface AddPassengerMutationVariables {
  bookingId: number;
  name: string;
  email: string | null;
  mobile: string | null;
  phone: string | null;
}
```
becomes

```typescript
export interface AddPassengerMutationVariables {
  bookingId: number;
  name: string;
  email?: string | null;
  mobile?: string | null;
  phone?: string | null;
}
```

Frankly I just started with typescript yesterday so this might not even be a sound idea but I just wanted to leave this here for any discussion...

```js
diff --git a/src/typescript/codeGeneration.js b/src/typescript/codeGeneration.js
index fb5dfad..d5b5b89 100644
--- a/src/typescript/codeGeneration.js
+++ b/src/typescript/codeGeneration.js
@@ -126,7 +126,7 @@ export function interfaceVariablesDeclarationForOperation(
   interfaceDeclaration(generator, {
     interfaceName,
   }, () => {
-    const properties = propertiesFromFields(generator.context, variables);
+    const properties = propertiesFromFields(generator.context, variables, false, true);
     propertyDeclarations(generator, properties, true);
   });
 }
@@ -179,17 +179,17 @@ export function interfaceDeclarationForFragment(
   });
 }
 
-export function propertiesFromFields(context, fields, forceNullable) {
-  return fields.map(field => propertyFromField(context, field, forceNullable));
+export function propertiesFromFields(context, fields, forceNullable, isVariable) {
+  return fields.map(field => propertyFromField(context, field, forceNullable, isVariable));
 }
 
-export function propertyFromField(context, field, forceNullable) {
+export function propertyFromField(context, field, forceNullable, isVariable) {
   let { name: fieldName, type: fieldType, description, fragmentSpreads, inlineFragments } = field;
   fieldName = fieldName || field.responseName;
 
   const propertyName = fieldName;
 
-  let property = { fieldName, fieldType, propertyName, description };
+  let property = { fieldName, fieldType, propertyName, description, isVariable };
 
[...]

PR #155, opened 2017-06-28
Optional Flow Inputs

tgriesser:
Removes some unused variables (`inInterface`, `forceNullable`) and adds an explicit `isNullable` property which is passed down to all property declarations, as well as an `isInput` property which is used to mark that the property should be optional in the type definition if it's nullable, as pointed out in #134. Will open a similar PR in TS to address #81.

### DIFF
--- a/src/flow/codeGeneration.js
+++ b/src/flow/codeGeneration.js
@@ -222,11 +222,11 @@ export function typeDeclarationForFragment(
   });
 }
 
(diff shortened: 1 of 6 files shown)


lewisf:
Thanks for cleaning up a bunch of stuff (and doing the same for Typescript) -- I'll get to reviewing this soon -- just wanted to let you know that I'm on it :)

PR #161, opened 2017-07-05
0.14.0

kompfner:
Bumping version to 0.14.0 to unlock work on `apollo-ios` that depends on the new `operationIdentifier` property [recently introduced](https://github.com/apollographql/apollo-codegen/pull/147) in `apollo-codegen`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "apollo-codegen",
-  "version": "0.13.2",
+  "version": "0.14.0",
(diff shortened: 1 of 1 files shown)


martijnwalraven:
It turns out the tests are currently failing (see this [PR](https://github.com/apollographql/apollo-codegen/pull/156#issuecomment-313313411)), so I don't want to risk publishing an `@latest` version with broken TypeScript output. But I'll publish it as `@next` for now.

I'm closing this pull request in favor of using `npm version minor` however, because that automatically creates a tag and is more consistent.

Issue #135, opened 2017-06-09
flow: Consider using $ReadOnlyArray instead of Array


yenbekbay:
It's generally good practice to use immutable data types. Flow has a special type called `$ReadOnlyArray` (introduced in [v0.38](https://github.com/facebook/flow/blob/master/Changelog.md#v0380)) that could be used to type GraphQL lists.

rricard:
Thanks for pointing that out! I'll definitely take a look if I have time. But I think that creating a PR could be pretty easy and a good first contribution!

Issue #129, opened 2017-06-03
Add an option to make all properties covariant


gajus:
By default, all Flowtype annotations allow rewriting property values. Add an option (or do by default) to generate type annotations where each property is covariant, i.e.

```js
/* @flow */
//  This file was automatically generated and should not be edited.

export type GetEventsQueryVariables = {|
  coordinates: ?CoordinatesInput,
  date: ?string,
  movieId: ?string,
  venueId: ?string,
|};

```

should become:

```js
/* @flow */
//  This file was automatically generated and should not be edited.

export type GetEventsQueryVariables = {|
  +coordinates: ?CoordinatesInput,
  +date: ?string,
  +movieId: ?string,
  +venueId: ?string,
|};

```

This is important, because without this a user needs to guard against all possible mutations of the data.

```js
componentWillReceiveProps (nextProps) {
  if (!this.props.reservation && nextProps.reservation) {
    // nextProps.reservation is known to be truthy
    this.props.dispatch(setTicketCount(nextProps.reservation.tickets.length));

    // without covariance, nextProps.reservation could have changed.
    // code requires additional check to ensure that nextProps.reservation is truthy.
    for (const seat of nextProps.reservation.seats) {
      this.props.dispatch(selectSeat(seat.id));
    }
  }
}

```

rricard:
@gajus Thanks for the report I did not know of that. Yes, that could be there by default. It also relates to #135 that puts read-only arrays.

I'll be happy to accept a PR on that. But I'll definitely try to implement it if I have time.

PR #163, opened 2017-07-08
Make generated flow types immutable using covariant properties (#129) and $ReadOnlyArray (#135)

hendrikniemann:
Hey,

I created this pull request to push to idea further to make the created flow types immutable. In both issues #129 and #135 @rricard said he would welcome pull request. This is not necessarily to be merged but to give an outlook how the result could look like. Unfortunately I don't have a big code base that uses apollo-codegen to test how this would affect a bigger project in practise.

These changes make covariant fields and read only arrays the *only default*. This might not be the wanted behaviour.

This is my fist PR, I would love to get some feedback and work further on this. I can also separate the commits in two PRs.

Example query `test/starwars/HeroAndFriendsNames.graphql`:
```graphql
query HeroAndFriendsNames($episode: Episode) {
  hero(episode: $episode) {
    name
    friends {
      name
    }
  }
}
```

Output (I removed the comments to highlight the changes):
```javascript
// before
export type Episode = "NEWHOPE" | "EMPIRE" | "JEDI";

export type HeroAndFriendsNamesQueryVariables = {|
  episode?: ?Episode,
|};

export type HeroAndFriendsNamesQuery = {|
  hero: ?( {
      name: string,
      friends: ?Array< ( {
          name: string,
        } | {
          name: string,
        }
      ) >,
    } | {
      name: string,
      friends: ?Array< ( {
          name: string,
        } | {
          name: string,
        }
      ) >,
    }
  ),
|};

// after
export type Episode = "NEWHOPE" | "EMPIRE" | "JEDI";

export type HeroAndFriendsNamesQueryVariables = {|
  +episode?: ?Episode,
|};

export type HeroAndFriendsNamesQuery = {|
  +hero: ?( {
      +name: string,
      +friends: ?$ReadOnlyArray< ( {
          +name: string,
        } | {
          +name: string,
        }
      ) >,
    } | {
      +name: string,
      +friends: ?$ReadOnlyArray< ( {
          +name: string,
        } | {
          +name: string,
        }
      ) >,
    }
  ),
|};
```

### DIFF
--- a/src/flow/language.js
+++ b/src/flow/language.js
@@ -43,7 +43,7 @@ export function propertyDeclaration(generator, {
   }
 
   if (closure) {
-    generator.printOnNewline(name)
+    generator.printOnNewline(`+${name}`)
     if (isInput && isNullable) {
       generator.print('?')
     }
@@ -69,7 +69,7 @@ export function propertyDeclaration(generator, {
     }
 
   } else {
-    generator.printOnNewline(name)
+    generator.printOnNewline(`+${name}`)
     if (isInput && isNullable) {
       generator.print('?')
     }
@@ -89,7 +89,7 @@ export function propertySetsDeclaration(generator, property, propertySets, stand
       })
   }
   if (!standalone) {
-    generator.printOnNewline(`${name}:`);
+    generator.printOnNewline(`+${name}:`);
   }
 
   if (isNullable) {
--- a/src/flow/language.js
+++ b/src/flow/language.js
@@ -55,7 +55,7 @@ export function propertyDeclaration(generator, {
       if (!isNullable) {
         generator.print(' ');
       }
-      generator.print(' Array<');
+      generator.print(' $ReadOnlyArray<');
     }
 
     generator.pushScope({ typeName: name });
@@ -97,7 +97,7 @@ export function propertySetsDeclaration(generator, property, propertySets, stand
   }
 
   if (isArray) {
-    generator.print('Array< ');
+    generator.print('$ReadOnlyArray< ');
   }
 
   generator.pushScope({ typeName: name });
--- a/src/flow/types.js
+++ b/src/flow/types.js
@@ -34,7 +34,7 @@ export function typeNameFromGraphQLType(context, type, bareTypeName, nullable = 
 
   let typeName;
   if (type instanceof GraphQLList) {
-    typeName = `Array< ${typeNameFromGraphQLType(context, type.ofType, bareTypeName)} >`;
+    typeName = `$ReadOnlyArray< ${typeNameFromGraphQLType(context, type.ofType, bareTypeName)} >`;
   } else if (type instanceof GraphQLScalarType) {
     typeName = builtInScalarMap[type.name] || (context.passthroughCustomScalars ? context.customScalarsPrefix + type.name : 'any');
   } else {
--- a/test/flow/__snapshots__/codeGeneration.js.snap
+++ b/test/flow/__snapshots__/codeGeneration.js.snap
@@ -5,9 +5,9 @@ exports[`Flow code generation #generateSource() should annotate custom scalars a
 //  This file was automatically generated and should not be edited.
 
 export type CustomScalarQuery = {|
-  misc: ? {|
-    __typename: string,
-    date: ?any,
+  +misc: ? {|
+    +__typename: string,
+    +date: ?any,
   |},
 |};"
 `;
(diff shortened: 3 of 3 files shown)


turadg:
> Unfortunately I don't have a big code base that uses apollo-codegen to test how this would affect a bigger project in practise.

Fwiw, I tried this with the frontend codebase at Remind. Our `download-schema` is 18K lines and our `generate --target flow` is 4000 lines. No errors.

> These changes make covariant fields and read only arrays the only default. This might not be the wanted behaviour.

Is there legit reason to mutate the query results or the mutation inputs? I think it's right to make readonly the only option.

rricard:
@lewisf we should review this PR. I think we need to know if we want to have this feature in 1.0 (maybe as an option?). It makes sense to me but it may break existing codebases relying (maybe wrongly on mutated fields).

lewisf:
Hey sorry for taking so long to get to this -- been caught up in other things.

I think we should make a decision on how our types are used before 1.0 (whether or not they are mutable or not) and have that API be set in stone. To me, it seems completely reasonable that everything is read only for both Flow and Typescript targets.

@rricard this looks like a reasonable breaking change to make since it looks like we would want immutability of these objects as default usage.

I'm open to the idea of providing a flag for existing behavior but am a fan of merging in this PR for 1.0.

I'll give this a review soon.

martijnwalraven:
I haven't really been involved in this discussion, so I may be missing context, but one thing to consider is that [the Apollo Client update API](http://dev.apollodata.com/react/api-mutations.html#graphql-mutation-options-update) actually depends on results being mutable. For example, you would make changes to the result of `readQuery` and write them back with `writeQuery`.

rricard:
@martijnwalraven that's a good point. I think the best we could do here is the following: by default make it immutable because even in the `readQuery`/`writeQuery` situation, you can actually do immutable operations if you want to. But for those that came to rely on mutation, we should have an option to deactivate strict immutability. I would call it something like `--allow-mutable` or something like this

martijnwalraven:
Hmmm, I don't think immutable types should be the default, because I expect most code using the update API will rely on mutating results.

To be honest, I'm not sure I understand the benefits of immutability at the type level, because it doesn't guarantee the behavior and only works with Flow (or is something similar possible with TypeScript?). But if people have a need for it, they should probably enable it explicitly.

rricard:
This is important, I think, when manipulating query results in a function for instance, I want Flow to ensure I'm not mutating anything in the results that I could use somewhere else. The only instances where you would want to mutate the results, you could actually avoid to do it using correctly array/object spreads and filter/map operations. Immutability is a good practice and we should help enforce it by default. However, we should make it extremely clear that we enforce that and that can be deactivated. I'll take a look at TS to see if we can do the same with it but I highly doubt it.

rricard:
However, I'll remove it from the milestone for now... If it does not make it to 1.0 however, it should be deactivated by default though.

rricard:
TS seems to support `readonly` fields. We should create a similar PR on the TS side.

lewisf:
@rricard agreed, and after further thought i'm okay without this being part of 1.0 :)

rricard:
So to get all of you updated on that, we plan to support this feature **as an option** and **support typescript as well. Since it will be integrated a s an option with no breaking change by default, **it does not take part in the 1.0 milestone**.

langpavel:
I like this as default.

rricard:
@langpavel can't do it, as @martijnwalraven noted, we expect everything to be potentially mutable. That's not the codegen that says that but the apollo client itself. Adding an option to your CLI call seems reasonable for now. We'll see very soon how to make the developer experience better on those tools and we're going to consider how to make those options easier to opt-in.

erydo:
Quick note on the Apollo client update API example that was linked:

```js
const query = gql`{ todos { ... } }`
export default graphql(gql`
  mutation ($text: String!) {
    createTodo(text: $text) { ... }
  }
`, {
  options: {
    update: (proxy, { data: { createTodo } }) => {
      const data = proxy.readQuery({ query });
      data.todos.push(createTodo);
      proxy.writeQuery({ query, data });
    },
  },
})(MyComponent);
```

That API could still be used immutably without much difficulty:
```js
    update: (proxy, { data: { createTodo } }) => {
      const orig = proxy.readQuery({ query });
      const data = {...orig, todos: [...orig.todos, createTodo]};
      proxy.writeQuery({ query, data });
    },
```

If the issue instead is that the apollo client library itself is mutating data that's shared across interface boundaries, I wonder if that should be considered an issue in the library?

jonaskello:
I noticed the same discussion for graphql-code-generator is had [here](https://github.com/dotansimha/graphql-code-generator/issues/146) (although that discussion is for typescript). To me it makes perfect senes to make the types immutable even by default. Mutation bugs will creep in as soon as something is mutable. IMO, if the current API for apollo-client requires the calling application to mutate objects, then I think the apollo-client API should change to work smoothly with immutable objects rather than forcing the calling application to work with mutable objects :-).

lewisf:
@jonaskello i agree, but while some of us live and die by immutability, it's still a preference. it seems like `apollo-client` could work w/ or w/o immutability, but while docs exist in the wild that recommend using results in a mutable way, I feel uncomfortable making immutability the default.

adding an option is easy enough -- and it seems like enabling the flag in a `npm script` (or config in the future) is a one time thing.

i can add this to the newer `flow-modern` target (and likewise for the `typescript-modern` target that is incoming, and see if that satisfies your use cases?

hendrikniemann:
Flow released [0.59](https://github.com/facebook/flow/blob/master/Changelog.md#0590) last week, which contains the `$ReadOnly` utility type. This type basically makes all properties covariant. This can be very interesting because - as far as I understand - the difference between mutable and immutable type definitions will just be wrapping the result in this type.

jonaskello:
@lewisf Yes, that makes sense. I was under the misunderstanding that the API somehow forced the use of mutable objects.

lewisf:
@hendrikniemann ah yes, if only Flow was easier to target :( In my experience, it's rare that projects at companies are on the latest versions of Flow -- generally taking at least a couple of months to catch up.

@jonaskello do you mind trying out the new `flow-modern` target (what will eventually replace the current `flow` target) in master now and seeing if it suits your needs? if so, i'm more than happy to adapt your PR to work w/ the `flow-modern` target.

hendrikniemann:
Thanks for the discussion. I will close since the code has gone through quite a change and #346 exists with the same intention.

PR #166, opened 2017-07-10
add smoke test to ci

trevorah:
I had the same lodash problem that was fixed in https://github.com/apollographql/apollo-codegen/pull/165 by @kompfner. It looks like the cli would have tested fine as the lodash dependency was still being pulled in by jest.

This PR adds a simple smoke test to make sure that the cli still starts up even when jest isnt around.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,4 @@ install:
 
 script:
   - npm test
+  - npm run test:smoke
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Ah, thanks, that should help avoid this in the future!

trevorah:
@martijnwalraven no problem, glad I could help!

Issue #168, opened 2017-07-10
Add an option to skip adding `import Apollo` in auto-generated Swift files


mochimoshi:
Since Apollo on iOS is generally installed with Cocoapods, apollo-codegen defaults to having an `import Apollo` on top of the auto-generated `API.swift` file. This is an enhancement request to allow the option to not have this import statement for projects that use other ways of including Apollo (e.g. with submodules).

For example, having such a flag like:
`generate $(find . -name '*.graphql') --schema schema.json --output API.swift --skip-swift-import-statement true`

would output

```
//  This file was automatically generated and should not be edited.

public final class MyClassQuery: GraphQLQuery {
  public static let operationString = ...
```
 instead of

```
//  This file was automatically generated and should not be edited.

import Apollo

public final class MyClassQuery: GraphQLQuery {
  public static let operationString = ...
```

Would be happy to make a quick PR to support this option.

PR #169, opened 2017-07-10
Add flag to skip adding an import statement in auto-generated Swift files (#168)

mochimoshi:
This is a PR for issue #168 

---

Since Apollo on iOS is generally installed with Cocoapods, apollo-codegen defaults to having an `import Apollo` on top of the auto-generated API.swift file. This is an enhancement request to allow the option to not have this import statement for projects that use other ways of including Apollo (e.g. with submodules).

For example, having such a flag like:
`generate $(find . -name '*.graphql') --schema schema.json --output API.swift --skip-swift-import-statement true`

would output

```

public final class MyClassQuery: GraphQLQuery {
  public static let operationString = ...
```

instead of

```

import Apollo

public final class MyClassQuery: GraphQLQuery {
  public static let operationString = ...
```


martijnwalraven:
Hmmm, the use of `import Apollo` isn't specific to CocoaPods, but is also required if you install the Apollo framework through Carthage or manually copy the framework into your project (both of which could also use submodules).

The only situation where you wouldn't use `import Apollo` is if you copied the files directly into your project and compiled them into the main module. If that is what you're doing, I would really recommend against that. Is there any particular reason you feel you need to avoid a separate `Apollo` module?

mochimoshi:
Thank you for your reply! Currently the way Apollo is integrated in our project is that it is added as a submodule, and then the files in the Sources folder is added manually to the project. This is because if I add the project or the workspace, the code-signing doesn't play well with the current project structure (it says "Team: Unknown Name (5TL2....)) and it no longer builds for device.

martijnwalraven:
It seems to me that we should fix the codesigning issue then, instead of modifying the codegen output. You really don't want third party dependencies compiled in, and there are also more parts of the code that rely on it being in an `Apollo` module.

What do you mean by 'the code-signing doesn't play well with the current project structure'? How are you adding the Apollo framework to the project? Does it work with other third party frameworks?

mochimoshi:
I see - I've updated my project to add Apollo as a subproject - I'll make the pull request for code-signing on the apollo-ios repository and close this one. By 'code-signing not playing well', we would often run into problems where subproject don't seem to be code signed correctly by Xcode (i.e. it archives into a generic archive rather than an iOS one). Thanks!

PR #171, opened 2017-07-11
Pass in HTTP Request Method for introspect-schema

alexbrashear:
The method was previously hard-coded as `POST`, this PR allows the user to specify the request method as either `GET` or `POST`.


martijnwalraven:
Thanks!

PR #173, opened 2017-07-13
Fix multi-line enum comments

bchrobot:
Line breaks in enum descriptions were not being preserved as multi-line comments.

I chose to move the comment above the enum case rather than replacing newlines with spaces to keep it all on one line.


martijnwalraven:
Thanks!

PR #176, opened 2017-07-13
Set merge-in-fields-from-fragment-spreads as boolean type

BenSchwab:
Looks like I'm a big javascript noob, but looks like I needed to include a boolean type flag to make "merge-in-fields-from-fragment-spreads" read a boolean.

I don't see any tests for the CLI, and I'm not sure how to test this locally. @martijnwalraven any advice so I can make sure I did it right this time?

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -152,8 +152,9 @@ yargs
       },
       "merge-in-fields-from-fragment-spreads": {
         demand: false,
-        describe: "Merge fragment fields onto its enclosing type [default: true]",
-        default: true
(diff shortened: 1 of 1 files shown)


martijnwalraven:
The way to test this locally is to use `npm link`, which creates a symlink in the global npm bin folder that allows you to run `apollo-codegen`.

BenSchwab:
👍 perfect tested with npm link and it works now. Sorry about that!

martijnwalraven:
No problem, published as 0.15.2!

PR #186, opened 2017-07-25
Detect inner array elements that are nullable and mark them as nullable

### LINKED ISSUE #122: Flow array elements should be marked as "maybe"
If a GraphQL list has a nullable element, its Flow element type should be prefixed with `?`.

An incorrect test snapshot is here: 

https://github.com/apollographql/apollo-codegen/blob/master/test/flow/__snapshots__/codeGeneration.js.snap#L421

Since the type of the `friends` field is `[Character]`, the generated Flow type should be:

```
  friends: ?Array< ?{|
    // The name of the character
    name: string,
  |} >,
```

instead of:
```
  friends: ?Array< {|
    // The name of the character
    name: string,
  |} >,
```

rricard:
Fixes #122

### DIFF
--- a/src/flow/codeGeneration.js
+++ b/src/flow/codeGeneration.js
@@ -242,15 +242,18 @@ export function propertyFromField(context, field) {
   if (isCompositeType(namedType)) {
     const typeName = typeNameFromGraphQLType(context, fieldType);
     let isArray = false;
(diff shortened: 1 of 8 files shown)


rricard:
@lewisf done! Good catch, this is indeed an important case

PR #187, opened 2017-07-25
Remove forced pascal casing

### LINKED ISSUE #63: Incorrect code gen of field name with multiple caps in a row
Currently in our JSON schema we have an input_obect that contains the following field
```
{
                            "name": "deviceID",
                            "description": null,
                            "type": {
                                "kind": "SCALAR",
                                "name": "String",
                                "ofType": null
                            },
                            "defaultValue": null
}
```

In the generated swift code, however, this is becoming `deviceId` and this same capitalization is being used in the endpoint request, causing it to fail as an unrecognized field.

Generated swift looks like:
```swift
public struct addUserInput: GraphQLMapConvertible {
  public var graphQLMap: GraphQLMap

  public init(firstName: String, lastName: String, identifier: String, identifierType: String, deviceId: String? = nil, clientMutationId: String? = nil) {
    graphQLMap = ["firstName": firstName, "lastName": lastName, "identifier": identifier, "identifierType": identifierType, "deviceId": deviceId, "clientMutationId": clientMutationId]
  }
}
```

### LINKED ISSUE #92: Leave typenames the way they are defined in queries for Flow/Typescript targets
I'm on the latest release as of this writing:  apollo-codegen@0.10.11

I have a Product Bill Of Materials (BOM) structure named as follows:

GraphQL input
==============
input InputProductBOMItem {
  ...
}

Generated interface
==================
export interface InputProductBomItem {    // The "om" should NOT be lower case
  ...
}

This causes errors everywhere InputProductBOMItem is referenced.

rricard:
Fixes #92 
Fixes #63 
Breaking Change

### DIFF
--- a/src/flow/codeGeneration.js
+++ b/src/flow/codeGeneration.js
@@ -14,7 +14,6 @@ import {
 
 import  { isTypeProperSuperTypeOf } from '../utilities/graphql';
 
-import { camelCase, pascalCase } from 'change-case';
 import * as Inflector from 'inflected';
 
 import {
@@ -84,7 +83,7 @@ function structDeclarationForInputObjectType(
   generator,
   type
 ) {
-  const interfaceName = pascalCase(type.name);
+  const interfaceName = type.name;
   typeDeclaration(generator, {
     interfaceName,
   }, () => {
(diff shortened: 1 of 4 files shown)


rricard:
@martijnwalraven We'll need to cut a minor release for that (since it's breaking, patch won't work...)

rricard:
@lewisf Last time I checked you were favorable to such change, I just want to confirm if we're good to roll this.

lewisf:
Yeah this looks fine to me.

Though TBH, lower case type names in Flow or Typescript just look weird. I don't think that's recommended at all. It might make sense to add warning / validation at the parsing phase? :P

rricard:
Yes, I agree, that's why I put the pascal casing originally. Do we have any preferred way to output warnings to the user @martijnwalraven? During validation? In the meantime I think we should merge this and create an issue to track the warning thing.

lewisf:
@rricard mostly what pascal casing was breaking was the case that someone names a query with acronyms -- like ACHTransaction, or something like that.

rricard:
I know, it also broke things with underscores. So it had to go away. I agree with re-adding a warning though.

PR #188, opened 2017-07-25
Top-line __typename is not a string anymore

### LINKED ISSUE #183: [TS] Use all possible strings for interface types rather than using "string"
As of a couple version ago `apollo-codegen` has been pretty amazing. The majority of the use cases around fragments were solved, but I've found a small issue.

When interfaces are used, the `__typename` is split out into a discriminated union when different data types are used. If you only fetch field from the top-line interface, however, it will generate the `__typename` simply as `string`. This is an issue if you try to combine fragments for an implementation with a fragment for the interface.

For example, if FooBar implements the Foo interface Typescript complains that `string` isn't compatible with `FooBar` when combining these fragments together.

When dealing with interface / union types, this would be solved by **always** using all available types in a union type rather than `string`.

rricard:
Replaced with the correct type name

Fixes #183

### DIFF
--- a/src/flow/codeGeneration.js
+++ b/src/flow/codeGeneration.js
@@ -145,6 +145,22 @@ export function typeDeclarationForOperation(
   }
 ) {
   const interfaceName = interfaceNameFromOperation({operationName, operationType});
+  fields = fields.map(rootField => {
+    const fields = rootField.fields.map(field => {
+      if (field.fieldName === '__typename') {
+        return {
+          ...field,
+          typeName: `"${rootField.type.name}"`,
+          type: { name: `"${rootField.type.name}"` },
+        };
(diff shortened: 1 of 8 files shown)


rricard:
@lewisf I'd like your review for two things:

- can we be more efficient/cleaner on this
- does it really fix #183? I don't think I have the right test cases to see it...

rricard:
This is definitely something we could generalize in the IR (cc @martijnwalraven), I'll try to apply your diff to see if I can test it correctly.

lewisf:
I haven't tested that code :P hopefully that works though (or something along those lines)

rricard:
@lewisf it works! I'll create an issue to track the refactor into the IR.

PR #191, opened 2017-07-26
Initial support for Scala.js with working queries and mutations

shadaj:
Adds a new target to `apollo-codegen` that emits Scala types for queries through `case class` model types and companion objects that provide type conversions.

### DIFF
--- a/src/scala/codeGeneration.js
+++ b/src/scala/codeGeneration.js
@@ -101,11 +101,11 @@ export function classDeclarationForOperation(
   switch (operationType) {
     case 'query':
       objectName = `${operationClassName(operationName)}Query`;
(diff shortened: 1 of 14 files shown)


src/scala/codeGeneration.js
```diff
@@ -0,0 +1,373 @@
...
+    operationId
+  }
+) {
+  let objectName;
+  let protocol;
+
+  switch (operationType) {
+    case 'query':
+      objectName = `${operationClassName(operationName)}Query`;
+      protocol = 'me.shadaj.apollo.GraphQLQuery';
```
martijnwalraven: What does the `me.shadaj.apollo` namespace refer to? Is this a runtime dependency on code you have written? Is there any way we can make this `com.apollographql`? :)

PR #196, opened 2017-07-30
`Replace `sjcl` with standard `crypto` module

everett1992:
I'm not familiar with `sjcl` but it seemed silly to introduce a
new dependency when node includes sha256 hashing in the standard libs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "lodash": "^4.2.0",
     "mkdirp": "^0.5.1",
     "node-fetch": "^1.5.3",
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Good catch, thanks!

everett1992:
What's the release process like? When will this be published to npm?

martijnwalraven:
We can publish a new patch release at any time, but we don't usually do that for smaller commits unless it's a bug fix. There will probably be another release either today or tomorrow.

PR #197, opened 2017-07-31
Convert compiler to TypeScript and refactor

martijnwalraven:
This PR converts the compiler to TypeScript and refactors it to make the code more understandable and fix some issues. I also used prettier to reformat the code, but I didn't add prettier to the project yet; we should discuss that separately. I've also converted the Swift target to TypeScript.

It doesn't change the IR in any way, so it should be ready to merge. I'm working on a follow-up PR with changes to the IR to support `@skip/@include` and make the structure more flexible.


lewisf:
Yes, let's use prettier! I'll take a look at this asap

martijnwalraven:
@lewisf: Great! We should also standardize on a line length then. I started out with 80 but felt it wasn't always as readable as I would like, so I switched to 110. Do you have any preference?

lewisf:
@martijnwalraven I don't have a preference -- as long as everything is consistent. Let's just go with what you have -- if we need to change this in the future it's easy enough to re-run prettier with the new setting :)

src/compilation.ts
```diff
@@ -0,0 +1,655 @@
...
+
+export interface CompiledFragment {
+  filePath?: string;
+  fragmentName: string;
+  source: string;
+  typeCondition: GraphQLCompositeType;
+  possibleTypes: GraphQLObjectType[];
+  fields: Field[];
+  fragmentSpreads: string[];
+  inlineFragments: any[];
```
lewisf: can we have a more specific type for this?

src/compilation.ts
```diff
@@ -0,0 +1,655 @@
...
+
+  const operations: {
+    [operationName: string]: CompiledOperation;
+  } = Object.create(null);
+
+  compiler.compiledOperations.forEach(operation => {
+    operations[operation.operationName] = operation;
+  });
+
+  const fragments: { [fragmentName: string]: CompiledFragment } = Object.create(null);
```
martijnwalraven: The main reason was to stay compatible with the current compiler output, because that is what the targets depend on. It also makes it easier to stringify as JSON (for Android), but there are other ways to get around that.

src/compilation.ts
```diff
@@ -0,0 +1,655 @@
...
+    const fragmentsReferenced = Array.from(fragmentsReferencedSet.keys());
+
+    const sourceWithFragments = [
+      source,
+      ...fragmentsReferenced.map(fragmentName => {
+        return this.compiledFragmentNamed(fragmentName).source;
+      })
+    ].join('\n');
+
+  const hash = createHash('sha256')
```
martijnwalraven: Yep, I think that's all prettier :)

src/compilation.ts
martijnwalraven: Oh, I thought we already did, I'll have a look at it.

martijnwalraven: Ah, it turned out I had only changed it in one location but forgot the other one. Updated.

martijnwalraven:
If you're ok with this, I think we should merge it. This PR doesn't change the compiler output, and the tests are all still passing.

I'm working on more substantial changes to the compiler and IR that are needed for `@skip`/`@include` support, among other things. But I also think they will make some other things easier (like the `__typename` literal typing we were talking about before). I hope to have something to share within the next couple of days, and I think it shouldn't be too difficult to integrate.

PR #213, opened 2017-08-02
Allow spaces between tag name and backquote

Quramy:
The following tagged template literal `q1` and `q2` are valid syntax in JS and TS, however apollo-codegen extracts GraphQL query from not `q1` but only `q2`.

```ts
const q1 = gql    `query MyQuery { ... }`; // Not extracted
const q2 = gql`query MyQuery { ... }`;   // Extracted 
```


stubailo:
Great to see you contributing here @Quramy!

PR #214, opened 2017-08-03
close apollographql/apollo-ios#124 Add in deprecations for swift

MaximusMcCann:
Swift only.  Apologies the issue was added to the appollo-ios repo.  If need be, I can create a ticket here, but I do feel it's reasonable as it's Swift specific.  Certainly open to any feedback on this. Cheers

Inserts deprecation code under code comment for properties if there's deprecation. 
uses: `__Field.isDeprecated` & `__Field.deprecationReason`
```
/// The categorization of this pie.
@available(*, deprecated, message: "Obsolete and will eventually be removed. Use the hierarchical `categorization` field instead.")
public var category: PieCategoryEnumType {
  get {
    return snapshot["category"]! as! PieCategoryEnumType
  }
  set {
    snapshot.updateValue(newValue, forKey: "category")
  }
}
```

Inserts deprecation code above the enum case if there's deprecation.
uses: `__EnumValue.isDeprecated` & `__EnumValue.deprecationReason`
```
/// A system pie status.
public enum SystemPieStatusEnum: String {
  /// Active
  @available(*, deprecated, message: "test active deprecation")
  case active = "ACTIVE"
  /// Inactive
  case inactive = "INACTIVE"
}
```
The two places graphQL allows deprecations can be seen here: https://facebook.github.io/graphql/#sec-Schema-Introspection

If no deprecation, nothing is added:
```
/// The name of this pie.
public var name: String {
  get {
    return snapshot["name"]! as! String
  }
  set {
    snapshot.updateValue(newValue, forKey: "name")
  }
}
```

### DIFF
--- a/src/swift/codeGeneration.ts
+++ b/src/swift/codeGeneration.ts
@@ -37,6 +37,7 @@ import {
   propertyDeclarations,
   escapeIdentifierIfNeeded,
   comment,
+  deprecation,
   Property
 } from './language';
 
@@ -434,7 +435,7 @@ function operationIdentifier(
 }
 
 function propertyDeclarationForField(generator: CodeGenerator, field: Field) {
-  const { kind, propertyName, typeName, type, isConditional, description } = propertyFromField(
+  const { kind, propertyName, typeName, type, isConditional, description, isDeprecated, deprecationReason } = propertyFromField(
     generator.context,
     field
   );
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Thanks! That's definitely a useful addition.

Issue #215, opened 2017-08-03
Automatically generate Equatable definition for Swift


jasonsilberman:
I think it would be very helpful to automatically generate `Equatable` definitions for the Swift code that is generated in `API.swift`.

This would be very helpful because it would allow for comparison of fetch objects. I think if `GraphQLSelectionSet` was made to conform to `Equatable`, it would allow a lot of future use cases. One immediate use case is comparing objects to see if a `UITableView` needs to be updated or not. Without `GraphQLSelectionSet` conforming to `Equatable`, it is very hard to do this.

I don't think implementing this would be very hard, we could simply generate code that goes through each property and compares them.

martijnwalraven:
Yep, I agree, that would definitely be useful. Maybe we could add in `Hashable` support as well. Want to take a stab at this?

jasonsilberman:
Great! I think I might be able to that, but I am not really familiar with the codebase. Any hints to help me get started @martijnwalraven?

martijnwalraven:
I think you'll probably want to take a look at [`structDeclarationForSelectionSet`](https://github.com/apollographql/apollo-codegen/blob/master/src/swift/codeGeneration.ts#L211), and add code to generate `Equable` (and possibly `Hashable`) support from `properties`.

adysart:
Note that Swift 4 is supposedly bringing automatic synthesis of Equatable and Hashable for structs:

https://github.com/apple/swift-evolution/blob/master/proposals/0185-synthesize-equatable-hashable.md

So this might not be necessary in the long run.

bclymer:
Swift 4.1 does bring this for free now, so I put up a PR - https://github.com/apollographql/apollo-cli/pull/578

PR #216, opened 2017-08-04
Add support for Equatable in generated structs

jasonsilberman:
As talked about in #215, this PR will automatically generate `Equatable` definitions.

There are two new functions: `equtableDeclarationForStruct` and `equtableDeclarationForFields`. The test snapshots have been updated as well.

Let me know if any changes need to be made.

### DIFF
--- a/test/swift/__snapshots__/codeGeneration.js.snap
+++ b/test/swift/__snapshots__/codeGeneration.js.snap
@@ -1103,7 +1103,7 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
   public static func ==(lhs: HeroDetails, rhs: HeroDetails) -> Bool {
     return
       lhs.name == rhs.name &&
-      lhs.friends == rhs.friends
+      lhs.friends ?? [] == rhs.friends ?? []
   }
 }"
 `;
@@ -1272,7 +1272,7 @@ exports[`Swift code generation #structDeclarationForSelectionSet() should genera
 
   public static func ==(lhs: Hero, rhs: Hero) -> Bool {
     return
-      lhs.friends == rhs.friends
+      lhs.friends ?? [] == rhs.friends ?? []
   }
 }"
 `;
--- a/test/swift/__snapshots__/codeGeneration.js.snap
+++ b/test/swift/__snapshots__/codeGeneration.js.snap
@@ -237,6 +237,149 @@ exports[`Swift code generation #classDeclarationForOperation() should generate a
 }"
 `;
 
+exports[`Swift code generation #classDeclarationForOperation() should generate a class declaration for a query with an inline fragment 1`] = `
+"public final class HeroQuery: GraphQLQuery {
+  public static let operationString =
+    \\"query Hero {\\" +
+    \\"  hero {\\" +
+    \\"    name\\" +
+    \\"    ... on Droid {\\" +
+    \\"      primaryFunction\\" +
+    \\"    }\\" +
+    \\"  }\\" +
+    \\"}\\"
+
+  public init() {
+  }
+
+  public struct Data: GraphQLSelectionSet, Equatable {
+    public static let possibleTypes = [\\"Query\\"]
+
+    public static let selections: [GraphQLSelection] = [
+      GraphQLField(\\"hero\\", type: .object(Hero.self)),
+    ]
+
+    public var snapshot: Snapshot
+
+    public init(snapshot: Snapshot) {
+      self.snapshot = snapshot
+    }
+
+    public init(hero: Hero? = nil) {
+      self.init(snapshot: [\\"__typename\\": \\"Query\\", \\"hero\\": hero])
+    }
+
+    public var hero: Hero? {
+      get {
+        return (snapshot[\\"hero\\"]! as! Snapshot?).flatMap { Hero(snapshot: $0) }
+      }
+      set {
+        snapshot.updateValue(newValue?.snapshot, forKey: \\"hero\\")
+      }
+    }
+
+    public struct Hero: GraphQLSelectionSet, Equatable {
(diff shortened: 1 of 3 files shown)


jasonsilberman:
Is there a way I can test this branch in Xcode to ensure it will compile?

jasonsilberman:
So I am testing this in my own project in Xcode and everything seems to be working but comparing optional arrays. There are no built in functions in Swift for comparing optional arrays.

To address this we could include a very small helper function that would allow the comparison of optional arrays like [this one](https://stackoverflow.com/a/40903286/1740273). However, the Apollo iOS project might be a better place to include it.

martijnwalraven:
Thanks, this looks good! I'm wondering if we got the semantics right however. I also thought we could just compare the properties, but `GraphQLSelectionSet` is a bit weird because it stores properties in a `snapshot` and allows other views on the same data.

Given this query for example:

```graphql
query Hero {
  hero {
    name
    ... on Droid {
      primaryFunction
    }
  }
}
```

Comparing two `Hero`s would only check `name`, so they may seem equal even if the result of `asDroid?.primaryFunction` is different.

I haven't thought this through completely, but would it make sense to implement equality by comparing the `snapshot`s instead?

jasonsilberman:
Thanks for the response. Here are a few thoughts:

1. We cannot really just compare `Snapshot` because they are just aliases to `[String: Any?]` which does not conform to `Equatable` natively in Swift (see image below). We could make `Snapshot` an alias to something like `[String: Equatable?]` but I'm not sure how much work would be required in other areas of the codebase to accommodate this change.
screen shot 2017-08-07 at 10 01 28 am

2. I'm not sure what you mean by "allows other views on the same data". Could you expand on this?

3. In terms of just comparing whatever properties the query asks for, I think that makes the most sense because I think people are expecting to just be looking at those properties when comparing objects. I realize this can lead to some unintentional side-effects, but it seems to be the most straight-forward approach to this.

4. Lastly, for the query you gave in your comment earlier, I think [this test](https://github.com/jasonsilberman/apollo-codegen/blob/master/test/swift/__snapshots__/codeGeneration.js.snap#L91) has a similar scenario happening. Does this test solve what you are asking?

martijnwalraven:
Hmmm, you're right, comparing `snapshot` won't work right now because the value type is `Any?`. I think changing it to `Equatable?` might work, because these values are all primitives or collections, but I'd have to try it out to know for sure.

What I mean by "allows other views on the same data" is that the properties of a `GraphQLSelectionSet` do not necessarily cover all data returned in a query. For the query I mentioned for example, `Hero` only contains properties common to all possible types (`name`), but to know if two heros are equal you may also need to check type-specific properties (like `primaryFunction`). So `AsDroid` is a type-specific view on the same result data, and the same may be true of other fragments.

test/swift/__snapshots__/codeGeneration.js.snap
```diff
@@ -237,6 +237,149 @@ exports[`Swift code generation #classDeclarationForOperation() should generate a
...
+        }
+
+        public static func ==(lhs: AsDroid, rhs: AsDroid) -> Bool {
+          return
+            lhs.name == rhs.name &&
+            lhs.primaryFunction == rhs.primaryFunction
+        }
+      }
+
+      public static func ==(lhs: Hero, rhs: Hero) -> Bool {
```
jasonsilberman: Here is where we also need to be comparing the `asDroid` property. I think that will fix what you are talking about.

jasonsilberman:
Ah okay I get what you are saying. We cannot simply switch out `Any?` to `Equatable?` though (see below).
screen shot 2017-08-07 at 3 07 32 pm

I just added an additional test with the query we are talking about. It is not doing what we want it to do here.

jasonsilberman:
@martijnwalraven, ok I think I figured out a fix to at least part of the problem. If use the `??` operator we can convert `nil` optional lists to empty lists. This way they are `Equtable` without any additional overloads.

However, we still have the problem of inline fragments, I don't think they are being properly compared.

martijnwalraven:
Hey, I'm sorry I haven't gotten to this yet. I'll be on vacation for the next week and a half, but I'll take a look after. 

I'm pretty sure we should be comparing `snapshot`s and not properties, because it will be hard to account for fragments and type conditions correctly otherwise. But that may require some changes to the `Snapshot` type.

LittleRockInGitHub:
I think an type-erased Equatable is needed here. I wrote a prototype here, and hope it may help. [AnyEquatable](https://github.com/LittleRockInGitHub/AnyEquatable/blob/master/Sources/AnyEquatable.swift)

JaviSoto:
Was this taken care of by https://github.com/apollographql/apollo-codegen/pull/290/files ?

martijnwalraven:
@JaviSoto No, that's a separate issue.

JaviSoto:
Ah, you're right, that one only added `Equatable` to enums.

nixzhu:
Is there more progress? How about generate `Codable` structs, not just a container of json?

SirensOfTitan:
I used `Sourcery` to codegen Equatable methods for the generated GraphQLAPI that others might find useful here: https://gist.github.com/SirensOfTitan/5f9feb2dca067e4b4152609e1851c028

... it's a small extension of the included Sourcery AutoEquatable template, with the gist of my additions being:
```swift
{% if type.based.GraphQLSelectionSet or type.based.GraphQLFragment or type.parent.based.GraphQLSelectionSet or type.parent.based.GraphQLFragment %}
{% call compareVariables type.computedVariables %}
```

I'm not sure if I missed any corners here, but this rule seems to have covered the entire GraphQL tree for all of my fragments in my project.

jbaxleyiii:

PR #217, opened 2017-08-05
'extend type' schema support

matthewerwin:
Added 'glob' for the schema argument (defaults to non-glob behavior).
Added extendSchema( ) calls for unified handling of the server-schema AST for both introspect/generate.
Modified star wars test script to include "extend type" in the graphqls test file.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -62,12 +62,25 @@ yargs
       }
     },
     async argv => {
-      const { schema, output, header, insecure, method } = argv;
+      let { schema, output, header, insecure, method } = argv;
 
       const urlRegex = /^https?:\/\//i;
       if (urlRegex.test(schema)) {
         await downloadSchema(schema, output, header, insecure, method);
       } else {
+        // Use glob if the user's shell was unable to expand the pattern
+        if (glob.hasMagic(schema)) {
+          schema = glob.sync(schema);
+        }
+        else {
+          schema = [schema];
+        }
+
+        schema = schema
+          .map(schema => path.resolve(schema))
+          // Sort to normalize different glob expansions between different terminals.
+          .sort();
+
         await introspectSchema(schema, output);
       }
     }
@@ -170,6 +183,12 @@ yargs
         // Sort to normalize different glob expansions between different terminals.
         .sort();
 
+      let schemaPaths = argv.schema ? [argv.schema] : [];
+      // Use glob if the user's shell was unable to expand the pattern
+      if (schemaPaths.length === 1 && glob.hasMagic(schemaPaths[0])) {
+        schemaPaths = glob.sync(schemaPaths[0]);
+      }
+
       const options = {
         passthroughCustomScalars: argv["passthrough-custom-scalars"] || argv["custom-scalars-prefix"] !== '',
         customScalarsPrefix: argv["custom-scalars-prefix"] || '',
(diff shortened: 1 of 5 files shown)


martijnwalraven:
I'm sorry we're only getting back to you now, but I'd like to better understand your use case to see what makes sense here. From what I understand, the idea of this PR is to construct a schema on the fly using multiple `.graphql` files in the schema definition language, with the ability to use `extend type` to tie things together.

Ideally, we'd like to move away from doing any schema management ourselves in favor of integrating with [graphql-config]([https://github.com/graphcool/graphql-config). That only allows you to specify a single schema file however. One option would be to leave merging up to the server, so you can then use introspection to get an already merged schema for an endpoint. Alternatively, merging on the fly might be behavior that could be added to `graphql-config`.

matthewerwin:
@martijnwalraven thanks for the thoughtful response. I understand things get busy.

Yes, the point of this is to avoid the need simply to introspect an endpoint and allow the tool to be utilized for a broader set of scenarios without having the schema mounted to a server endpoint and making use of the codegen tool for server scenarios (used in a pre-build step).

`extend type` is a pretty easy to support. See the original bug ticket (https://github.com/apollographql/apollo-codegen/issues/210) for more info.

pm-dev:
This would be a nice addition. Here's my use case:
I define my schema in various `.graphqls` files then my server uses [graphql-tools](https://www.apollographql.com/docs/graphql-tools) to instantiate a `GraphQLSchema`. There are many reasons I don't want to define my entire schema in the same file, but one example is because some mutations should only be available in 'internal' builds and I exclude these mutations for non-internal builds by not loading the `.graphqls` files they're in. Yes it's possible to introspect my server endpoint to generate the schema.json file but it would be pretty great to not need to launch my server (and deal with auth) to create the schema.json file.

squidfunk:
Any updates on this?

HideDev:
Any updates on this?

squidfunk:
I currently switched to [okgrow/merge-graphql-schemas](https://github.com/okgrow/merge-graphql-schemas) for merging schemas and have a PR open to merge arbitrary types. This removes the need to use `extend type`.

matthewerwin:
@HideDev and @squidfunk  -- The problem is that Apollo designed this tool on the basis that most people would be using client schema (i.e. consuming a 3rd party API). Using server-schema is a small addition but I feel there is a philosophical stance where they want to compartmentalize everything into very small modules....you can see it in all the v2.0 work where large breaking changes were made just to provide segregation.

While this makes tons of sense in the tree-shaking, optimization side of things it doesn't really for a CLI tool like this one which is run as a pre-build step to get type-checking against an API endpoint. The advantage of using server-schema when you own both API and UI is that you don't have to be running the API server in order to get this. I'd love to see it incorporated but am not invested in chasing people down to get it to happen. We just run our version internally.

shadaj:
With the new Apollo Config format, specifying schema dependencies is now supported in the Apollo CLI! https://github.com/apollographql/apollo-cli#schema-dependencies

matthewerwin:
@shadaj that isn't related to this. "extend type" syntax is a standard part of graphql and used in dispersed *.graphql files to indicate the type is already present, just in a different file. Having a config file that has some custom syntax to connect client/server schema isn't the same thing.

shadaj:
Hi @matthewerwin, I think there was a little confusion here. Support for `extend type` syntax is related to the config because you can specify "extending" relationships between schemas so that the CLI knows which schema AST to extend when it comes across `extend type`. This was required for client/server schemas since client schemas also use the `extend type` syntax to augment server-defined types.

I'll work on updating the docs to make this relationship between the syntax and the config more clear.

PR #226, opened 2017-08-11
Rewrite typescript generator and utilities classes to Typescript

Bolisov:
Hi,

I have rewritten typescript generator and utilities classes to Typescript.
It can be part of https://github.com/apollographql/apollo-codegen/issues/139 issue.

This pull request also contains improved **__typename** handling.

1) Now **__typename** work for not nullable root types - it was "undefined" in previous version

2) Now it is specified **__typename** for nested fields in response. Before this only string type was specified for nested properties.

Before:
```
export type MailboxThreadsQuery = {
  labels:  {
    __typename: "undefined",
    nodes:  Array< {
      __typename: string,
      threads:  {
        __typename: string,
        nodes:  Array< {
          __typename: string,
          id: string,
          subject: string | null,
          author: Contact | null,
          unread: number | null,
          time: Date,
          updated: Date | null,
          properties: Array< Property > | null,
          message:  {
            __typename: string,
            text: string | null,
          } | null,
        } >,
        total: number,
      },
    } >,
  },
};
```

Now:
```
export type MailboxThreadsQuery = {
  labels:  {
    __typename: "LabelNodesCollection",
    nodes:  Array< {
      __typename: "Label",
      threads:  {
        __typename: "ThreadNodesCollection",
        nodes:  Array< {
          __typename: "Thread",
          id: string,
          subject: string | null,
          author: Contact | null,
          unread: number | null,
          time: Date,
          updated: Date | null,
          properties: Array< Property > | null,
          message:  {
            __typename: "Message",
            text: string | null,
          } | null,
        } >,
        total: number,
      },
    } >,
  },
};
```

Also I made a little changes in Swift generator to take generateOperationIds option from context options instead of generator instance. And update Protocol interface type typing.


martijnwalraven:
Thanks for working on this! I'll merge it right away, because I'm also getting ready to merge https://github.com/apollographql/apollo-codegen/tree/new-compiler, and I think it's easier for me to rebase on top of you than the other way around (not least because of the help we get from the TypeScript compiler now that the files have been converted).

martijnwalraven:
For now, the new compiler contains a legacy IR transform, so the existing targets should continue to work. But especially https://github.com/apollographql/apollo-codegen/blob/new-compiler/src/compiler/flattenIR.ts should be of interest to the TypeScript and Flow targets. It will make it easier to generate a single type for a union of `__typename`s that share the same fields.

martijnwalraven:
I'm also working on a refactoring of the Swift target and tests that I think make the code a lot more readable. So we may want to do the same for the other targets.

Bolisov:
Great,
I think it is error in Swift generator test. I create pull request to fix it - https://github.com/apollographql/apollo-codegen/pull/227

PR #234, opened 2017-08-29
Flow code generation for scalar fields in document root

pczarn:
Now it works. There was an exception when `rootField.fields` was `undefined`


martijnwalraven:
Thanks!

PR #236, opened 2017-08-29
Allow multiline comments in enums

pczarn:
Before, multiline comments were printed correctly only in record types.


martijnwalraven:
Thanks!

PR #247, opened 2017-09-12
Use TypeScript string enums

### LINKED ISSUE #193: [TS] Use Typescript string enum for enums
As of Typescript 2.4 Typescript supports string enums. Use these instead of union string types for enums.

Example:
```typescript
// old
export type Test = "FOO" | "BAR" | "FOOBAR"

// suggested
export enum Test = {
  FOO = "FOO",
  BAR = "BAR",
  FOOBAR = "FOOBAR"
}
```

This will allow us to import and use the enums within our apps rather than having to create an identical enum.

JasonBerry:
Resolves #193

### DIFF
--- a/src/typescript/codeGeneration.ts
+++ b/src/typescript/codeGeneration.ts
@@ -73,11 +73,10 @@ function enumerationDeclaration(generator: CodeGenerator, type: GraphQLEnumType)
         generator.printOnNewline(`// ${line.trim()}`);
       })
   }
-  generator.printOnNewline(`export type ${name} =`);
-  const nValues = values.length;
-  values.forEach((value, i) => {
+  generator.printOnNewline(`export enum ${name} {`);
+  values.forEach((value) => {
     if (!value.description || value.description.indexOf('\n') === -1) {
-      generator.printOnNewline(`  "${value.value}"${i === nValues - 1 ? ';' : ' |'}${wrap(' // ', value.description)}`)
+      generator.printOnNewline(`  ${value.value} = "${value.value}",${wrap(' // ', value.description)}`)
     } else {
       if (value.description) {
         value.description.split('\n')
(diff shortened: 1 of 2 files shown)


pranava:
Thank you for working on this PR; I was wondering if there was any progress towards getting this PR merged. This PR will allow us to iterate over enum values which is really handy for UI (I.e iterate enums and generate filter options). Thanks again.

evanshauser:
@JasonBerry Thank you for the PR! This is a great addition. I'm worried that this may break some current uses of apollo-codegen by removing the union. Is that the case? If it does can we keep the current union type and export the enum?

@martijnwalraven @peggyrayzis any other blockers from merging?

martijnwalraven:
I think the main blocker could be that @lewisf is currently working on rewriting the Flow and TypeScript targets to use the new compiler, so we may want to check with him to make sure we're not interfering with his work.

JasonBerry:
This will almost certainly break current usages, though (according to semver) 0.x.x releases are considered initial development where anything can change at any time and the API shouldn't be considered stable. Given the simplicity of the fix post-upgrade, and the fact that this is definitely the better way of handling enums in Typescript I'd say it's a reasonable breaking change.

I can't imagine why anyone would ever actually want the union type once the proper enum is available?

lewisf:
Meant to respond sooner, sorry! This looks fine to merge. I'll consider this in the new typescript target I'm working on -- shouldn't be a problem at all

lewisf:
@martijnwalraven going to merge this, should we do a release to get this out?

martijnwalraven:
@lewisf: Just released 0.17.0-alpha.8 with this fix. I've been releasing alphas because I'm still working on the Swift target and I also didn't feel 100% sure about the state of the Flow/TypeScript targets with the new compiler. But if you're ok with it, I think we could release a proper 0.17.0 version soon? Or did you want to wait for your rewrite to land?

PR #251, opened 2017-09-21
Correct typename values for fragments on non-abstract types

piotr-m-jurek:
Previously `__typename: string` was generated for cases like `DroidWithName` in the tests.


leethree:
Could this be applied to Flow type generation as well?

leethree:
I just created a PR #292 to port this to Flow.

PR #258, opened 2017-09-29
help for noobs.

johndpope:
n.b.
currently need to remove this otherwise the code generation doesn't work. 
rm ExplicitTypename.graphql
rm AnonymousQuery.graphql
rm TypenameAlias.graphql


martijnwalraven:
Thanks for contributing to the docs! I think it may be confusing to let people use the test fixtures to try however. I also expect most people who come here and use `apollo-codegen` directly to be interested primarily in the TypeScript / Flow targets. For Swift developers, I think it may make more sense to improve the Apollo iOS docs and examples.

johndpope:
I think having a way to spit out some code is most needed.
 other points taken. 
the confusion around schema -> model generation needs some clarification.
query models was it for me. going to close this. 

as an academic exercise - I got this code compiling  which does spit out type models. 
https://github.com/johndpope/graphql-swift-codegen

PR #276, opened 2017-10-21
Fix __typename for List types in Flow

lydell:
I noticed that I sometimes got `__typename: "undefined"` when generating Flow code. It didn't seem to happen in TypeScript. I compared the TypeScript and Flow implementations and copied some things over from TypeScript and managed to fix the bug!

Schema:

```graphql
schema {
  query: Query
}

type Query {
  persons: [Person]
}

type Person {
  name: String
}
```

Query:

```graphql
query Test {
  persons {
    name
  }
}
```

Generated Flow code, before:

```js
export type TestQuery = {|
  persons: ? Array<? {|
    __typename: "undefined", // NOTE THE "undefined"!
    name: ?string,
  |} >,
|};
```

Generated Flow code, after:

```js
export type TestQuery = {|
  persons: ? Array<? {|
    __typename: "Person", // Yay!
    name: ?string,
  |} >,
|};
```

### DIFF
--- a/src/flow/codeGeneration.js
+++ b/src/flow/codeGeneration.js
@@ -9,7 +9,9 @@ import {
   GraphQLList,
   GraphQLNonNull,
   GraphQLID,
-  GraphQLInputObjectType
+  GraphQLInputObjectType,
+  GraphQLObjectType,
+  GraphQLUnionType
 } from 'graphql'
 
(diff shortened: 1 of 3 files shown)


martijnwalraven:
Thanks!

PR #286, opened 2017-11-01
Flow: add 'unsealed-objects' option

tempname11:
After using Flow in real projects for around a year, I strongly prefer using unsealed objects almost all the time. Exact objects bring a lot more pain in the nether regions for only a modest amount of type safety. So IMO they are not worth it.

This PR only adds an option and should not influence anyone who is not using it.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -143,6 +143,11 @@ yargs
         describe: "For non-swift targets, always add the __typename GraphQL introspection type when generating target types",
         default: false
       },
+      "unsealed-objects": {
+        demand: false,
+        describe: "Generate unsealed object typed [Flow only]",
+        default: false
+      },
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Thanks! @lewisf is working on a new Flow target, so I'd like to ask him for his opinion before merging. It may make sense to make unsealed objects the default, and I'd really prefer not to add additional options.

lewisf:
Thanks for bringing this up! We'll make this configurable in the new target. I'll open a PR on that sometime this week and ping you on it.

7ynk3r:
can't wait for this! thanks @tempname11

tempname11:
@lewisf Thanks, please do! I guess I can close this one then.

Issue #284, opened 2017-10-30
Add support for unknown enum values


abhisood:
**Problem**: whenever server returns new unsupported by client enum value it's a breaking change, as we force cast the raw string value to appropriate generated enum constant.

Simillar to apollographql/apollo-android#712 on Android
iOS issue: https://github.com/apollographql/apollo-ios/issues/174

JustinDSN:
@abhisood this should be addressed by #290.  Can you take a look and let us know?

PR #290, opened 2017-11-05
Swift Support Unknown Enums

JustinDSN:
This PR addresses the issue presented in #284.  

* Update generated enums to implement `RawRepresentable` protocol directly so that we can wrap the unknown value received from the server and echo it back on encoding.  This allows clients to use the `.unknown` enum case and actually use the value they received from the server if they choose to do so.
* Implement `Equatable` protocol on Enum to allow for `XCTAssertEqual` to work properly.

### DIFF
--- a/src/swift/codeGeneration.ts
+++ b/src/swift/codeGeneration.ts
@@ -732,30 +732,62 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
 
     this.printNewlineIfNeeded();
     this.comment(description);
-    this.printOnNewline(`public enum ${name}: String, Apollo.JSONDecodable, Apollo.JSONEncodable`);
+    this.printOnNewline(`public enum ${name}: RawRepresentable, Equatable, Apollo.JSONDecodable, Apollo.JSONEncodable`);
     this.withinBlock(() => {
+      this.printOnNewline('public typealias RawValue = String')
+
       values.forEach(value => {
         this.comment(value.description);
         this.deprecationAttributes(value.isDeprecated, value.deprecationReason);
         this.printOnNewline(
-          `case ${escapeIdentifierIfNeeded(this.helpers.enumCaseName(value.name))} = "${value.value}"`
+          `case ${escapeIdentifierIfNeeded(this.helpers.enumCaseName(value.name))}`
         );
       });
-
       this.comment('Auto generated constant for unknown enum values');
-      this.printOnNewline('case unknown');
+      this.printOnNewline('case unknown(RawValue)');
 
       this.printNewlineIfNeeded();
-      this.printOnNewline('public init(jsonValue value: JSONValue) throws');
+      this.printOnNewline('public init?(rawValue: RawValue)');
       this.withinBlock(() => {
-        this.printOnNewline('let rawValue = try RawValue(jsonValue: value)');
-        this.printOnNewline(`if let tempSelf = ${name}(rawValue: rawValue)`);
+        this.printOnNewline('switch rawValue');
         this.withinBlock(() => {
-            this.printOnNewline('self = tempSelf');
+          values.forEach(value => {
+            this.printOnNewline(
+              `case "${value.value}": self = ${escapeIdentifierIfNeeded(this.helpers.enumDotCaseName(value.name))}`
+            );
+          });
+          this.printOnNewline(`default: self = .unknown(rawValue)`);
         });
-        this.printOnNewline('else');
+      });
+
+      this.printNewlineIfNeeded();
+      this.printOnNewline('public var rawValue: RawValue');
+      this.withinBlock(() => {
+        this.printOnNewline('switch self');
         this.withinBlock(() => {
-          this.printOnNewline('self = .unknown');
+          values.forEach(value => {
+            this.printOnNewline(
+              `case ${escapeIdentifierIfNeeded(this.helpers.enumDotCaseName(value.name))}: return "${value.value}"`
+            );
+          });
+          this.printOnNewline(`case .unknown(let value): return value`);
+        });
+      });
+
+      this.printNewlineIfNeeded();
+      this.printOnNewline(`public static func == (lhs: ${name}, rhs: ${name}) -> Bool`);
+      this.withinBlock(() => {
+        this.printOnNewline('switch (lhs, rhs)');
+        this.withinBlock(() => {
+          values.forEach(value => {
+            const enumDotCaseName = escapeIdentifierIfNeeded(this.helpers.enumDotCaseName(value.name));
+            const tuple = `(${enumDotCaseName}, ${enumDotCaseName})`
+            this.printOnNewline(
+              `case ${tuple}: return true`
+            );
+          });
+          this.printOnNewline(`case (.unknown(let lhsValue), .unknown(let rhsValue)): return lhsValue == rhsValue`);
+          this.printOnNewline(`default: return false`);
         });
       });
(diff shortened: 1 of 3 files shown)


martijnwalraven:
Thanks! I think this approach makes a lot of sense, and the code looks great.

I also came across [`swift-protobuf`](https://github.com/apple/swift-protobuf/blob/master/Documentation/API.md#enum-api), which deals with unknown enum values in a similar way.

MrAlek:
This just broke our app because we actually have enum cases called `unknown`. Could we perhaps call it something like `unknownEnumValue` or `apolloUnknownEnumValue` which would lower the risk for collisions?

MrAlek:
Also, I'm a bit skeptic to this approach in general. New cases for `required` enums *is* a breaking change in a GraphQL API and should be treated as such, with an error. If the enum is optional, it could fall back to `nil` with an error in the main body, just as when the server can't fetch something.

Adding `unknown` as a case to the enum is a dangerous way to get complacent with API changes and forces the application code receiving those enum values to always handle the unexpected, losing the benefits of type safety.

JustinDSN:
@MrAlek - This one seems a little open for interpretation and for us to shape the library how we want it to behave.

Your statement for `required` enums is a breaking change works well in a development environment before the App is in the wild.  But once the app is released, when users are using your app, do you really want your app to break just by adding an Enum to your server?

Also for `optional` enums, how could you tell the difference between a truly `nil` enum and one that you received from the server, but just hasn't been added to the `enum` in Swift?

For these reasons, I feel like its valuable to have both `nil` meaning not set with a value, and `unknown` (or some other name that doesn't collide) so you get the best of both.  If you want unknown to be a breaking change you could always throw when you encounter an .unknown in your App code.

Another idea that came to mind while typing up this reply is that we could wrap the enum in another enum as a way to keep the .unknown out of the GraphQL enum type.  I'm not sure people would like that API very much, but it certainly gives us a path forward.  For example:

```
enum GrapQLEnum<T> {
    case none
    case some<T>
    case unknown<T>
}
```

MrAlek:
The problem is that, if you add new enum cases in production, your app will likely break in either case because it's impossible to plan for the unknown. It's just as much a breaking change to add an enum case as switching the type altogether. In your API, if that enum isn't critical to business logic, you could leave it optional.

Simple example: you're driving a car and come up to a traffic light. You're expecting `red`, `yellow`, or `green`. Suddenly, there's a `blue` light. What do you do? Drive or stop? If you're lucky, you pick the right thing, if not, you end up causing a traffic accident and get massively fined by the police. This is a breaking change, meaning, you shouldn't have even been driving from the start without being informed of what the blue light means.

Adding an extra `unknown` case to all enums makes every line of application code using that enum have to deal with error handling instead of centralising it do the API client. This is why we have a type system in the first place.

I think unknown enum values should be handled as any other errors, including for other types of deserialization. Error handling is however somewhat an unfinished business in Apollo-iOS though. I like the approach you're proposing at the end, it could be applied to all sorts of values & errors, not just for enums.

```
enum GrapQLOptionalValue<T> {
    case none
    case some<T>
    case error(Error)
}
```

You could map both errors sent from the server and deserialization errors in the same way. It wouldn't be that hard to work with either after implementing `flatMap` and the `?` operator.

What do you think? Also, cc @martijnwalraven

JustinDSN:
@abhisood can you weigh in on this discussion?  I quite like the code I've added and feel it addresses your issues in with a solution that prevents clients from breaking when the server adds an enum value.  It also enables the client to send back that enum value which can prevent data from being lost (ie client sending a NULL for a particular enum value, server assumes that's what you want to do and writes the NULL - this kind of stuff happens all the time with JSON+REST).  So I'm reluctant to revert this.

Since you've asked for this issue to be addressed I wanted to double check with you now that you've had some time to use it.

cc: @MrAlek, @martijnwalraven.  We can also discuss this on the contributor call on Monday.

abhisood:
@JustinDSN - First of all, thank you for fixing this issue so quickly. 

From the discussion here: https://github.com/graphql/graphql-js/issues/968, it looks like that adding an new enum value is _not_ considered a breaking change but a dangerous change.

I agree that working with enums that are changing often is troublesome and should be avoided, but the reality is that enums sometimes change and new values are added. What should be the behavior in that case? Breaking old clients is not an option. So the work around is, add a new field that is queried by new clients that knows about new enum values. 

In @MrAlek's example if the original enum looked like 
```
enum TrafficLights {
    case red
    case yellow
    case green
}
```
and we wanted to add a new option, we need to create a new enum on server.
```
enum TrafficLightsV2 {
    case red
    case yellow
    case green
    case blue
}
```

Server would now need to have another graphql API defined for getting `TrafficLightsV2` so newer clients will query for this new field. But then they have to maintain a mapping between `TrafficLights` and `TrafficLightsV2` and that logic is going to be in maintenance mode forever. 
This sort of versioning isn't great either.

MrAlek:
@abhisood The discussion in https://github.com/graphql/graphql-js/issues/968 is from the perspective of making queries, not parsing responses. It is true that adding an enum case does not break any existing queries from running. Nor does this happen if you change the return type from a `Bool` to a `String`. 

From a Swift client's perspective though, these are breaking changes to the type system. This is how Swift works and any other enum in Swift or Cocoa cannot be instantiated with an unknown value. That's the benefit of defining enums, unlike in the Javascript world where there are no enum types.

@JustinDSN Any client should of course never send anything it doesn't know what it is unless it's explicitly supposed to be opaque, like an `id` or a token of some sorts. I don't see where you would rather send an unknown enum value than an opaque id. Sending something `null` without knowing it is impossible unless you're doing some hack by converting the snapshot dictionary of a receiving data structure to something that goes back to the server. We already have protection for those kind of mistakes with double-optional parameters on input types.


The promise of Apollo on iOS is to leverage Swift to bring type safety to networked APIs. I agree that applications must be able to handle living in an imperfect world, but we should do it in a uniform manner. Type errors, unknown enum values and unknown union types should all be treated in a cohesive way because they all represent a state where the client is receiving something unexpected.

By making it uniform, once you get a value, it is strongly typed and views, models, view models can work with those values without having to constantly question their validity. We don't want to go back to Objective-C where you had to guard against `nil` everywhere, that's why the `Optional` type was created.

Looking forward to speaking with you guys tomorrow :)

martijnwalraven:
Let's discuss this at our meeting, but I think we do need at least the option for clients to deal with unknown enum values. As Lee Byron mentions in [this comment](https://github.com/graphql/graphql-js/issues/968#issuecomment-317834310):
> When building clients, it's best to be defensive against possible future expansions and handle those cases. For enums, we typically include an `else` or `switch default` clause when branching on them to handle cases the client doesn't know about. If your clients aren't programming defensively like this, then it's true that expanding the possible response values could cause issues for those clients - it's not an entirely safe change.

Without some way of representing an unknown enum value, there would be no way to do this in Swift. That is why [swift-protobuf](https://github.com/apple/swift-protobuf/blob/master/Documentation/API.md#enum-api) is using a similar solution.

I agree forcing people to deal with `unknown` is a pain, but I think it does reflect the reality of the contract between client and server.

MrAlek:
Hey guys, good meeting yesterday! Here's my proposed, alternative solution: https://gist.github.com/MrAlek/5c47645558a0dc2883dd1619dc523d3c .

Didn't get the `?` operator (was scrapped in Swift 2), but `flatMap` and mapping to optional works great, makes it easy to pass type safe values to functions, etc. What do you think?

PR #291, opened 2017-11-06
Escaping struct names

markhuot:
This isn't ready to merge, but I wanted to run it by you for a second set of eyes. The issue is that if I have a SelectionSet that asks for a field named `type` it generates a Swift struct with,

```swift
public struct Type: GraphQLSelectionSet
```

But `Type` there is a reserved word and needs to be escaped. This change runs all structs through the escaping function so the above issue doesn't happen any more.


martijnwalraven:
Yeah, that makes sense to me!

markhuot:
I added a test to check that `Type` is escaped, correctly. I'm not sure if it's worth adding a field and a test to the larger Star Wars schema files?

Anything else you need from me here?

martijnwalraven:
No, this looks good, thanks!

PR #292, opened 2017-11-07
Flow: Correct typename values for fragments on non-abstract types

leethree:
This PR just ports https://github.com/apollographql/apollo-codegen/pull/251 from Typescript to Flow.


martijnwalraven:
Thanks!

PR #294, opened 2017-11-07
 Add typeName for nested property objects

leethree:
Fix the issue that nested property objects has `__typename: string`.

Before:
```js
{
  __typename: "Human",
  friendsConnection: {|
    ___typename: string,
    edges: ? Array<? {|
      __typename: string,
      node: ?( {
          __typename: "Human",
          name: string,
        } | {
          __typename: "Droid",
          name: string,
        }
      ),
    |} >,
  |},
  starships: ? Array<? {|
    __typename: string,
    name: string,
  |} >,
}
```

After:
```js
{
  __typename: "Human",
  friendsConnection: {|
    __typename: "FriendsConnection",
    edges: ? Array<? {|
      __typename: "FriendsEdge",
      node: ?( {
          __typename: "Human",
          name: string,
        } | {
          __typename: "Droid",
          name: string,
        }
      ),
    |} >,
  |},
  starships: ? Array<? {|
    __typename: "Starship",
    name: string,
  |} >,
}
```

### DIFF
--- a/src/flow/codeGeneration.js
+++ b/src/flow/codeGeneration.js
@@ -370,7 +370,19 @@ export function propertyDeclarations(generator, properties, isInput) {
         || property.fragmentSpreads && property.fragmentSpreads.length > 0
       ) {
         propertyDeclaration(generator, property, () => {
-          const properties = propertiesFromFields(generator.context, property.fields);
+          const fields = property.fields.map(field => {
+            if (field.fieldName === '__typename') {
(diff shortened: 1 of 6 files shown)


lewisf:
I actually had to revert this since it caused the build to fail with type problems -- can we try this change with another pull request?

leethree:
@lewisf Sure, I just resubmitted PR #305

PR #296, opened 2017-11-09
Initial work on a new flow target

lewisf:
This new flow target uses babel 7 nodes to generate the resulting code. As a result, it does not extend the existing CodeGenerator.

Not having to print strings allows the code to be a little bit cleaner and makes managing the typescript target easier too.

New current defaults (subject to potential change depending on discussion):
- This new target exports `nested` types
- Multiple files in a `__generated__` folder co-located by default
    - it seems more reasonable to be able to say `import { MyQuery_hero } from './__generated__/MyQuery' than `import { MyQuery_hero } from '../../../../types/graphql_operations`

Organizationally, I've decided to co-locate tests in `__tests__` folders. Not married to this but I found it easier to work with.

Still TODO:

- [x] figure out how to consolidate the printer here w/ the existing CodeGenerator
- [x] strategy for custom scalars
- [x] solidifying api contract
- [x] optional `exact objects`

The new typescript target will require few changes from the flow target code (i've tried this already, but don't want to submit it along with the flow target until it's complete)


lewisf:
@martijnwalraven this branch is getting pretty big so I'd love to commit this work to master before continuing work on this new flow target.

It'll be accessible via a different target name "flow-modern".

I did some minor changes to the CLI, please double check those!

cc @dquang-tran who helped out with some of this work

Issue #280, opened 2017-10-25
`operation-ids-path` option is broken (Swift)


kompfner:
When passed in an `--operation-ids-path`, `apollo-codegen` should generate a file mapping between operation ids and operation sources (including dependent fragments). This functionality was added as part of https://github.com/apollographql/apollo-codegen/pull/147.

This functionality is now broken. Running `apollo-codegen` with the `--operation-ids-path` argument yields `TypeError: path must be a string or Buffer`.

There are two problems, as far as I can tell. In `writeOperationIdsMap` in `generate.ts`:
1. `context` no longer contains `operationIdsPath`
2. Each operation in `context.operations` no longer contains `operationId` or `sourceWithFragments`

My guess is that this broke as a result of the changes in https://github.com/apollographql/apollo-codegen/pull/228.

martijnwalraven:
I'm sorry the compiler rewrite broke the `--operation-ids-path` feature, but since you are the only one using it, and you know what is needed, could you open a PR to fix this?

inakianduaga:
I can confirm this is still broken

JustinDSN:
I put up a PR to address this issue.

PR #299, opened 2017-11-11
Fix Swift Operation Id Path 

JustinDSN:
* Adds an optional operationId property to the Operation object
* Bug fix for #280


martijnwalraven:
Thanks!

PR #304, opened 2017-11-25
Add typescript-modern target

lewisf:
Move flow/typescript modern targets into a `javascript` folder to clean
things up.

Make minimal modifications to cp/pasted flow target to create valid
typescript code.

This is just a starting point -- as the babel 7 ts types mature, and we
start to get complete type definitions for babel-types w/ babel7, we
will need to adjust the nodes to use proper ts nodes.


lewisf:
As I was working through this, I noticed some bugs in the flow-modern target so i fixed those up too!

ptal @martijnwalraven

PR #310, opened 2017-12-01
Sort enums for cleaner diffs

### LINKED ISSUE #250: Sort enums for cleaner diff
apollo-codegen doesn’t seem to consistently order enum values so a small schema change might result in a noisy diff.

Should be easy enough to sort alphabetically

tgriesser:
Fixes #250


lewisf:
im going to merge this in -- but we also need the same change in the *-modern targets

PR #311, opened 2017-12-01
Flow: handle multiline comments in GraphQL schema

### LINKED ISSUE #306: flow-modern target fails on some multiline comments
Using version 0.18.0 with `--target flow-modern`, I get this error:

```
TypeError: Cannot read property 'trim' of undefined
    at Printer.fixCommas (./node_modules/apollo-codegen/src/javascript/flow/printer.ts:87:24)
    at printQueue.reduce (./node_modules/apollo-codegen/src/javascript/flow/printer.ts:19:36)
```

I added this debugging,
```
                const [contents, comment] = currentLineContents.split('//');
                if (!comment) {
                  console.log({contents, comment, currentLineContents, currentLine, strippedNextLine, commentColumn, documentPart});
```
And that outputs:
```
{ contents: 'TODO make this available under another key with a Date type',
  comment: undefined,
  currentLineContents: 'TODO make this available under another key with a Date type',
  currentLine: 21,
  strippedNextLine: ',',
  commentColumn: -1,
  documentPart: 'export type AdminPeopleTab_organizations_class_owners = {\n  __typename: "ClassMembership",\n  id: number // The membership ID.\n  ,\n  group_id: number // The group ID for this membership.\n  ,\n  can_become_owner: boolean // True if this user may be promoted to an owner.\n  ,\n  can_unsubscribe: boolean // True if you may unsubscribe this user from the class\n  ,\n  can_chat: boolean // True if you may initiate a chat with this user\n  ,\n  chat_eligibility_state: ChatEligibilityState // The chat eligibility state of the user, e.g. eligible, pending, under_13\n  ,\n  role: ClassMembershipRole // The member\'s role in the class\n  ,\n  tag: ?string // An optional tag to show in the member list.\n  ,\n  query_key: string // The string to match client side filtering against.\n  ,\n  created_at: ?string // When the user joined\nTODO make this available under another key with a Date type\n  ,\n  user: AdminPeopleTab_organizations_class_owners_user,\n  org_affiliation_view: ?AdminPeopleTab_organizations_class_owners_org_affiliation_view,\n  permissions: ?AdminPeopleTab_organizations_class_owners_permissions,\n};' }
TypeError: Cannot read property 'trim' of undefined
```

So it appears to be choking on the multiline description, which in our JSON schema is
```json
              "description": "When the user joined\nTODO make this available under another key with a Date type",
```

I see that the repo has an explicit test for multiLine which is passing. My guess as to why this one isn't is that the `AdminPeopleTab` query uses a fragment. Maybe there's some interaction with the `.gql` loader?

lewisf:
Fixes #306

Just print everything on the same line for now to unbreak output.
Modifications to make the output prettier will come later

### DIFF
--- a/src/javascript/flow/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/src/javascript/flow/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -252,6 +252,41 @@ export type Episode = \\"NEWHOPE\\" | \\"EMPIRE\\" | \\"JEDI\\";
 }
 `;
 
+exports[`Flow codeGeneration handles multiline graphql comments 1`] = `
+Object {
+  "__generated__/CustomScalar.js": FlowGeneratedFile {
+    "fileContents": "
+
+/* @flow */
+// This file was automatically generated and should not be edited.
+
+// ====================================================
+// GraphQL query operation: CustomScalar
+// ====================================================
+
+export type CustomScalar_commentTest = {
+  __typename: \\"CommentTest\\",
+  multiLine: ?string, // This is a multi-line comment.
+};
+
+export type CustomScalar = {
+  commentTest: ?CustomScalar_commentTest
+};
+
+//==============================================================
+// START Enums and Input Objects
+// All enums and input objects are included in every output file
+// for now, but this will be changed soon.
+// TODO: Link to issue to fix this.
+//==============================================================
+
+//==============================================================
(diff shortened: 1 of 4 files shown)


lewisf:
@martijnwalraven can we get a build to update the flow-modern and typescript-modern targets with the fix here and in #312 ? I'm going to merge them in now. cc @turadg

PR #313, opened 2017-12-05
Redo filtering out commented out gql strings

### LINKED ISSUE #307: flow target fails on gql literals containing comments
Upgrading from 0.17.2 to 0.18.0 caused this gql literal to be left out of codegen:
```
gql`
query UserProfileView {
  me {
    id
    uuid
    # TODO https://app.asana.com/0/52068106376728/221203365474362
    role
  }
}
`
```

Copying that same query into a `.gql` file works as expected. Removing the comment line also works as expected.

lewisf:
Fixes #307

Instead of using a regex, which is going to make mistakes, pass the documents
we find through `graphql.parse` to ensure that it's something we can generate
types from.


lewisf:
@martijnwalraven could we get another patch release for this? :)

this should fix extracting `gql` from javascript documents that have comments with '//' within them.

PR #324, opened 2017-12-18
Use typescript interfaces where possible

rosskevin:
When using typescript, it is often more powerful/easier to use an `interface` as you can destructure it for reuse.  The easiest example is taking query data that is passed to react widgets, where you want those react widgets to remains strongly typed.  As a ts `type`, you cannot `Pick` from it, whereas an interface, you can `Pick<MyQuery, 'user'>` as a widget type or simply use `MyQuery['user']` to access the type inside the interface.

### DIFF
--- a/src/typescript/language.ts
+++ b/src/typescript/language.ts
@@ -30,7 +30,12 @@ export function interfaceDeclaration(generator: CodeGenerator, {
   closure: () => void) {
   generator.printNewlineIfNeeded();
   generator.printNewline();
-  generator.print(`export type ${interfaceName} = `);
+  if (noBrackets) {
+    generator.print(`export type ${interfaceName} = `);
+  } else {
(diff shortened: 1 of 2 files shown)


rasdaniil:
@rosskevin Thanks for very enlightening note on TS interfaces!

rosskevin:
Tests pass locally on node 9.3 - question out on chat as to what node versions should be tested - not sure if this is a node version dependent failure on CI.

lewisf:
i'm going to take a look at this in a bit -- and just verify that it works with other use cases.

if you look in `javascript/typescript`, which will be the target moving forward, i suspect we want to make the same change there

lewisf:
Going to fix the test failure separate from this PR. Looks like an issue w/ the npm prepublish "deprecation"

PR #327, opened 2017-12-19
Use npm4 and move prepublish -> prepare

lewisf:


lewisf:
actually, this wasn't the issue :/

PR #330, opened 2017-12-19
Generate using TS node types in Typescript Modern

leoasis:
This updates the typescript modern generator to use TS nodes from the `@babel/types` library. This fixes some invalid output in the TS generated files since we were really using flow nodes, which have slight syntax differences.

### DIFF
--- a/src/javascript/typescript/__tests__/codeGeneration.ts
+++ b/src/javascript/typescript/__tests__/codeGeneration.ts
@@ -22,7 +22,7 @@ function compile(
   return compileToIR(schema, document, options);
 }
 
-describe.only('Typescript codeGeneration', () => {
+describe('Typescript codeGeneration', () => {
   test('multiple files', () => {
     const context = compile(`
       query HeroName($episode: Episode) {
--- a/src/javascript/typescript/__tests__/codeGeneration.ts
+++ b/src/javascript/typescript/__tests__/codeGeneration.ts
@@ -22,7 +22,7 @@ function compile(
   return compileToIR(schema, document, options);
 }
 
-describe('Typescript codeGeneration', () => {
+describe.only('Typescript codeGeneration', () => {
   test('multiple files', () => {
     const context = compile(`
       query HeroName($episode: Episode) {
--- a/src/javascript/typescript/codeGeneration.ts
+++ b/src/javascript/typescript/codeGeneration.ts
@@ -77,7 +77,7 @@ export function generateSource(
   Object.values(context.operations)
     .forEach((operation) => {
       generator.fileHeader();
-      generator.typeAliasesForOperation(operation);
+      generator.interfacesForOperation(operation);
       printEnumsAndInputObjects(generator, context);
 
       const output = generator.printer.printAndClear();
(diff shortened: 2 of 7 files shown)


src/javascript/typescript/codeGeneration.ts
```diff
@@ -164,19 +163,18 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
...
-      this.typeAliasObject(operationName, properties)
+      this.interface(operationName, properties)
     );
 
     this.printer.enqueue(exportedTypeAlias);
     this.scopeStackPop();
   }
 
-  public typeAliasesForFragment(fragment: Fragment) {
+  public typesForFragment(fragment: Fragment) {
```
lewisf: thinking about this now, we might as well call this `interfacesForFragment` since that's actually whats being generated

src/javascript/typescript/__tests__/codeGeneration.ts
```diff
@@ -22,7 +22,7 @@ function compile(
   return compileToIR(schema, document, options);
 }
 
-describe('Typescript codeGeneration', () => {
+describe.only('Typescript codeGeneration', () => {
```
lewisf: i don't think you want this describe.only here

src/javascript/typescript/language.ts
```diff
@@ -69,109 +68,108 @@ export default class TypescriptGenerator {
...
+  public makeNullableType(type: t.TSType) {
+    return t.TSUnionType([
+      type,
+      // t.TSUndefinedKeyword(),
+      t.TSNullKeyword()
     ])
   }
+
+  public isNullableType(type: t.TSType) {
+    return t.isTSUnionType(type) && t.isTSNullKeyword(type.types[1]);
```
lewisf: oh cool -- would we want this to check `type.types.some(t.isTSNullKeyword)` just in case?

src/javascript/typescript/__tests__/codeGeneration.ts
leoasis: whoops!

src/javascript/typescript/language.ts
leoasis: Yeah, probably more robust, just didn't do it since we are generating the types ourselves, and the `makeNullableType` method already puts the `null` in the end. But will do it as you suggest

leoasis:
@lewisf regarding the comment about when to use types vs interfaces, the idea I was going with was just using interfaces with "simple" types, which are not part of unions. We could definitely do what you suggest, though. I think it could be useful to have that intermediate type exposed for example to write type guard functions.

src/javascript/typescript/codeGeneration.ts
leoasis: You're right 👍

leoasis:
Alright! I addressed all your comments @lewisf, ready for another look!

rosskevin:
Regarding the unions - here's a real case where I could use a well defined type for each side of the union:

We have a polymorphic type on the backend and on the front where we query (expecting one of the types) it ends up being:

```js
export type LaunchpadListQuery_user_accesses_to =
  | {}
  | {
      id: number
      name: string
      url: string
    }
```
Where one side polymorphic is `{}`.  This means we need to cast any iteration to the type we queried - which in this case I can't reference, only redefine.

TL;DR defining each side of the unions would be a useful enhancement.  Thanks @leoasis

leoasis:
Yeah, already added in the last commit. I agree they are useful to expose, so I added them to the generated output.

rosskevin:
I think I was commenting as you were committing.  You're too quick!

lewisf:
this looks good. really appreciate all your hard work on this @leoasis!

PR #342, opened 2017-12-23
Fixes around GraphQLList fields for JS Targets

### LINKED ISSUE #325: Flow Modern flattens Arrays in output
The code generation for a standard Relay connection is generated like the following. Notice that the `edges` field is not generated as an array, but instead a single object type:
> ```ts
>   edges: ?GetFoosById_foos_edges,      // A list of edges.
> ```

Schema:
```graphql
type Foo {
  id: ID!
  description: String
}

# A connection to a list of items.
type FooConnection {
  # Information to aid in pagination.
  pageInfo: PageInfo!

  # A list of edges.
  edges: [FooEdge]
}

# An edge in a connection.
type FooEdge {
  # The item at the end of the edge
  node: Foo

  # A cursor for use in pagination
  cursor: String!
}
```

Generated:
```ts
export type GetFoosById_foos_pageInfo = {|
  __typename: "PageInfo",
  hasNextPage: boolean,     // When paginating forwards, are there more items?
  hasPreviousPage: boolean, // When paginating backwards, are there more items?
  startCursor: ?string,     // When paginating backwards, the cursor to continue.
  endCursor: ?string,       // When paginating forwards, the cursor to continue.
|};

export type GetFoosById_foos_edges_node = {|
  __typename: "Foo",
  id: string, // The ID of an object
  description: ?string,
|};

export type GetFoosById_foos_edges = {|
  __typename: "FooEdge",
  cursor: string,                     // A cursor for use in pagination
  node: ?GetFoosById_foos_edges_node, // The item at the end of the edge
|};

export type GetFoosById_foos = {|
  __typename: "FooConnection",
  pageInfo: GetFoosById_foos_pageInfo, // Information to aid in pagination.
  edges: ?GetFoosById_foos_edges,      // A list of edges.
|};

export type GetFoosById = {|
  foos: ?GetFoosById_foos
|};
```

lewisf:
- Singularize type names when they are 'items' in an array
- Fix flow-modern type generation around arrays

Fixes #325

### DIFF
--- a/src/flow/codeGeneration.js
+++ b/src/flow/codeGeneration.js
@@ -16,8 +16,6 @@ import {
 
 import  { isTypeProperSuperTypeOf } from '../utilities/graphql';
 
-import * as Inflector from 'inflected';
-
 import {
   join,
   wrap,
--- a/src/javascript/flow/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/src/javascript/flow/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -12,7 +12,7 @@ Object {
 // GraphQL query operation: HeroName
 // ====================================================
 
-export type HeroName_hero_Human_friends = {
+export type HeroName_hero_Human_friend = {
   __typename: \\"Human\\",
   name: string, // What this human calls themselves
 } | {
@@ -22,15 +22,15 @@ export type HeroName_hero_Human_friends = {
 
 export type HeroName_hero = {
   __typename: \\"Human\\",
-  name: string,                          // What this human calls themselves
-  id: string,                            // The ID of the human
-  homePlanet: ?string,                   // The home planet of the human, or null if unknown
-  friends: ?HeroName_hero_Human_friends, // This human's friends, or an empty list if they have none
+  name: string,                              // What this human calls themselves
+  id: string,                                // The ID of the human
+  homePlanet: ?string,                       // The home planet of the human, or null if unknown
+  friends: ?(?HeroName_hero_Human_friend)[], // This human's friends, or an empty list if they have none
 } | {
   __typename: \\"Droid\\",
-  name: string,                          // What others call this droid
-  id: string,                            // The ID of the droid
-  appearsIn: (?Episode)[],               // The movies this droid appears in
+  name: string,                              // What others call this droid
+  id: string,                                // The ID of the droid
+  appearsIn: (?Episode)[],                   // The movies this droid appears in
 };
 
 export type HeroName = {
@@ -90,7 +90,7 @@ export type Episode = \\"NEWHOPE\\" | \\"EMPIRE\\" | \\"JEDI\\";
 // GraphQL fragment: humanFragment
 // ====================================================
 
-export type humanFragment_friends = {
+export type humanFragment_friend = {
   __typename: \\"Human\\",
   name: string, // What this human calls themselves
 } | {
@@ -100,8 +100,8 @@ export type humanFragment_friends = {
 
 export type humanFragment = {
   __typename: \\"Human\\",
-  homePlanet: ?string,             // The home planet of the human, or null if unknown
-  friends: ?humanFragment_friends, // This human's friends, or an empty list if they have none
+  homePlanet: ?string,                 // The home planet of the human, or null if unknown
+  friends: ?(?humanFragment_friend)[], // This human's friends, or an empty list if they have none
 };
 
 //==============================================================
@@ -338,22 +338,22 @@ Object {
 // GraphQL query operation: HeroName
 // ====================================================
 
-export type HeroName_hero_Human_friends = {
+export type HeroName_hero_Human_friend = {
   __typename: \\"Human\\" | \\"Droid\\",
   name: string, // The name of the character
 };
 
 export type HeroName_hero = {
   __typename: \\"Human\\",
-  name: string,                          // What this human calls themselves
-  id: string,                            // The ID of the human
-  homePlanet: ?string,                   // The home planet of the human, or null if unknown
-  friends: ?HeroName_hero_Human_friends, // This human's friends, or an empty list if they have none
+  name: string,                              // What this human calls themselves
+  id: string,                                // The ID of the human
+  homePlanet: ?string,                       // The home planet of the human, or null if unknown
+  friends: ?(?HeroName_hero_Human_friend)[], // This human's friends, or an empty list if they have none
 } | {
   __typename: \\"Droid\\",
-  name: string,                          // What others call this droid
-  id: string,                            // The ID of the droid
-  appearsIn: (?Episode)[],               // The movies this droid appears in
+  name: string,                              // What others call this droid
+  id: string,                                // The ID of the droid
+  appearsIn: (?Episode)[],                   // The movies this droid appears in
 };
 
 export type HeroName = {
@@ -389,28 +389,28 @@ Object {
 // GraphQL query operation: HeroName
 // ====================================================
 
-export type HeroName_hero_Human_friends = {
+export type HeroName_hero_Human_friend = {
   __typename: \\"Human\\" | \\"Droid\\",
   name: string, // The name of the character
 };
 
-export type HeroName_hero_Droid_friends = {
+export type HeroName_hero_Droid_friend = {
   __typename: \\"Human\\" | \\"Droid\\",
   id: string, // The ID of the character
 };
 
 export type HeroName_hero = {
   __typename: \\"Human\\",
-  name: string,                          // What this human calls themselves
-  id: string,                            // The ID of the human
-  homePlanet: ?string,                   // The home planet of the human, or null if unknown
-  friends: ?HeroName_hero_Human_friends, // This human's friends, or an empty list if they have none
+  name: string,                              // What this human calls themselves
+  id: string,                                // The ID of the human
+  homePlanet: ?string,                       // The home planet of the human, or null if unknown
+  friends: ?(?HeroName_hero_Human_friend)[], // This human's friends, or an empty list if they have none
 } | {
   __typename: \\"Droid\\",
-  name: string,                          // What others call this droid
-  id: string,                            // The ID of the droid
-  appearsIn: (?Episode)[],               // The movies this droid appears in
-  friends: ?HeroName_hero_Droid_friends, // This droid's friends, or an empty list if they have none
+  name: string,                              // What others call this droid
+  id: string,                                // The ID of the droid
+  appearsIn: (?Episode)[],                   // The movies this droid appears in
+  friends: ?(?HeroName_hero_Droid_friend)[], // This droid's friends, or an empty list if they have none
 };
 
 export type HeroName = {
--- a/src/javascript/flow/codeGeneration.ts
+++ b/src/javascript/flow/codeGeneration.ts
@@ -16,6 +16,10 @@ import {
 } from '../../compiler';
 
 import {
+  getTypeNameFromFieldName
+} from '../naming';
+
+import {
   typeCaseForSelectionSet,
   Variant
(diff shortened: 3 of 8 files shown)


erydo:
[The core changes here](https://github.com/apollographql/apollo-codegen/pull/342/files#diff-65d5abdd03adbf766bbb7854931167bbR53) look great 👍  and I'd love to see those in!

I agree with you that the type names ending in plurals aren't aesthetically appealing, and can be even misleading, since they're actually representing `T` in each of the field types `foos: [T]`, `foos: [T!]!`, `foo: [T]!`, rather than the type of the field itself.

But as a user, the idea of heuristically singularizing the field names in the types makes me a little hesitant:

1. It might cause issues with the canonical Relay plural identifying root field by emitting multiple `Query_node` types for queries against schemas like:
    ```graphql
    type Query {
      node(id: ID!): Node
      nodes(ids: [ID!]!): [Node]
    }
    ```
2. Queries against types that have evolved to support newly plural relationships (_"we thought we would only need `friend` but need to add `friends`; and we can't remove the old field yet"_) would possibly also break.
3. The predictable mapping to the field name was helpful. But for example:
    ```graphql
    query { foos { bar { bazzes { name } } } }
    ```
    If that generated types like `Query_foo_bar_baz`, that could be unexpected: there's no identifier `foo` or `baz` present anywhere in the schema or the query; grepping for 'bazzes' would skip those type definitions. Not sure if that's practically important, though.
3. Assuming identifiers are plural in the first place is error-prone. Depending who designed the schema, not all field names are necessarily nouns. Since apollo-codegen is mainly used for creating client-side types, the users of it may not have the ability to adjust the type/field names in the schemas they're talking to. For example, for a type defined like:
    ```graphql
    type Person {
      enjoys: [Topic]
      maintains: [Repository]
    }
    ```
    It might be confusing to call the type of that field a `Person_enjoy` or `Person_maintain`. Or in the example earlier, perhaps `bazzes` is a verb relationship and there is no such thing as a "baz".

An alternative idea could be to append to the type name, instead:
```ts
type HeroName_hero_Human_friends$Element = {
 // …
}
```
(I don't know whether this is recommended practice, but [flow's core library definitions](https://github.com/facebook/flow/blob/fd719fcb856989c81b692566dfb62cbea502e6ff/lib/core.js#L110) use the dollar sign as a separator for associated types. That could also fix the ambiguity with badly named schema fields like `hero_Human` and `Human_friends`, since `$` isn't valid field name in GraphQL.)

Just a thought. It's really awesome to see so much progress on this target!

lewisf:
@erydo 

thanks for the detailed response! i totally get what you mean -- and I agree it makes sense to not singularize.

other folks have suggested doing something similar with suffixing it with `Element` or `Item`. That sounds good to me. I am hoping we can follow "recommended practice" here. 

@leoasis @rosskevin I know y'all have chatted about this in the apollo slack a bit -- any thoughts to add?

rosskevin:
I am definitely **for** singularizing - at least as I see it using `typescript-modern`.  
- All generated files are localized (directory-wise) to the usage/query/mutation - so I see no problem needing to find the source of the generated interfaces/types
- Assumption of singular/plural could be a problem for a project that is inconsistent, but most projects are and at least 80% of the time (I would say closer to 99%) this is a good thing that makes more sense.
- I am against a suffix such as `Item`, this is just longer and altering a current name further.  If I use `ShoppingCartItem` it would then generate `ShoppingCartItemItem`.

So - cost/benefit - this is good for almost everyone - and not good for a very few.

I suggest moving forward as-is and if it ends up being a problem, adding a cli option to `--omitPluralization` (or that could be added now).

I definitely would like to see singularization merged - this is common and has worked well/consistently in stacks like ruby on rails.

erydo:
@rosskevin, I hope you're enjoying the holiday weekend! I'm sorry, but I still disagree. Hopefully this is a helpful discussion. I noticed my reply somehow turned into a big wall of text, so I put some separator headings in.

### Searchability/File Locations
> All generated files are localized (directory-wise) to the usage/query/mutation - so I see no problem needing to find the source of the generated interfaces/types

That's the current behaviour, but there's an open issue regarding where the `*-modern` targets place their outputs and why that's problematic for some practical use cases: https://github.com/apollographql/apollo-codegen/issues/322

### Schema Consistency
> Assumption of singular/plural could be a problem for a project that is inconsistent, but most projects are and at least 80% of the time

I'm sorry, but that doesn't seem to address the examples above. Schemas with adjacent singular and plural fields _will_ have problems with this (which includes the canonical Relay identifying root fields!). And outside of that, the issue isn't just about consistency within a particular schema. Even an internally-consistent schema with verb relationship names would have bad output here. 

**Design Goal:** Since apollo-codegen is a _client_ tool, not primarily targeted at the schema designers, I'd suggest that maintainers should consider a design requirement that users _should not need to change the schema_ to be able to use apollo-codegen.

### Suffix Collision
> I am against a suffix such as Item, this is just longer and altering a current name further. If I use ShoppingCartItem it would then generate ShoppingCartItemItem

That's a helpful example, and you're right: You wouldn't want `[ShoppingCart]` and `ShoppingCartItem` to have the same typedef names. My suggestion above was to to suffix with `$Element` or `$Item` to unambiguously demarcate from the field/fragment/selection name. I think your examples are a good argument against tacking on a suffix without a delimiter.

(Just a note, because I had to check: Although most types generated by this codegen don't end in a typename, but in a field name; like `HeroName_hero_Human_friends`; the interface names _can be_, for example see `HeroName_hero_Human_friends_Human` in the tests).

### Prior Art
> this is common and has worked well/consistently in stacks like ruby on rails.

Unless my memory is wrong, that's not what Ruby and Rails or Django do: They _pluralize_ given _type_ names, they don't attempt to _singularize_ arbitrary _field_ names. What they do is more reasonable for both of those differences: pluralization is a far more reliable inflection than singularization, and model names are always nouns ("this is an instance of a _"); field names are not.

I would like to see the array fix merged, but strongly believe that the singularization change is a step backwards.

rosskevin:
Let's put it this way: it's an option that many will choose to use.  Just as we shouldn't rule out the few use cases, we also shouldn't rule out the many. I see no reason this should not be integrated for those that want to use it, which will be greater than 80% of the population. 

We can always find a reason not to, but we should find reasons to accommodate most without alienating the few.

erydo:
I didn't mean to seem like I'm looking for holes to poke in something. There are specific technical reasons not to attempt to interpret field names. "I want this and it'll work for 80% of the population" is not a technical rebuttal and it's frustrating that we seem to be talking past each other.

From your earlier example:
>  If I use `ShoppingCartItem` it would then generate `ShoppingCartItemItem`

The structural naming scheme of this target does not work that way.

This is about the attempt to interpret _field names_ as nouns, apply a grammar inflection, and use that in the generated type names. Not about taking schema types and inflecting them. So if you had:
```graphql
type ShoppingCart {
  veggies: [Vegetable]
  items: [ShoppingCartItem]
}
```
And a query:
```graphql
query GetShoppingCart {
  cart {
    veggies { name }
    items { name }
  }
}
```
It currently would generate `GetShoppingCart_cart_items` and `GetShoppingCart_cart_veggies` types. With singularizing the field names it creates `GetShoppingCart_cart_veggy` (_sic_). The only time your type name gets put in there is if there's a further fragment, and even then it's still not helping, because you're still going to get `GetShoppingCart_cart_item_ShoppingCartItem`.

Are you arguing that `GetShoppingCart_cart_item_ShoppingCartItem` is a so much more usable than `GetShoppingCart_cart_items_ShoppingCartItem` or `GetShoppingCart_cart_items_ShoppingCartItem$Element` that it outweighs the technical issues of that grammar transformation?

It shouldn't be exposed as an option; that's just permutation debt. It shouldn't even be included. It accommodates _zero_ additional use cases  but it prevents working behaviour in basic scenarios.

Anyway, it's Christmas, so I'm signing off—have a good one.

leoasis:
I think I agree with @erydo here. As much as I would like to singularize types whenever they make sense, and somehow avoid the cases where it doesn't, there will still be cases where the generated type name will be unintuitive for the user, or will require some reverse engineering in their minds to know what the tool tried to do when converting the type names. It won't work for the 100% of the use cases.

On the other hand, his proposed idea of adding a _delimited_ suffix, it may not read as nice as a human generated type name, but it works for 100% use cases, and you can understand what it refers to once you know the rule, since it will apply the same way all the time. It even works for fields named in different languages or fields that are not actual words. The only downside of this approach has to do with style (maybe people don't like a $ delimiter), but I think that is subjective and is something people can adapt to, and I think it's lower priority compared to consistency, understandability and working for 100% use cases.

leoasis:
One thing to take into account though (and this may lead to some bikeshedding), is maybe to consider the suffix to be _short_ so that we don't have long type names very quickly, maybe instead of `$Element` have something like `$El`

rosskevin:
I'd like a custom mapping ability.  Even with the long-unchanged names, I end up aliasing them if I need to reference them.  Given that I'm manually renaming them, perhaps we should just provide an easy way for a user to re-map the generated names - which solves the problem for anyone.  

This allows me to map:
- `ContactListQuery_currentUser_contacts_addresses` -> `Address`

Then import/use it.  We could add these mappings as comments in a graphql file, or provide global mappings.  I'd prefer to keep the mapping local to the graphql file, as I can see global name conflicts already (e.g. same query tree but different selected fields).

erydo:
You already have a custom mapping ability:
```ts
export type {
  ContactListQuery_currentUser_contacts_addresses as Address,
} from '../generated/ContactListQuery';
```

If you create and mix incompatible type aliases, then Flow will warn you that they're incompatible. If those type names are hand-chosen anyway, it's not going to save anyone any effort to add a codegen DSL into query comments.

Defining generated type names in the query files would be harder than it sounds, since the types of a selection set's fields depends on the set of refinements that reach them, and on sibling fragments, which is why there's a structured naming scheme to begin with.

rosskevin:
@erydo I use typescript but type aliases are the same.  

I'm trying to brainstorm a solution that keeps unreadable types out of our coding environment (think IDE intellisense) and also solves this pluralization problem which is quite confusing in the above example I provided. 

So my goal would be: the generated code could be adapted by the user to look like code I would write.  

This seems feasible as the mappings would be gathered/localized per graphql file, and before a type/interface is emitted, the re-mappings are consulted and if not found, the original is returned.

lewisf:
Wow a lot of discussion here. Great -- I think it's important that we get this right.

The singularization issue with @erydo brought up happens often enough that we shouldn't singularize automatically. I just looked at my schema and found a bunch of cases that look like this:

```
type CourseItem {
  state: CourseState!,
  states: [CourseState!]!
  ...
}

type Enrollment {
  type: 'PREMIUM',
  types: ['PREMIUM', 'FREE']
}
```

Anything that is similar to "state" in how we use it will run into this issue: *model*, *node*, *shape*, *type*, etc. this could be solved if schema designers consistently used "current*" to represent the existing thing or "possible*" to represent the possible things. Unfortunately, as previously stated by @erydo client developers generally don't have good control over it, and when it breaks, it breaks horribly.

cc @martijnwalraven I think there's singularization happening in the other targets, have you run into this issue before with the swift target?

---

Something else worth considering is what happens when the types get even more nested.

For instance while this might look fine.

```
type HeroName_hero_Human_friends$Element = {
 // …
}
```

It starts to look weird here:
```
type HeroName_hero_Human_friends$Element_friends$Element_friends$Element = {
  //...
}
```

But maybe this is still fine because ^ is a rare occurrence.

One way `singularization` could work in our case is to singularize unless two field types are going to end up having the same name, in which case $El. While this might also be okay, it ends up being more logic that we'd have to maintain (and might also be a weird experience).

Keep your thoughts coming :) I'm curious to see if we can come up with something that feels right.

jbaxleyiii:

Daniellt1:
Is this issue expected to be fixed anytime soon, please?

PR #345, opened 2017-12-29
Fix Flow-Modern incorrectly-flattened array generation

erydo:
Fixes #325 (_Flow Modern flattens Arrays in output_)

This is an alternative implementation versus #342 which:
  * Provides full support of arbitrary nullable/nested lists without special casing top-level lists/nulls. (The `typeAnnotationFromGraphQLType` function now recursively covers all cases; the type annotations for field properties are now correctly generated from this same logic as well).
  * Does not include the inferred inflection change discussed in that PR, which is unrelated to the array-flattening issue

PR #346, opened 2017-12-29
Flow-Modern covariant properties/$ReadOnlyArray

### LINKED ISSUE #129: Add an option to make all properties covariant

### LINKED ISSUE #135: flow: Consider using $ReadOnlyArray instead of Array

erydo:
This is based on the same branch as #345 (_Fix Flow-Modern incorrectly-flattened array generation_).

Fixes #129 (_Add an option to make all properties covariant_)
Fixes #135 (_flow: Consider using $ReadOnlyArray instead of Array_)

This is like #163, except for the `flow-modern` target.


erydo:
This PR has been rebased on current master, since #345 was merged in.

erydo:
@lewisf @martijnwalraven — I'd love to get your review on this, whenever possible. Fully immutable types would be a huge benefit in our codebase. It enables much deeper type inferences.

Because this change is opt-in, it should be very low-impact to existing users.

erydo:
Awesome! Yes, I agree about the awkwardness of the flags, but wanted to keep it consistent with what's there.

It's a challenge to decide which of those flags should be safely shared between targets, when semantics vary subtly between the languages. Does the concept of "readonly" map similarly between languages? I'm not certain how TypeScript even represents "exact" types, for example

Read-only properties are a slightly different guarantee than covariant ones, in that read only properties aren't necessarily static (e.g. computed properties), covariant properties aren't necessarily immutable. Some languages support copy-on-write; structural v. prototypical v. hierarchical subtyping rules, and so on. It could be difficult to nail down shared flags without mis-representing what they guarantee in a particular target language.

But looking forward to usability cleanups along those lines, in any case.

lewisf:
I'm going to merge this in for now so the feature exists while we work on figuring out the best way to split out targets and their cli args

Issue #321, opened 2017-12-16
Flow Modern target does not create types for operation arguments


erydo:
In this example:
```graphql
mutation DoFoo($someId: ID!, $someText: String, $clientMutationId: String) {
  doFoo(input: {
      someDefault: "FOO",
      id: $someId,
      text: $someText,
      clientMutationId: $clientMutationId}) {
    someResult
  }
}
```

The old `flow` target would generate a `DoFooVariables` type including `someId`, `someText`, and `clientMutationId`.

That appears to be missing from the `flow-modern` target. In `flow-modern`, only referenced mutation input type are generated (as in—if there is an additional input type referenced in the mutation), independent of the arguments needed for the operation.

This is related to:
  * (#318) _Feature request: flow types incorporating operation name and associated variables_
  * (#320) _Flow Modern does not export "input" types_

lewisf:
Thanks for filing! I'll take a look at implementing this when I get the chance.

erydo:
This was closed in #348

PR #348, opened 2018-01-03
*-Modern variables

leoasis:
This PR adds the generation of the variables object type for each operation. In the case of flow it generates a flow annotation, and in the case of Typescript, an interface. It follows the same nullability rules as when creating input objects (they can be set to undefined or not even set).

### DIFF
--- a/src/javascript/flow/codeGeneration.ts
+++ b/src/javascript/flow/codeGeneration.ts
@@ -154,6 +154,7 @@ export class FlowAPIGenerator extends FlowGenerator {
     const {
       operationType,
       operationName,
+      variables,
       selectionSet
     } = operation;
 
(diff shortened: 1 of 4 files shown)


lewisf:
Ah -- happy that it wasn't terrible hard to do. Looks good, thanks @leoasis !

leoasis:
Indeed! Very straightforward!

src/javascript/flow/codeGeneration.ts
```diff
@@ -178,6 +179,19 @@ export class FlowAPIGenerator extends FlowGenerator {
 
     this.printer.enqueue(exportedTypeAlias);
     this.scopeStackPop();
+
+    // Generate the variables interface if the operation has any variables
+    if (variables.length > 0) {
```
erydo: Suggestion: Is there a good reason to skip this type definition here, versus creating an empty object type or null type? Omitting a definition here prevents other code generators from consistently importing `*Query, *Variables` types for each operation without having to do the same kind of check.

It would be helpful in those cases to have a type generated for consistency, like:
```ts
export type SomeStaticQueryVariables = {};
```

erydo:
Looks like this closes #321 (_Flow Modern target does not create types for operation arguments_).

src/javascript/flow/codeGeneration.ts
erydo: (Added a PR to implement as a compiler option: #350)

PR #349, opened 2018-01-04
Fix Lists for TS

lewisf:
Mirror #345 for Typescript-Modern target

### DIFF
--- a/src/javascript/typescript/language.ts
+++ b/src/javascript/typescript/language.ts
@@ -13,7 +13,7 @@ import {
   sortEnumValues
 } from '../../utilities/graphql';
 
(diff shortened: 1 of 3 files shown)


src/javascript/typescript/language.ts
```diff
@@ -29,7 +28,7 @@ export interface TypescriptCompilerOptions extends CompilerOptions {
 
 export default class TypescriptGenerator {
   options: TypescriptCompilerOptions
-  typeFromGraphQLType: (graphQLType: GraphQLType, options?: TypeFromGraphQLTypeOptions) => t.TSType
+  typeFromGraphQLType: Function
```
leoasis: Do we want to type this with a stricter signature? It will help catch errors if we refactor it in the future.

lewisf: Good point. Let's do that

PR #350, opened 2018-01-07
Add option to always generate *Variables type definition

erydo:
Adds the `--always-generate-variables` cli flag and `alwaysGenerateOperationVariables` compiler option to generate `*Variables` definitions for all operations, regardless of whether a given operation require any variables. This facilitates consistent type imports per operation.

The option defaults to `false` for backwards compatibility, however has been enabled in some of the tests+snapshots to demonstrate that it's working.

Noticed during review here:
https://github.com/apollographql/apollo-codegen/pull/348#discussion_r160053935

> Omitting a definition prevents other code generators from consistently importing `*Query, *Variables` types for each operation without having to do the same kind of check.
> 
> It would be helpful in those cases to have a type generated for consistency, like:
> ```ts
> export type SomeStaticQueryVariables = {};
> ```

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -127,6 +127,11 @@ yargs
         describe: 'Optional namespace for generated types [currently Swift and Scala-only]',
         type: 'string'
       },
+      "always-generate-variables": {
+        demand: false,
+        describe: "Always generate operation variables type declarations, even if there are no variables",
+        default: false
+      },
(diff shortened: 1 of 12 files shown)


erydo:
Looks like I screwed up some trailing commas in the original commit; should be better now.

jbaxleyiii:

PR #361, opened 2018-01-18
Remove `mkdirp`.

wtgtybhertgeghgtwtg:
It doesn't seem to be used.


lewisf:
thanks for catching this!

PR #363, opened 2018-01-18
Support semicolons in header values

planser:
Header values are currently corrupted if they contain semicolons (introspect-schema and download-schema commands). As of https://tools.ietf.org/html/rfc7230 semicolons in header values are allowed.


lewisf:
thanks for addressing this!

PR #370, opened 2018-02-01
Log errors to stderr instead of stdout

Caerbannog:
Replace all occurrences of console.log by console.error in logError functions.
This allows using apollo-codegen with other commands in a script, and separating errors from normal output.
A work-around would be to use shell redirections.

I have not written any test in the PR.


lewisf:
thanks @Caerbannog!

PR #371, opened 2018-02-12
Fix the existence of the wrong schema judgment

choujimmy:
Like the following response of graphql server:
```json
{ 
  data: {},
  errors: [],
  extensions: null
}
```

The `errors` field may be an empty array.


martijnwalraven:
Is this something you've encountered? According the the GraphQL spec, `errors` should never be empty:
> The errors entry in the response is a non‐empty list of errors, where each error is a map.
> If no errors were encountered during the requested operation, the errors entry should not be present in the result.
http://facebook.github.io/graphql/October2016/#sec-Errors

jbaxleyiii:

PR #372, opened 2018-02-12
Move repository to mono-repo structure

lewisf:
This introduces yarn workspaces + lerna

TODO: 
- [x] Introduce lerna (do we want dependent or independent versioning?)
- [x] I moved things into packages roughly based on logical groupings
- [ ] Double check if the new packages are good
- [x] Moved tests to be colocated under `__tests__` dirs
- [x] Most tests pass  (with the exception of a couple of things that need fixing)
- [x] All tests pass
- [x] tsc compile works without errors
- [ ] development experience is smooth (probably make sure npm scripts work as expected)
- [ ] packages are publishable and `apollo-codegen` as a command is still backwards compatible

cc @martijnwalraven @evans


lewisf:
@martijnwalraven, I'm running into this test failure:

image

And I'm not quite sure if this is an actual failure or if something changed elsewhere?

lewisf:
Running into some issues for getting this to work, namely having each independent project having their own `outDir`. it looks like there are issues on the typescript project related to this, namely: https://github.com/Microsoft/TypeScript/issues/3469

martijnwalraven:
Hmmm, this turn out to be a lot more complicated than I thought! I found [this example project](https://github.com/Quramy/lerna-yarn-workspaces-example) and tried to get things working using `tsconfig.json` inheritance, but that depends on having each package be independently buildable and testable, so that requires some changes. This setup seems to similar to what we use for [Apollo Client](https://github.com/apollographql/apollo-client) though.

I'll probably give this another try later today or tomorrow. Let me know what your thoughts are about the setup we want.

martijnwalraven:
Ok, this feels way too complicated! But at least I think I got things working. I followed [this example project](https://github.com/Quramy/lerna-yarn-workspaces-example) to not rely on building all packages in development and testing and load files directly from each packages's `src` directory. Building compiles to individual `lib` directory so packages can be published.

I haven't tested installing and running the CLI yet, but `yarn test` and `yarn build` work, and CI passes.

evans:
Closing since we've migrated to lerna with #432. Thank you @lewisf for the initial inspiration!

Issue #375, opened 2018-02-15
--passthrough-custom-scalars option not handled in Flow generator


asa-taka:
v0.18.3 not handle **--passthrough-custom-scalars** option in Flow generator.

Similar to https://github.com/apollographql/apollo-codegen/issues/223, I think same fix can resolve this fault too, like below.

```diff
  let typeName;
  if (type instanceof GraphQLList) {
    typeName = `Array< ${typeNameFromGraphQLType(context, type.ofType, bareTypeName)} >`;
  } else if (type instanceof GraphQLScalarType) {
-        typeName = builtInScalarMap[type.name] || (context.passthroughCustomScalars ? context.customScalarsPrefix + type.name : 'any');
+        typeName = builtInScalarMap[type.name] || (context.options.passthroughCustomScalars ? context.options.customScalarsPrefix + type.name : 'any');
  } else {
    typeName = bareTypeName || type.name;
  }
```

at https://github.com/apollographql/apollo-codegen/blob/master/src/flow/types.js#L39

I use apollo-codegen in [this repo](https://github.com/asa-taka/try-apollo-codegen-with-custom-scalar).

PR #378, opened 2018-02-16
Fix unhandled passthroughCustomScalars option in Flow generator

asa-taka:
This is fix for #375 

Should I add test for it? (I couldn't find test about `options` of Flow generator....)


shadaj:
Thanks for the fix!

PR #392, opened 2018-03-14
[Swift] Don't singularize non-list-typed field names

cysp:
The impetus for this request is a field named `sys` of remote type `Sys`, which gets turned into a struct named `Sy` because its name gets singularised.
This _may_ result in source-incompatibility, but in a good direction?

### DIFF
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -1466,6 +1466,62 @@ exports[`Swift code generation #structDeclarationForSelectionSet() should genera
 }"
 `;
 
+exports[`Swift code generation #structDeclarationForSelectionSet() should not singularize non-list properties 1`] = `
+"public struct SingularDetails: GraphQLSelectionSet {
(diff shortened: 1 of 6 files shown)


cysp:
Looks like the `apollo-cli` tests are failing on Travis because they want a `STAGING_API_KEY`?

(A bunch of engine-related tests fail on my machine too, complaining that mock responses don't exist. 😦)

jbaxleyiii:

cysp:
Is the change something that the ApolloGraphQL project wants, @jbaxleyiii ?
I updated it across the reorganisation but then stopped updating it after a few months with seemingly no interest.

mrrhysbutler:
Just stumbling on this one now, we are experiencing an issue relating to this. We would be very interested in seeing this fix or one like it go through.

TBH, i think infering pluralisation is frought with danger, i am surprised this is acceptable. The comments from @erydo here: https://github.com/apollographql/apollo-tooling/pull/342#issuecomment-353758561 seem to be pertinent.

In our case, we have a `roles` property which is an object within a query (not an array). It is being singularized to a `Role` type nested within the query. This generates code internal to the query which conflicts with an independant `Role` enum type, which is referenced within the same query.

PR #398, opened 2018-03-21
Typo on command name

chibicode:
screen shot 2018-03-20 at 11 31 03 pm

### Also

> If you're using a client like apollo-client that **does** this automatically for your GraphQL operations

Is **does** the correct word here, as opposed to **doesn't**?


chibicode:
Already signed!

abernix:
Thanks @chibicode!

---

Refs:

https://github.com/apollographql/apollo-codegen/blob/d15219b181257de0a834f8cfddbea5cc975cbc8e/src/cli.js#L143-L147

PR #399, opened 2018-03-22
Emit JSDoc comments for TypeScript types and enums

### LINKED ISSUE #384: Output JSDoc for field descriptions in TypeScript
Currently, a generated TypeScript interface looks like this:
```ts
export interface CurrentUserQuery {
  // The currently authenticated user.
  // Returns null if the user is not authenticated.
  currentUser:  {
    // ...
  }
}
```
If instead apollo-codegen would use the JSDoc `/** ... */` syntax, editors like VS Code will show it as part of autocomplete/intellisense. Currently the very helpful descriptions from the schema are not surfaced in any meaningful way.

The above would need to change to:
```ts
export interface CurrentUserQuery {
  /**
   * The currently authenticated user.
   * Returns null if the user is not authenticated.
   */
  currentUser: {
    // ...
  }
}
```

TedDriggs:
Many IDEs that work with TypeScript will use JSDoc comments on types, fields, and enum variants in hover and completion contexts. Changing the generated code to wrap the `description` property in a JSDoc comment instead of a regular comment or postfix comment makes those IDEs surface the GraphQL description to the developer as they're working.

Fixes #384

### DIFF
--- a/src/typescript/codeGeneration.ts
+++ b/src/typescript/codeGeneration.ts
@@ -15,10 +15,6 @@ import {
 } from 'graphql'
 
 import {
-  wrap
-} from '../utilities/printing';
-
-import {
   sortEnumValues
 } from '../utilities/graphql';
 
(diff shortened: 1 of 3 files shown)


fbartho:
This looks really good IMHO. Hopefully somebody can review it soon!

PR #402, opened 2018-03-25
[Scala] Fix naming conflicts and update API references

### LINKED ISSUE #357: [BUG] Scala codegen: Namespace collision from inner case classes w/ different types
For the `ScalaJS` codegen, when generating a class from this query:

```graphql
query SomeQuery {
  filters {
    make {
      values { .
        id .     # this id is of type int in schema
      }
    }
    model {
      values {
        id .    # this id is of type string string
      }
    }
}
```

The generated classes have collisions due to the `Value` class being put on the root level (which actually have different type signatures)

```scala
...

  object Make {
    val possibleTypes = scala.collection.Set("MakeFilter")
  }

  case class Value(id: Int) {
  }

  object Value {
    val possibleTypes = scala.collection.Set("MakeValue")
  }

  case class Model(values: Seq[Value]) {
  }

  object Model {
    val possibleTypes = scala.collection.Set("ModelFilter")
  }

  case class Value(id: String) {
  }

  object Value {
    val possibleTypes = scala.collection.Set("ModelValue")
  }
```

The same problem doesn't ocurr for TypeScript (didn't check Swift). 

So the inner classes would need to be nested inside the top object to avoid "same-key" namespace collisions

shadaj:
Fixes #357 by placing field types inside the companion object of the type defining the field's parent. Also fixes a bug with package declarations not being generated due to the `namespace` value being moved from `context` into `context.options` (in the future, I hope to switch the generator to TypeScript to prevent similar issues).

A lot of references to Slinky APIs were out of date, so also this updates the generator to use the latest APIs.


src/scala/codeGeneration.js
```diff
@@ -60,7 +60,7 @@ export function generateSource(context, options) {
   generator.printOnNewline('//  This file was automatically generated and should not be edited.');
   generator.printNewline();
 
-  if (context.namespace) {
+  if (context.options.namespace) {
```
joseph-whiting: Would a test be good for this? Causes a headache when it breaks! Happy to commit to this PR and see what you think?

shadaj: Got some free time today, so I went ahead and converted everything to TypeScript and added a test for adding the package declaration!

joseph-whiting: Nice!

Issue #308, opened 2017-11-30
Generated Flow/TypeScript doesn't include the actual queries


jesseschalken:
### Problem

At the moment running `apollo-codegen` with `--target` of either `flow`, `flow-modern`, `typescript` or `typescript-modern` generates only type definitions, but the generated code doesn't include the GraphQL queries themselves, leaving the programmer to get their GraphQL queries into their code by some other means (eg `gql` tagged templates, importing the parsed contents of `.graphql` files with `graphql-document-collector` or `webpack-graphql-loader`, etc).

This means the programmer has to correlate the query itself with the types generated by `apollo-codegen` manually, and could get it wrong.

Example:

```graphql
# getUser.graphql

query getUser($id: Int!) {
  user(id: $id) {
    id
    name
    email
  }
}
```

`apollo-codegen` produces:

```typescript
// graphql_types.ts

export type getUser = {
  id: number,
  name: string,
  email: string,
}

export type getUserVariables = {
  id: number
}
```

The programmer can write something erroneous like this, and it will still type check.

```typescript
// some_code.ts

let apollo = new ApolloClient(...)

// This isn't the right type!
type User = {
  firstName: string,
  lastName: string,
}

function getUser(username: string): Promise<ApolloQueryResult<User>> {
  return apollo.query({
    query: require("./getUser.graphql"),
    // getUser() takes an id, not a username!
    variables: {username}
  }) 
}
```

This is because:

1. `ApolloClient.query` has type `query<T>(options: WatchQueryOptions): Promise<ApolloQueryResult<T>>`, which means the output can be used as whatever type the caller wants.
2. `WatchQueryOptions.variables` has type `{ [key: string]: any }`, so the caller can pass whatever variables they want.
2. `WatchQueryOptions.query` has type `DocumentNode`, so the caller can pass whatever GraphQL document they want.

There is no static typing to make sure these three things actually match.

Is there a solution to this? Maybe I am not using `apollo-client` and `apollo-codegen` correctly?

### Solution?

I would have expected the types in `apollo-client` to be something like:

```typescript
class Query<Tvars, Tret> {
  constructor(public document: Document) {}

  // Dummy method to force TypeScript to pay attention to the Tvars and Tret types
  private __nominal(x: Tvars): Tret { return 1 as any }
}

export interface WatchQueryOptions<Tvars, Tret> {
  variables?: Tvars
  query: Query<Tvars, Tret>;
  // ...
}

class ApolloClient {
  // ...
[...]

PR #414, opened 2018-04-19
Generate operation functions that execute typechecked query & mutations

infogulch:
This is a PoC solution to [Generated Flow/TypeScript doesn't include the actual queries](#308). It generates what I've been writing manually, which is a typed function that accepts the apollo client and the generated variables interface as parameters and returns the generated result interface.

It's a proof-of-concept because it's directly tied to `apollo-client`, and (even worse) specifically `apollo-angular`. Unsurprisingly I don't expect this PR to be accepted as-is, **the idea is to have something concrete to start talking about** in relation to #308. That said, I'd be more than happy to take suggestions to make this more worthy.

All tests pass except 16 typescript snapshot tests, because the generated code changes (obviously) and I didn't want the PR diffs to be bogged down by changes to test scripts at this point.

It generates code like this: (taken from test output)

```
import gql from 'graphql-tag';
import { Observable } from 'rxjs/Observable';
import { ApolloQueryResult } from 'apollo-client';
import { Apollo } from 'apollo-angular';
import { FetchResult } from 'apollo-link';

export function CustomScalar(
apollo: Apollo,): Observable<ApolloQueryResult<CustomScalarQuery>> {
  return apollo.query<CustomScalarQuery>({
    query: gql`query CustomScalar {
  commentTest {
    __typename
    enumCommentTest
  }
}`,
    variables: variables,
  });
}
```

```
export function ReviewMovie(
apollo: Apollo,
variables: ReviewMovieMutationVariables,): Observable<FetchResult<ReviewMovieMutation>> {
  return apollo.mutate<ReviewMovieMutation>({
    mutation: gql`mutation ReviewMovie($episode: Episode, $review: ReviewInput) {
  createReview(episode: $episode, review: $review) {
    __typename
    stars
    commentary
  }
}`,
    variables: variables,
  });
}
```

Some changes this PR needs imo:
- [ ] Remove reliance on apollo-client, apollo-angular (especially), and apollo-link. This could just be generating accepted interfaces inline like @ecstasy2 suggested, or some other idea, suggestions welcome.
- [ ] Make sure it's not possible to support this without importing `rxjs/Observable` and `graphql-tag`. I'd prefer to reduce imports.
- [x] Make it work with fragments (may be as simple as using `LegacyOperation.sourceWithFragments` instead of `.source`).
- [ ] This adds the `__typename` parameter to the source, which may cause problems. Some feedback on how this should work would be great.
- [ ] Generate prettier-formatted code.
- [ ] Add tests that actually use the generated code, and update the existing snapshot tests.
- [ ] Flow support could be a separate PR since I'm not familiar with it, but the decisions in this PR would probably help a flow implementation.... *flow* through the PR process more easily.

Feedback and suggestions welcome!

### DIFF
--- a/src/typescript/codeGeneration.ts
+++ b/src/typescript/codeGeneration.ts
@@ -26,6 +26,7 @@ import CodeGenerator from '../utilities/CodeGenerator';
 
 import {
   interfaceDeclaration,
+  functionDeclaration,
   propertyDeclaration,
   propertySetsDeclaration,
   Property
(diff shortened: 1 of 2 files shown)


infogulch:
After investigating this over the last couple of days I've come to the conclusion that this is fundamentally the wrong approach. I'm closing this and will be opening a ticket on apollo-client and/or apollo-angular instead.

PR #415, opened 2018-04-20
Update graphql-config

victorandree:
graphql-config@2.0.1 properly installs graphql as a peer dependency, which avoids issues with loading schemas using different GraphQL versions.

This should also fix https://github.com/apollographql/apollo-codegen/issues/406


mike-marcacci:
It would be nice to publish this, since it was merged so long ago.

PR #422, opened 2018-05-24
fix #421 - generate $Exact InputObject annotations

### LINKED ISSUE #421: InputObject annotations should always be $Exact in flow/flow-modern
/label bug
/label flow

According to [the graphql spec](https://github.com/facebook/graphql/blob/master/spec/Section%205%20--%20Validation.md#input-object-field-uniqueness) and the reference implementation, an operation is considered invalid if any variables have fields in addition to those in the corresponding `InputObject` definition.

Currently, `flow-modern` generates `InputObject` types *without* `$Exact` criteria, and doesn't provide an option to enable it (even when using exact types for output types [which I personally avoid]).

I believe that the correct behavior is to *always* generate exact type annotations for `InputObject`s.

It may be the case that other generators are missing this too, but I've only noticed this with `flow-modern`.

mike-marcacci:
/label bug
/label flow

This fixes #421 by ensuring that all `InputObject` annotations are `$Exact` when using the `flow-modern` generator.

EDIT: I accidentally included the "typescript" label, and I don't see a way to remove it. If someone could remove that for me, it would be appreciated!

### DIFF
--- a/src/javascript/flow/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/src/javascript/flow/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -628,18 +628,18 @@ export type HeroNameVariables = {
 export type Episode = \\"EMPIRE\\" | \\"JEDI\\" | \\"NEWHOPE\\";
 
 // The input object sent when someone is creating a new review
-export type ReviewInput = {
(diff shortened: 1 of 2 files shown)


shadaj:
Thanks for the fix!

Issue #320, opened 2017-12-16
Flow Modern does not export "input" types


erydo:
https://github.com/apollographql/apollo-codegen/blob/8c46ca32e6971884a716f4973c5539745f8f3115/src/javascript/flow/__tests__/__snapshots__/codeGeneration.ts.snap#L815-L820

These should be annotated like `export type ReviewInput`

lewisf:
this should be as easy as changing the ast that makes up this node to be wrapped with an export statement. was an oversight on my part.

accepting PR's for this! otherwise i can try to get to it the next time i get time to work on this

lewisf:
Fixed in #340 @martijnwalraven let's schedule a release!

PR #411, opened 2018-04-10
Export input types for typescript-modern

blocka:
This was down for flow-modern but not typescript

PR #424, opened 2018-05-29
[ts-modern] export generated input types

nilshartmann:
In `ts-modern` the generated input types are not exported and thus not usable from outside the generated module. (Similiar issue in Flow was #320 ). 

This PR adds the missing 'export's to the types.

*Example **before**:*
```
//==============================================================
// START Enums and Input Objects
// All enums and input objects are included in every output file
// for now, but this will be changed soon.
// TODO: Link to issue to fix this.
//==============================================================

// ===> SearchFilter is an input type and not exported:
interface SearchFilter {
  searchString?: string | null;
  maxHits?: number | null;
}
```

*Example **after**:*
```
//==============================================================
// START Enums and Input Objects
// All enums and input objects are included in every output file
// for now, but this will be changed soon.
// TODO: Link to issue to fix this.
//==============================================================

// ===> SearchFilter is an input type and with this PR now exported:
export interface SearchFilter {
  searchString?: string | null;
  maxHits?: number | null;
}
```

### DIFF
--- a/src/javascript/typescript/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/src/javascript/typescript/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -544,14 +544,14 @@ export enum Episode {
 }
 
 // The input object sent when someone is creating a new review
-interface ReviewInput {
+export interface ReviewInput {
   stars: number;
   commentary?: string | null;
   favorite_color?: ColorInput | null;
 }
 
 // The input object sent when passing in a color
-interface ColorInput {
+export interface ColorInput {
   red: number;
   green: number;
   blue: number;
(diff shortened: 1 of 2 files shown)


smkhalsa:
duplicate of #411? In any case, I'd love to see one of these merged as this is a blocker for me.

teijo:
Hack in the meantime (`find`/`sed` flags might vary on GNU/Linux): `apollo-codegen generate '**/*.ts?(x)' --addTypename --schema ../../generated-graphql-schema.json --tag-name gql --target typescript-modern && find . -regex '.*/__generated__/.*.ts' -exec sed -i '' -e 's/^interface/export interface/g' {} \;`

lirbank:
Another benefit of exporting all types is that you can enable `noUnusedLocals` TS compiler option if you like. Without exporting them, the compiler will complain since they are part of all generated files (and not used in all). I'm using `--target typescript-modern`.

shadaj:
Thanks for the fix!

PR #437, opened 2018-06-18
Release v0.20.0

shadaj:


mrm007:
@shadaj There's a typo in the arg name for `flow-legacy`:

https://github.com/apollographql/apollo-codegen/blob/v0.20.0/packages/apollo-codegen/src/cli.ts#L128

PR #442, opened 2018-06-22
[cli, flow-legacy] Fix CLI typo for flow-legacy

kristiehowboutdat:
This typo breaks backwards compatibility with `flow` users from `0.19.0` because we can no longer use `flow-legacy` since you can't provide the right CLI argument and the switch statement looks for the correct argument (`flow-legacy`) instead of the only one you can put on the CLI (`flow-leagcy`)

### DIFF
--- a/packages/apollo-codegen/src/cli.ts
+++ b/packages/apollo-codegen/src/cli.ts
@@ -125,7 +125,7 @@ yargs
       target: {
         demand: false,
         describe: 'Code generation target language',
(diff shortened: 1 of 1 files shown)


kristiehowboutdat:
👋 The Meteor CLA has been signed. What's the process for getting changes like this in? This is a breaking change for 0.20 as the `flow-legacy` cannot be run, and the `flow` (formerly `flow-modern` is broken with `__typename` per #443. This is a critical part of my team's workflow (thank you!). Happy to do whatever I can to help.

shadaj:
👋 The changes look great, thanks for the patch!

shadaj:
@kristiehowboutdat released in version 0.20.1!

PR #445, opened 2018-06-25
Use block comments in modern targets to better handle multiline comments

### LINKED ISSUE #441: Bug/Regression: [0.20.x] Multi-line Doc-strings get generated as a single-line comment for typescript
When you have a multi-line doc-string like:

```graphql
"""
This line is safe
any
further lines are not.
"""
input TestInput {
	id: ID
}
```

It gets code-generated to this:

```typescript
// This line is safe
any
further lines are not.
export interface TestInput {
  id?: string | null;
}

```
- [x] has-reproduction
- [x] blocking

shadaj:
Fixes #441


packages/apollo-codegen-core/src/utilities/printing.ts
```diff
@@ -31,3 +31,13 @@ export function wrap(start: string, maybeString?: string, end?: string) {
...
   return maybeString && maybeString.replace(/\n/g, '\n  ');
 }
+
+/**
+ * Generates the body of a JSDoc style comment
+ */
+export function commentBlockContent(commentString: string) {
+  return '*\n' + commentString
+    .split('\n')
+    .map(line => ` * ${line}`)
```
erydo: Although unlikely to occur, possibly a good idea to replace `*/` to avoid prematurely terminating the block.

fbartho:
Looking pretty good to me!

PR #447, opened 2018-06-26
Migrate in Apollo CLI and add codegen commands

### LINKED ISSUE #449: [0.20.x] Output generated Types next to the source files (ts-modern behavior)
It's nice so get the flexibility to output types in any directory with `0.20.0+`.
But especially for projects which where using the previous behavior but also for projects with lot's of queries I liked the way the `ts-modern` target placed the types next to the query files.

Do you plan to support the old "types next to source" approach again?

### LINKED ISSUE #450: [0.20.x] tidy command line error output when --output is omitted
Before 0.20.x if you omitted the `--output` param the type files were colocated with the source files, but now apollo-codegen throws an error.

Colocate with source files, multiple files (omit the `--output` flag) - does not work anymore:
```
apollo-codegen generate \"src/**/*.{ts?(x),graphql}\" --schema schema.json --target typescript-modern --addTypename --tag-name gql
```

The above outputs `TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be one of type string, Buffer, or URL. Received type undefined`

Single file - works:
```
apollo-codegen generate \"src/**/*.{ts?(x),graphql}\" --schema schema.json --target typescript-modern --addTypename --tag-name gql --output src/__generated__/operation-result-types.d.ts
```

Multiple files - works:
```
apollo-codegen generate \"src/**/*.{ts?(x),graphql}\" --schema schema.json --target typescript-modern --addTypename --tag-name gql --output src/__generated__/
```

Not sure if `--output` is required going forward or if this is a regression?

Possibly related to #449 and #448

shadaj:
Fixes #449 
Fixes #450


packages/apollo-cli/src/commands/codegen/generate.ts
```diff
@@ -0,0 +1,209 @@
...
+    }),
+    schema: flags.string({
+      description: "Path to your GraphQL schema introspection result",
+      default: "schema.json",
+    }),
+
+    key: flags.string({
+      description: "The API key for the Apollo Engine service",
+    }),
+    tag: flags.string({
```
jbaxleyiii: lets pull back tag for right now since we don't have that concept in Engine quite yet

packages/apollo-cli/src/commands/codegen/generate.ts
```diff
@@ -0,0 +1,209 @@
...
+      description: "The tag of the registered schema to get from Apollo Engine"
+    }),
+    engine: flags.string({
+      description: "Reporting URL for a custom Apollo Engine deployment",
+      hidden: true,
+    }),
+
+    target: flags.string({
```
jbaxleyiii: is this needed if we are inferring?

packages/apollo-cli/src/commands/codegen/generate.ts
```diff
@@ -0,0 +1,209 @@
...
+
+            if (engineSchema.data && engineSchema.data.service.schema) {
+              ctx.schema = buildClientSchema(
+                engineSchema.data.service.schema
+              );
+            } else {
+              this.error("Failed to get schema from Apollo Engine");
+            }
+          } else {
+            const schemaFileContent = await promisify(fs.readFile)(path.resolve(flags.schema as string));
```
jbaxleyiii: fyi there is `fs.readFileSync` that you can use here

packages/apollo-cli/src/commands/codegen/generate.ts
```diff
@@ -0,0 +1,209 @@
...
+              mergeInFieldsFromFragmentSpreads: flags.mergeInFieldsFromFragmentSpreads,
+              useFlowExactObjects: flags.useFlowExactObjects,
+              useFlowReadOnlyTypes: flags.useFlowReadOnlyTypes
+            }
+          )
+        },
+      },
+    ]);
+
+    return tasks.run();
```
jbaxleyiii: It would be great for this output to give information about what was generated. Something like `x types, x operations generated`. I'm not sure what the best result would be (does `generate` return anything). In practice its always a win if the CLI shows what it just did

packages/apollo-cli/src/commands/schema/checkQueries.ts
```diff
@@ -0,0 +1,146 @@
...
+// TODO: name this check and rename check to diff?
+export default class CheckQueries extends Command {
+  static description = "Generate static types for GraphQL queries.";
+
+  static flags = {
+    help: flags.help({
+      char: "h",
+      description: "Show command help",
+    }),
+    queries: flags.string({
```
jbaxleyiii: I'm wondering if we can have a base class or something to extend with these flags in a shared place so we don't have to replicate them everywhere

packages/apollo-cli/src/commands/schema/checkQueries.ts
```diff
@@ -0,0 +1,146 @@
...
+    }),
+    schema: flags.string({
+      description: "Path to your GraphQL schema introspection result",
+      default: "schema.json",
+    }),
+
+    key: flags.string({
+      description: "The API key for the Apollo Engine service",
+    }),
+    tag: flags.string({
```
jbaxleyiii: same comment about removing tag for now

packages/apollo-cli/src/commands/schema/checkQueries.ts
```diff
@@ -0,0 +1,146 @@
...
+
+            if (engineSchema.data && engineSchema.data.service.schema) {
+              ctx.schema = buildClientSchema(
+                engineSchema.data.service.schema
+              );
+            } else {
+              this.error("Failed to get schema from Apollo Engine");
+            }
+          } else {
+            const schemaFileContent = await promisify(fs.readFile)(path.resolve(flags.schema as string));
```
jbaxleyiii: same comment about `fs.readFileSync`

packages/apollo-cli/src/commands/schema/checkQueries.ts
```diff
@@ -0,0 +1,146 @@
...
+          const warnings: GraphQLError[] = []
+
+          docs.forEach((doc, i) => {
+            const validateResult = validate(ctx.schema, doc);
+            const deprecationResult = findDeprecatedUsages(ctx.schema, doc);
+
+            errors.push(...validateResult);
+            warnings.push(...deprecationResult);
+
+            task.title = `Checking query compatibility with schema (${i + 1}/${docs.length}, ${errors.length} errors, ${warnings.length} warnings)`;
```
jbaxleyiii: oh dope!

packages/apollo-cli/src/commands/schema/checkQueries.ts
```diff
@@ -0,0 +1,146 @@
...
+          if (errors.length > 0 || warnings.length > 0) {
+            throw { errors, warnings };
+          }
+        },
+      },
+    ]);
+
+    return tasks.run().catch(({ errors, warnings }) => {
+      if (warnings.length > 0) {
+        this.warn(warnings.join("\n"));
```
jbaxleyiii: lets give these both some padding around the title (i.e. a new line before and after?

can we give the full file name?

packages/apollo-cli/src/commands/schema/download.ts
```diff
@@ -0,0 +1,109 @@
...
+    }),
+    endpoint: flags.string({
+      description: "The URL of the server to fetch the schema from",
+      default: "http://localhost:4000/graphql", // apollo-server 2.0 default address
+    }),
+
+    key: flags.string({
+      description: "The API key for the Apollo Engine service",
+    }),
+    tag: flags.string({
```
jbaxleyiii: same

packages/apollo-cli/src/commands/schema/checkQueries.ts
```diff
@@ -0,0 +1,146 @@
...
+      {
+        title: "Scanning for GraphQL queries",
+        task: async (ctx, task) => {
+          const paths = await globby(flags.queries ? [ flags.queries ] : []);
+          task.title = `Scanning for GraphQL queries (${paths.length} found)`;
+          ctx.queryPaths = paths;
+        }
+      },
+      {
+        title: pullFromEngine ? "Loading schema from Apollo Engine" : "Loading GraphQL schema",
```
jbaxleyiii: should we extract this (does it make sense) into a common util function?

packages/apollo-cli/src/git.ts
```diff
@@ -0,0 +1,55 @@
...
+import * as git from "git-rev-sync";
+import { pickBy, identity } from "lodash";
+
+const findGitRoot = (start?: string | string[]): string => {
+  start = start || module.parent!.filename;
+  if (typeof start === "string") {
+    if (start[start.length - 1] !== path.sep) start += path.sep;
+    start = start.split(path.sep);
+  }
+  if (!start.length) throw new Error(".git/ not found in path");
```
jbaxleyiii: lets pull in a commit (or manually make the change) from upstream apollo-cli here. I make a fix that didn't throw to make git not required

https://github.com/apollographql/apollo-cli/commit/28c7d298adc0d48292c8f259b1670f00d6e61050

packages/apollo-cli/src/commands/codegen/generate.ts
shadaj: Inference doesn't work when you specify a directory instead of a file, so this is a fallback.

packages/apollo-cli/src/commands/codegen/generate.ts
shadaj: I decided to use the async version so that the spinner keeps spinning while we are reading content (not major but fun 😄)

PR #457, opened 2018-06-28
Allow legacy targets to be used when manually specified

shadaj:


packages/apollo-cli/src/commands/codegen/generate.ts
```diff
@@ -81,7 +81,7 @@ export default class Generate extends Command {
 
     let inferredTarget: TargetType = "" as TargetType;
     if (flags.target) {
-      if (["swift", "typescript", "flow", "scala"].includes(flags.target)) {
+      if (["swift", "typescript", "ts", "flow", "scala", "typescript-legacy", "ts-legacy", "flow-legacy"].includes(flags.target)) {
```
martijnwalraven: Maybe we should get rid of the `ts` aliases? It seems confusing to have multiple names for the same target.

shadaj: I was mostly thinking about compatibility with `apollo-codegen` here. Is it okay to drop target names now?

martijnwalraven: As this is a new CLI, I think we should favor clarity over backwards compatibility. Plus, I think in `apollo-codegen` we didn't have `typescript-legacy` but `typescript-modern` so it's breaking anyway?

shadaj: Good point, I'll drop `ts`.

PR #460, opened 2018-06-28
fix for downcast in swift not available for Enum

dskatz22:


martijnwalraven:
Could you explain in more detail what this is doing? I also think we should have a regression test for this, to make sure it's a case we take into account when we make changes to codegen later.

PR #461, opened 2018-06-29
Change generated flow files extension to .js instead of .ts

TLadd:
Currently when the output to `apollo codegen:generate` is not specified and the target is flow, all of the generated files have a .ts extension, but it should be .js.


shadaj:
Good catch! Thanks for the PR!

ryanashcraft:
Just updated to apollo-codegen-flow@0.22.0, but the npm-published version doesn't seem to actually carry this change. @shadaj

PR #464, opened 2018-06-29
Command-level unit tests for codegen

shadaj:


packages/apollo-cli/src/__mocks__/localfs.ts
```diff
@@ -0,0 +1,12 @@
+import {Volume, createFsFromVolume} from 'memfs';
```
shadaj: This is the main part of how fs mocking works. Instead of importing `fs` everywhere, we import `localfs` which points to `fs` when regularly running or the in-memory volume in unit tests.

packages/apollo-cli/src/__mocks__/localfs.ts
```diff
@@ -0,0 +1,12 @@
...
+import {patchFs} from 'fs-monkey';
+
+export const vol = Volume.fromJSON({});
+export const fs = createFsFromVolume(vol);
+
+export function withGlobalFS<T>(thunk: () => T): T {
```
shadaj: This patch is needed to handle libraries like `glob`, which directly use `fs`. For those situations we wrap the library call in `withGlobalFS` to patch the `fs` library with our in-memory volume for the duration of the call.

packages/apollo-cli/src/commands/codegen/generate.ts
```diff
@@ -122,7 +126,9 @@ export default class Generate extends Command {
       {
         title: "Scanning for GraphQL queries",
         task: async (ctx, task) => {
-          const paths = await globby(flags.queries ? flags.queries.split('\n') : []);
+          const paths = withGlobalFS(() => {
+            return (flags.queries ? flags.queries.split('\n') : []).flatMap(p => fg.sync(p));
```
shadaj: `globby` had some weird incompatibilities with the in-memory volume, so I switched to `glob` here.

packages/apollo-cli/tsconfig.json
```diff
@@ -19,5 +19,5 @@
     "noUnusedLocals": false
   },
   "include": ["./src/**/*"],
-  "exclude": [ "**/__tests__/*" ]
+  "exclude": [ "**/__tests__/*", "**/__mocks__/*" ]
```
shadaj: Jest was complaining about having a JS and TS file for the mocks, so I just disabled compiling the mock files to JS.

packages/apollo-codegen-core/src/localfs.ts
```diff
@@ -0,0 +1,5 @@
+export const fs = require("fs");
```
shadaj: This is what's loaded during regular runtime, where `localfs` just points to regular `fs`.

packages/apollo-cli/src/commands/codegen/__tests__/generate.test.ts
```diff
@@ -0,0 +1,146 @@
...
+const simpleQuery = fs.readFileSync(path.resolve(__dirname, "./fixtures/simpleQuery.graphql"));
+
+beforeEach(() => {
+  vol.reset();
+  vol.fromJSON({
+    "__blankFileSoDirectoryExists": ""
+  });
+})
+
+jest.setTimeout(15000);
```
jbaxleyiii: was this timeout increase due to the new fs mocks?

packages/apollo-cli/src/commands/codegen/__tests__/generate.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+describe("successful codegen", () => {
+  test
+    .do(() => {
+      vol.fromJSON({
```
jbaxleyiii: okay this is killer

packages/apollo-cli/src/generate.ts
```diff
@@ -1,4 +1,4 @@
-import * as fs from 'fs';
+import { fs } from "apollo-codegen-core/lib/localfs";
```
jbaxleyiii: should we just export `fs` from codegen core? I could see us allowing a browser version as well one day

packages/apollo-cli/src/commands/codegen/__tests__/generate.test.ts
shadaj: The tests were failing randomly, so I did this to make Travis more consistent.

packages/apollo-cli/src/generate.ts
shadaj: This is mostly for internal use, so I didn't export it. Plus, I don't think the current setup would actually allow for a browser version through `localfs`..

Issue #468, opened 2018-06-30
Generate one file per .graphql file


TLadd:
Specifically interesting in the flow/typescript targets, but may make sense for the other targets as well. Currently when using apollo codegen:generate, I can either output types to a single file/directory by specifying an output path, or I can use the default behavior where it generates a file per-operation next to the queries.graphql file.

While I like having the generated types next to the queries file, I would prefer all of the types generated from a queries.graphql file going into a single file rather than having one file per-operation. The main reason is just standardization and to reduce file clutter. When looking at a directory, I know that my queries are defined in queries.graphql, and my types for those queries are defined in queryTypes.js/ts (or whatever the default filename would be for this use case)

So given a queries.graphql file like 
```graphql
query GetList(
  $first: Int!
  $after: String
) {
  list(
    first: $first
    after: $after
  ) {
    pageInfo {
      totalCount
    }
  }
}

mutation DeleteItem($input: DeleteItemInput!) {
  deleteItem(input: $input) {
    success
  }
} 
```

I would like a way for apollo codegen:generate to output a single queryTypes.js/ts in the same directory as queries.graphql with types for both the GetList query and the DeleteItem mutation.

mike-marcacci:
Hey @TLadd, before the name switch (from `apollo-codegen` to `apollo-cli`) flow-modern types (and possibly ts-modern?) were stored under a `__generated__` directory local to the source file. 

I have issue #469 and PR #470 open to add that functionality back but in a configureable way. This isn't exactly what you're looking for, but perhaps it helps mitigate your concerns?

TLadd:
Yeah, putting the types under a commonly-named directory would work for me

PR #470, opened 2018-07-02
Restore ability to put generated files in a directory relative to source

### LINKED ISSUE #469: Breaking changes in transition from flow-modern to flow
There appear to be several big changes in the way the previously-named "flow-modern" generator works:

1. The latest version of apollo on NPM generates `.ts` files (this is fixed in master by 9cc628c174646f082cb2dae44aa0ee842e445893 but should probably be pushed ASAP) 
2. The generated files are no longer put in a `__generated__` directory, which breaks all existing imports, tooling configs (eslint, prettier, etc), and **risks clobbering non-generated source files**.

Is the latter intentional? If so, we definitely need a CLI option to add it back; if not, then we should fix this ASAP.

If somebody at Apollo can point me on one of those directions, I'll have a PR up 5 minutes later! (I'm trying to move off my fork right now.)

mike-marcacci:
Fixes #469. Before the change from `apollo-codegen` to `apollo` cli, flow type definitions were output into a `__generated__` directory, next to their respective source files.

This adds a `localDirectory` flag that accepts a name to use locally for generated types. For example:

```sh
apollo generate:flow **/*.graphql --localDirectory=__generated__
```

This flag is ignored if `--output` is specified.

### DIFF
--- a/packages/apollo-cli/README.md
+++ b/packages/apollo-cli/README.md
@@ -55,7 +55,7 @@ ARGUMENTS
       Directory to which generated files will be written.
       - For TypeScript/Flow generators, this specifies a directory relative
       to each source file by default.
-      - For TypeScript/Flow generators with the "outputRelativeToCWD" flag
+      - For TypeScript/Flow generators with the "outputFlat" flag
       is set, and for the Swift generator, this specifies a file or
       directory (absolute or relative to the current working directory) to
       which:
@@ -94,10 +94,10 @@ OPTIONS
       operation ids (hashes) as properties on operation types [currently
       Swift-only]
 
-  --outputRelativeToCWD
+  --outputFlat
       By default, TypeScript/Flow will put each generated file in a
       directory next to its source file using the value of the "output" as
-      the directory name. Set "outputRelativeToCWD" to put all generated
+      the directory name. Set "outputFlat" to put all generated
       files in the directory relative to the current working directory
       defined by "output".
 
--- a/packages/apollo-cli/README.md
+++ b/packages/apollo-cli/README.md
@@ -5,15 +5,18 @@
 Apollo CLI brings together your GraphQL clients and servers with tools for validating your schema, linting your operations for compatibility with your server, and generating static types for improved client-side type safety.
 
 <!-- toc -->
-* [Apollo CLI](#apollo-cli)
-* [Usage](#usage)
-* [Commands](#commands)
-* [Code Generation](#code-generation)
-* [Contributing](#contributing)
-<!-- tocstop -->
+
+- [Apollo CLI](#apollo-cli)
+- [Usage](#usage)
+- [Commands](#commands)
+- [Code Generation](#code-generation)
+- [Contributing](#contributing)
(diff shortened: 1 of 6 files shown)


mike-marcacci:
Please note that the format changes are the result of a fixed `precommit` hook, which was incorrectly defined and failing.

mike-marcacci:
Please also note that something in your test on travis fails nondeterministically. I force-pushed the exact same code a couple times with amended commits (separating out the formatting changes from the meaningful ones) and sometimes it would fail (with messages about using a custom engine key), other times it would succeed. This (obviously) has nothing to do with my PR, but figured you should be aware.

martijnwalraven:
I definitely think this behavior is worth supporting, maybe even as a default. But if we want to keep it under an option, I'm wondering if we can find a more descriptive alternative to `--localDirectory`. Are there any examples of other CLIs that have flags for similar options?

mike-marcacci:
Hmm, personally my preference would be to reuse the `--output` flag, and add a boolean `--output-relative-to-source` but I was looking to make as few ripples as possible (just to get it working again, since maintaining forks of a monorepo is a big nuisance).

Would that be a workable solution, and default to `--output=__generated__ --output-relative-to-source` for Flow/TypeScript generators? I think this would be my ideal solution.

martijnwalraven:
@mike-marcacci That makes sense to me, but curious to hear what @shadaj and @jbaxleyiii think.

mike-marcacci:
Placing Flow and TypeScript files in directories relative to their source files is now the default behavior. By default, the directory name used is `__generated__`, which can be overridden by setting the `output` (first) cli arg.

To use the `output` to specify a single locaation the new `--outputRelativeToCWD` flag can be set.

To continue placing generated files immediately adjacent their source files (which is the current behavior, albiet only for a couple days and arguably undesireable) `output` can be set to an empty string. While available, I suspect this feature will never be used, due to the problems outlined in #468 and #469.

---

Side note: for consistency with the other flags, I used `outputRelativeToCWD` instead of `output-relative-to-source` since [dash separations are part of oclif boolean flags, such as the `no-` prefix](https://oclif.io/docs/flags.html) but apollo CLI had chosen to use a lowerCamelCase convention. I don't really care either way, but I'm curious why this switch was made, since `apollo-codegen` had used dash separations up until now.

mike-marcacci:
Also, to reiterate my comment about problems with the travis setup, the difference between these builds is a newline *in the commit message*:

- [Failure](https://travis-ci.org/apollographql/apollo-cli/builds/399241840?utm_source=github_status&utm_medium=notification)
- [Success](https://travis-ci.org/apollographql/apollo-cli/builds/399244635?utm_source=github_status&utm_medium=notification)

shadaj:
Hi @mike-marcacci this looks great! `outputRelativeToCWD` seems a bit hard to understand for me, would something like `outputFlat` work to express the idea of not placing types next to sources?

mike-marcacci:
@shadaj good idea. The PR's been updated with your suggestion.

PR #475, opened 2018-07-04
added clarity to schema:download

mwarger:
Hopefully less pain moving from apollo-codegen.


deltaskelta:
I think you also have to add it to the CLI code here

https://github.com/apollographql/apollo-cli/blob/7d99a31be25b42d24284b90b30c0ae7ade365d2a/packages/apollo-cli/src/commands/schema/download.ts#L31

jbaxleyiii:
@mwarger thanks so much for this helpful addition!

PR #497, opened 2018-07-17
Load config from apollo.config.js or package.json

shadaj:
Needs documentation, probably in a separate PR, but the implementation is ready.


packages/apollo-cli/src/commands/schema/download.ts
```diff
@@ -42,29 +48,56 @@ export default class SchemaDownload extends Command {
...
       {
-        title: `Saving schema to ${args.output}`,
+        title: "Loading Apollo config",
         task: async ctx => {
-          await promisify(fs.writeFile)(args.output, JSON.stringify(ctx.schema));
-        },
+          if (flags.config) {
+            ctx.config = loadConfigFromFile(flags.config) || {};
+          } else {
+            ctx.config = findAndLoadConfig(resolve(".")) || {};
```
jbaxleyiii: what is the goal of `resolve(".")` here?

packages/apollo-cli/src/commands/schema/download.ts
```diff
@@ -42,29 +48,56 @@ export default class SchemaDownload extends Command {
...
-        })
-      }),
       {
-        title: `Saving schema to ${args.output}`,
+        title: "Loading Apollo config",
         task: async ctx => {
-          await promisify(fs.writeFile)(args.output, JSON.stringify(ctx.schema));
-        },
+          if (flags.config) {
+            ctx.config = loadConfigFromFile(flags.config) || {};
```
jbaxleyiii: it feels like we should be able to combine this and the next function for a shared `loadConfig(path?)` option that defaults to process.cwd()

packages/apollo-cli/src/config.ts
```diff
@@ -0,0 +1,84 @@
...
+  return {
+    schema: obj.schema,
+    endpoint: loadEndpointConfig(obj.endpoint),
+    projectName: basename(dirname(configFilePath)),
+    operations:
+      typeof obj.operations === "string"
+        ? [obj.operations]
+        : obj.operations
+          ? (obj.operations as string[])
+          : ["**/*.graphql"],
```
jbaxleyiii: lets make sure to ignore node_modules by default

packages/apollo-cli/src/commands/codegen/generate.ts
```diff
@@ -175,124 +175,65 @@ export default class Generate extends Command {
...
-
-            if (ctx.schema) {
-              ctx.schema = extendSchema(ctx.schema, ast);
-            } else {
-              ctx.schema = buildASTSchema(ast);
-            }
-          }
+          task.title = `Scanning for GraphQL queries (${(ctx.documentSets as ResolvedDocumentSet[])
+            .map(s => s.documentPaths.length)
+            .reduce((a, b) => a + b, 0)} found)`; // this is a bogus wrong value
```
jbaxleyiii: is this bogus because its files not operations?

packages/apollo-cli/src/config.ts
```diff
@@ -0,0 +1,267 @@
...
+
+export interface DocumentSet {
+  schema?: string;
+  includes: string[];
+  excludes: string[];
+}
+
+export interface ApolloConfig {
+  projectFolder: string;
+  projectName?: string;
```
jbaxleyiii: can we change this to be just `name`?

What is the goal of `projectFolder`?

packages/apollo-cli/src/config.ts
```diff
@@ -0,0 +1,267 @@
...
+  url?: string; // main HTTP endpoint
+  subscriptions?: string; // WS endpoint for subscriptions
+  headers?: Object; // headers to send when performing operations
+}
+
+export interface SchemaDependency {
+  schema?: string;
+  endpoint?: EndpointConfig;
+  engineKey?: string;
+  extends?: string;
```
jbaxleyiii: how do we handle multiple extends? I'm assuming right now we don't

packages/apollo-cli/src/commands/codegen/generate.ts
shadaj: Oops! It used to be really bogus (counted the number of document sets) before. I'll drop the comment.

packages/apollo-cli/src/config.ts
shadaj: `projectFolder` tracks where the root of the project is for handling relative glob paths. It can't be picked up from `cwd` always (in VSCode for example), so we treat it as a config value.

packages/apollo-cli/src/config.ts
shadaj: So for me multiple extends didn't make too much sense because you're taking a schema and adjusting it. I'm not sure what it would mean to extend multiple schemas. In the future though, I would like to support multiple documents extending a schema in the case that you split up your types in multiple files.

shadaj:
The order for loading schemas is based on which data is closest:
1. The schema value (if it's a file read it, if it's a URL introspect it)
2. The endpoint value (introspect it)
3. The engine key (grab the latest uploaded schema)

Issue #498, opened 2018-07-18
apollo codegen:generate flow fails if a newer version of @babel/types is installed


TLadd:
Example Repo: https://github.com/TLadd/apollo-babel-types-fail/tree/master

Perhaps the same bug as https://github.com/apollographql/apollo-cli/issues/443

Running `apollo codegen:generate --schema=schema.json --target=flow`, results in the error

```
 ✔ Scanning for GraphQL queries (1 found)

 ✔ Loading GraphQL schema
 ✔ Parsing GraphQL schema
 ↓ Loading client-side GraphQL schema [skipped]
   → Path to client schema not provided
 ✖ Generating query files with 'flow' target
   → Property value expected type of string but got null
TypeError: Property value expected type of string but got null
    at Object.validate (~/github/apollo-babel-types-fail/node_modules/@babel/types/lib/definitions/utils.js:177:13)
    at validate (~/github/apollo-babel-types-fail/node_modules/@babel/types/lib/validators/validate.js:17:9)
    at builder (~/github/apollo-babel-types-fail/node_modules/@babel/types/lib/builders/builder.js:50:27)
    at Object.StringLiteralTypeAnnotation (~/github/apollo-babel-types-fail/node_modules/@babel/types/lib/builders/generated/index.js:1109:27)
    at graphql_1.sortEnumValues.map (~/github/apollo-babel-types-fail/node_modules/apollo-codegen-flow/lib/language.js:20:24)
    at Array.map (<anonymous>)
    at FlowAPIGenerator.enumerationDeclaration (~/github/apollo-babel-types-fail/node_modules/apollo-codegen-flow/lib/language.js:18:8)
    at FlowAPIGenerator.typeAliasForEnumType (~/github/apollo-babel-types-fail/node_modules/apollo-codegen-flow/lib/codeGeneration.js:89:35)
    at context.typesUsed.filter.forEach (~/github/apollo-babel-types-fail/node_modules/apollo-codegen-flow/lib/codeGeneration.js:27:19)
    at Array.forEach (<anonymous>)
error Command failed with exit code 1.
```

When I run yarn ls @babel/types, the output is
```
├─ @babel/generator@7.0.0-beta.38
│  └─ @babel/types@7.0.0-beta.38
├─ @babel/types@7.0.0-beta.46
└─ apollo-codegen-core@0.24.0
   └─ @babel/types@7.0.0-beta.38
```

[...]

PR #499, opened 2018-07-18
Explicitly depend on @babel/types in flow/typescript codegen (#498)

TLadd:
* Ensures @babel/types doesn't resolve to another version included in a projects deps

Fixes https://github.com/apollographql/apollo-cli/issues/498


shadaj:
Thanks for the fix!

PR #503, opened 2018-07-20
fix broken source code links

chentsulin:


shadaj:
Hi @chentsulin! Unfortunately, the `README.md` file is auto-generated by the CLI library we use, `oclif`, so these fixes would be overriden by the next release.

PR #505, opened 2018-07-20
Add docs for Apollo config

shadaj:

### DIFF
--- a/packages/apollo-cli/README.md
+++ b/packages/apollo-cli/README.md
@@ -248,23 +248,20 @@ Or in `apollo.config.js` style:
 ...
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


klujanrosas:
Hi there, It's awesome that we get documentation for how to do config.

I was wondering, Does the `apollo.config.js`(or package.json) file support the other options that are used in the CLI?

I'm talking about `--outputFlat, --watch, --addTypename, etc`

shadaj:
Hi @klujanrosas! For the first release of config files, they won't support codegen options, only the configuration needed to associate queries with schemas, but we definitely plan on making those options configurable in the future.

PR #511, opened 2018-07-24
Stop default flag values from overriding custom config

klujanrosas:
* Check if a custom config for queries hasn't been setup, else fallback to flags.queries

### DIFF
--- a/packages/apollo-cli/src/load-config.ts
+++ b/packages/apollo-cli/src/load-config.ts
@@ -59,7 +59,7 @@ export function loadConfigStep(
         };
       }
 
-      if (flags.queries) {
+      if (!ctx.config.queries || ctx.config.queries.length == 0 && flags.queries) {
         ctx.config.queries = [
           {
             schema: "default",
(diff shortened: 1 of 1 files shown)


klujanrosas:
I checked Travis and it tells me that this key `$STAGING_API_KEY` is missing

shadaj:
Hi @klujanrosas! Allowing the `queries` flag to override the config was actually an intended behavior, as users may sometimes want to specifically generate types for a subset of queries.

However, we didn't catch that the queries would *always* be overriden by the default value. Maybe this could be fixed by only allowing overrides when the flag is not the default value?

klujanrosas:
Hi @shadaj , I went ahead and implemented your proposed solution.
Let me know if there are improvements or changes to be made.

Thanks!

shadaj:
@klujanrosas could you update with the latest master? We have another fix for the failing Travis issue that should make your changes pass.

shadaj:
Thanks so much for the fix @klujanrosas!

PR #518, opened 2018-07-30
[typescript] Only output comment for input if there is one

danilobuerger:
Only output comment for input if there is one, this is aligned with how enums are generated: https://github.com/apollographql/apollo-cli/blob/master/packages/apollo-codegen-typescript/src/language.ts#L57-L62


shadaj:
Thanks for the fix @danilobuerger!

PR #520, opened 2018-07-31
[TS] Dedup enums and inputs by using global types file

danilobuerger:
This change will deduplicate enum and input definitions by placing them in a global types file (`${outputPath}/globalTypes.ts`). Those global types will no longer be printed into the "local" type files, instead they will be imported.

### DIFF
--- a/packages/apollo-cli/src/generate.ts
+++ b/packages/apollo-cli/src/generate.ts
@@ -12,7 +12,7 @@ import { generateSource as generateSwiftSource } from "apollo-codegen-swift";
 import { generateSource as generateTypescriptLegacySource } from "apollo-codegen-typescript-legacy";
 import { generateSource as generateFlowLegacySource } from "apollo-codegen-flow-legacy";
 import { generateSource as generateFlowSource } from "apollo-codegen-flow";
-import { generateSource as generateTypescriptSource } from "apollo-codegen-typescript";
+import { generateLocalSource as generateTypescriptLocalSource, generateGlobalSource as generateTypescriptGlobalSource } from "apollo-codegen-typescript";
 import { generateSource as generateScalaSource } from "apollo-codegen-scala";
 import { GraphQLSchema } from "graphql";
 import { FlowCompilerOptions } from '../../apollo-codegen-flow/lib/language';
@@ -71,12 +71,9 @@ export default function generate(
       writeOperationIdsMap(context);
       writtenFiles += 1;
     }
-  } else if (target === "flow" || target === "typescript" || target === "ts") {
+  } else if (target === "flow") {
     const context = compileToIR(schema, document, options);
-    const { generatedFiles, common } =
-      target === "flow"
-        ? generateFlowSource(context)
-        : generateTypescriptSource(context);
+    const { generatedFiles, common } = generateFlowSource(context);
 
     const outFiles: {
       [fileName: string]: BasicGeneratedFile;
@@ -116,6 +113,52 @@ export default function generate(
       fs.writeFileSync(
         outputPath,
         generatedFiles.map(o => o.content.fileContents).join("\n") + common
+      );
+
+      writtenFiles += 1;
+    }
+  } else if (target === "typescript" || target === "ts") {
+    const context = compileToIR(schema, document, options);
+    const generatedFiles = generateTypescriptLocalSource(context);
+    const generatedGlobalFile = generateTypescriptGlobalSource(context);
+
+    const outFiles: {
+      [fileName: string]: BasicGeneratedFile;
+    } = {};
+
+    if (nextToSources || (fs.existsSync(outputPath) && fs.statSync(outputPath).isDirectory())) {
+      if (nextToSources && !fs.existsSync(outputPath)) {
+        fs.mkdirSync(outputPath);
+      }
+
+      const globalSourcePath = path.join(outputPath, "globalTypes.ts");
+      outFiles[globalSourcePath] = {
+        output: generatedGlobalFile.fileContents,
+      };
+
+      generatedFiles.forEach(({ sourcePath, fileName, content }) => {
+        let dir = outputPath;
+        if (nextToSources) {
+          dir = path.join(path.dirname(sourcePath), dir);
+          if (!fs.existsSync(dir)) {
+            fs.mkdirSync(dir);
+          }
+        }
+
+        const outFilePath = path.join(dir, fileName);
+        outFiles[outFilePath] = {
+          output: content({ outputPath: outFilePath, globalSourcePath }).fileContents,
+        };
+      });
+
+      writeGeneratedFiles(outFiles, path.resolve("."));
+
+      writtenFiles += Object.keys(outFiles).length;
+    } else {
+      fs.writeFileSync(
+        outputPath,
+        generatedFiles.map(o => o.content().fileContents).join("\n") +
+          generatedGlobalFile.fileContents,
       );
 
       writtenFiles += 1;
--- a/packages/apollo-codegen-typescript/src/codeGeneration.ts
+++ b/packages/apollo-codegen-typescript/src/codeGeneration.ts
@@ -74,10 +74,7 @@ function printGlobalImport(
   outputPath: string,
   globalSourcePath: string,
 ) {
-  const types = typesUsed.filter((type) => {
-    return type instanceof GraphQLEnumType || type instanceof GraphQLInputObjectType;
-  });
-  if (types.length > 0) {
+  if (typesUsed.length > 0) {
     const relative = path.relative(
       path.dirname(outputPath),
       path.join(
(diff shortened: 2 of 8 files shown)


packages/apollo-codegen-typescript/src/codeGeneration.ts
```diff
@@ -246,6 +335,68 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
...
+      nested: SelectionSet,
+      acc: GraphQLType[]
+    ) => {
+      acc = nested.selections
+        .reduce((selectionAcc, selection) => {
+          switch (selection.kind) {
+            case 'Field':
+            case 'TypeCondition':
+              if (selection.selectionSet) {
+                selectionAcc = reduceTypesForDocument(selection.selectionSet, selectionAcc);
```
shadaj: Updating `selectionAcc` and returning later introduces a bit of unnecessary mutability. Could you just directly return the additional types here and return `selectionAcc` for the default case?

packages/apollo-codegen-typescript/src/codeGeneration.ts
```diff
@@ -246,6 +335,68 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
...
+  ) =>{
+    if (type instanceof GraphQLNonNull) {
+      type = getNullableType(type);
+    }
+
+    if (type instanceof GraphQLList) {
+      type = type.ofType
+    }
+
+    acc = maybePush(acc, type);
```
shadaj: Same as above, could this be replaced by an immutable update `[...acc, type]`?

packages/apollo-codegen-typescript/src/codeGeneration.ts
```diff
@@ -246,6 +335,68 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
...
+
+  private reduceGlobalTypesUsed = (
+    acc: (GraphQLType | GraphQLOutputType)[],
+    type: GraphQLType
+  ) =>{
+    if (type instanceof GraphQLNonNull) {
+      type = getNullableType(type);
+    }
+
+    if (type instanceof GraphQLList) {
```
shadaj: What happens with nested lists here? This might need a recursive call so the body of this branch would be something like `return reduceGlobalTypesUsed(acc, type.ofType)`.

packages/apollo-codegen-typescript/src/codeGeneration.ts
```diff
@@ -246,6 +335,68 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
...
+      return (type instanceof GraphQLEnumType || type instanceof GraphQLInputObjectType)
+        && docTypesUsed.find((typeUsed) => type === typeUsed);
+    });
+  }
+
+  private reduceGlobalTypesUsed = (
+    acc: (GraphQLType | GraphQLOutputType)[],
+    type: GraphQLType
+  ) =>{
+    if (type instanceof GraphQLNonNull) {
```
shadaj: Same as below, probably best to just recurse in case you have a non-nullable list for example.

packages/apollo-codegen-typescript/src/codeGeneration.ts
```diff
@@ -246,6 +335,68 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
     this.scopeStackPop();
   }
 
+  public getGlobalTypesUsedForOperation(doc: Operation | Fragment, context: CompilerContext) {
+    let docTypesUsed: GraphQLType[] = [];
+
+    if (doc.hasOwnProperty('operationName')) {
```
shadaj: Could this be replaced by an `instanceof` or `kind` check? Then it wouldn't be necessary to explicitly cast the value later.

packages/apollo-codegen-typescript/src/codeGeneration.ts
```diff
@@ -246,6 +335,68 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
...
+      const operation = doc as Operation;
+      docTypesUsed = operation.variables.map(({ type }) => type);
+    }
+
+    const reduceTypesForDocument = (
+      nested: SelectionSet,
+      acc: GraphQLType[]
+    ) => {
+      acc = nested.selections
```
shadaj: Instead of setting and immediately returning `acc`, can you just directly return it here?

packages/apollo-cli/src/generate.ts
```diff
@@ -118,6 +115,52 @@ export default function generate(
...
+    const generatedFiles = generateTypescriptLocalSource(context);
+    const generatedGlobalFile = generateTypescriptGlobalSource(context);
+
+    const outFiles: {
+      [fileName: string]: BasicGeneratedFile;
+    } = {};
+
+    if (nextToSources || (fs.existsSync(outputPath) && fs.statSync(outputPath).isDirectory())) {
+      if (nextToSources && !fs.existsSync(outputPath)) {
+        fs.mkdirSync(outputPath);
```
shadaj: In this case you want to be checking and creating the joined path, right? I think you might need to move the `dir` adjustment logic from below up to the top of this branch so that you are using the same directory everywhere. Plus, it's expensive to check directories for every file so it would be faster to just do that once here.

packages/apollo-cli/src/generate.ts
```diff
@@ -118,6 +115,52 @@ export default function generate(
...
+  } else if (target === "typescript" || target === "ts") {
+    const context = compileToIR(schema, document, options);
+    const generatedFiles = generateTypescriptLocalSource(context);
+    const generatedGlobalFile = generateTypescriptGlobalSource(context);
+
+    const outFiles: {
+      [fileName: string]: BasicGeneratedFile;
+    } = {};
+
+    if (nextToSources || (fs.existsSync(outputPath) && fs.statSync(outputPath).isDirectory())) {
```
shadaj: This logic still seems pretty similar to Flow other than the change in the call to `content`. Would it be possible to extract out the common logic?

shadaj: In fact, you could probably use the TypeScript code for Flow as is. The Flow generator would receive the additional property `globalSourcePath` but it would just ignore it.

packages/apollo-cli/src/generate.ts
danilobuerger: I am not sure I understand what you are trying to say. This line would create the directory for `globalTypes.ts` if we are placing next to sources. In that case we are not guaranteed to have the `/__generated__` (or whatever it is set to) directory (as there is no source next to it), so its checked here. The later `dir` logic is for the directories we need for placing next to the source. That part is just copied from the `flow` target. If instead we are using output flat, no checks or creates would happen except the initial one (on Line 129).

packages/apollo-cli/src/generate.ts
danilobuerger: So the separation was with intent. Yes there is a lot of logic, but there is also https://github.com/apollographql/apollo-cli/pull/410 which I didn't want to interfere with. I wanted to keep them separate until Flow catches up and dedups as well. Then those paths could be merged again (if similar enough). What do you think?

If we use the TS code as is for flow, we would also have to modify the flow package to return content as a closure. (Which I am not really interested in doing to be honest as I don't care much about Flow.)

packages/apollo-cli/src/generate.ts
shadaj: Sorry, I was confusing `sourcePath` and `outputPath`. You're right, this makes sense for the `globalTypes.ts` file.

packages/apollo-codegen-typescript/src/codeGeneration.ts
danilobuerger: Yeah, this currently fails on `[[EnumCommentTestCase!]!]!`. Added a test case.

packages/apollo-codegen-typescript/src/codeGeneration.ts
danilobuerger: Yeah, this currently fails on `[[EnumCommentTestCase!]!]!`. Added a test case.

packages/apollo-codegen-typescript/src/codeGeneration.ts
danilobuerger: As far as I can tell, the intention of `maybePush` from the original `reduceTypesUsed` was that there are no duplicated types in `acc` as `maybePush` checks for dups. However since we filter later anyway, we could accumulate duplicates here. Or we could collect them as an object instead thus having O(1) later for the lookups.

danilobuerger:
Thanks for the thorough review @shadaj !

Initially I just copied the code for `getGlobalTypesUsedForOperation` from `getTypesUsedForOperation` and tried to make it work. But your review inspired me to do better, so I rewrote it from scratch. Now it looks far better (and fixed the nested non-null list bug).

PR #521, opened 2018-07-31
Add README for the VS Code extension

shadaj:


packages/apollo-vscode/README.md
```diff
@@ -0,0 +1,66 @@
...
+- Lets you run queries, mutations, and subscriptions within the IDE with code-completion for variables
+- Displays performance statistics from [Apollo Engine](https://www.apollographql.com/engine) inline with your queries
+- Jump-to-definition for fragments and schema types
+- Detects fragment references and shows them next to definitions
+
+## How to get it?
+
+Open up VS Code and search for the extension "Apollo".
+
+## How to get it set up?
```
peggyrayzis: What about uploading your schema to Engine? What if you're not running your GraphQL server locally?

packages/apollo-vscode/README.md
```diff
@@ -0,0 +1,66 @@
...
+- Adds syntax highlighting for GraphQL files and `gql` templates inside JavaScript files
+- Code-completes fields, arguments, types, and variables in your queries
+- Lets you run queries, mutations, and subscriptions within the IDE with code-completion for variables
+- Displays performance statistics from [Apollo Engine](https://www.apollographql.com/engine) inline with your queries
+- Jump-to-definition for fragments and schema types
+- Detects fragment references and shows them next to definitions
+
+## How to get it?
+
+Open up VS Code and search for the extension "Apollo".
```
peggyrayzis: Will this extension conflict with any other GraphQL extensions you have installed?

packages/apollo-vscode/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Get instant feedback and intelligent autocomplete as you write queries
+- Run queries against your GraphQL server without leaving your editor
+- View performance statistics next to your query definitions
+
+![Code completing queries](images/variable-argument-completion.gif)
+![Running a query with variables](images/query-with-vars.gif)
+![Viewing Engine statistics](images/engine-stats.png)
+
+## Features
```
peggyrayzis: Does this extension work with non-Apollo GraphQL servers?

packages/apollo-vscode/README.md
```diff
@@ -0,0 +1,66 @@
...
+    ]
+  }
+}
+```
+
+## How do we run queries?
+
+The extension enables query execution when an `endpoint` is specified for the schema you are targeting. When you hit `Run query/mutations/subscription`, the extension will open a menu for you to input variables if there are any defined or directly run the query otherwise.
+
+Queries and mutations are executed with the same [Apollo Link HTTP](https://www.apollographql.com/docs/link/links/http.html) configured to run queries against the specified endpoint. Subscriptions are handled with [Apollo Link WS](https://www.apollographql.com/docs/link/links/ws.html) configured to run against the `subscriptions` key in your schema or your endpoint with `http` replaced by `ws` if the key is not defined.
```
peggyrayzis: Instead of mentioning Apollo Link HTTP & WS, we should probably just mention that it uses the same network interface you use with Apollo Client.

packages/apollo-vscode/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Troubleshooting
+
+### Extension not starting correctly? Check the logs!
+
+If you're having trouble with the extension not launching or not detecting your configuration files, you can check the language server logs.
+
+1.  Open the VS Code output tab from the bottom of your editor
```
peggyrayzis: What's the VSCode output tab? This should be more clear.

packages/apollo-vscode/README.md
shadaj: Yes! Is there somewhere we can make this more clear?

packages/apollo-vscode/README.md
shadaj: Probably, since both register a GraphQL syntax. Should I add a line to warn users about installing it alongside other extensions?

shadaj:
Merging what we have so far since we depend on the images in this PR for the marketplace description.

PR #524, opened 2018-08-02
Fixed broken command links

danilobuerger:


danilobuerger:
@shadaj Updated. However, this will now always point to `master` instead of a specific version as `oclif`, as far as I can tell, only sees the package version and since `apollo-cli` is on a different one then the rest of the repos, it can't find out the proper tag.

shadaj:
This is so awesome! Thanks for the fix!

PR #533, opened 2018-08-06
Fall back to other methods of schema loading when one fails

shadaj:
Helps in situations where one method fails like when an endpoint exists but introspection is disabled.

### DIFF
--- a/packages/apollo-cli/src/load-schema.ts
+++ b/packages/apollo-cli/src/load-schema.ts
@@ -7,15 +7,28 @@ export async function loadSchema(
   config: ApolloConfig
 ): Promise<GraphQLSchema | undefined> {
   if (dependency.schema) {
-    return await fetchSchema({ url: dependency.schema }, config.projectFolder);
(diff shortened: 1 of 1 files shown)


danilobuerger:
@shadaj this now leads to unclear error messages when fetching from a server without introspection. 
Error message is now:

```
$ apollo schema:download --endpoint ... schema.json
 ✔ Loading Apollo config
 ✔ Fetching current schema
 ✖ Saving schema to schema.json
   → Expected undefined to be a GraphQL schema.
Error: Expected undefined to be a GraphQL schema.
```

previously it was:

```
$ apollo schema:download --endpoint ... schema.json
 ✔ Loading Apollo config
 ✖ Fetching current schema
   → Response not successful: Received status code 400
   Saving schema to schema.json
Error: Response not successful: Received status code 400
```

abernix:
@shadaj I've proposed reverting this in #613.  I'd appreciate your thoughts on that, but it seems like this approach won't work.

PR #546, opened 2018-08-12
[TS] Added outputGlobalTypes option to specify path for global types

danilobuerger:


williamboman:
This is great! I think it would also be nice to be able to set the filename (`globalTypes.ts`).

danilobuerger:
@williamboman Thats exactly what this option is for, the path includes the filename

williamboman:
Oh shit, my bad!

zionts:
@danilobuerger thanks for submitting this! Could you update https://github.com/apollographql/apollo-cli/blob/master/CHANGELOG.md with what this PR is doing?

packages/apollo-cli/src/commands/codegen/generate.ts
```diff
@@ -95,6 +95,10 @@ export default class Generate extends Command {
       description:
         'By default, TypeScript/Flow will put each generated file in a directory next to its source file using the value of the "output" as the directory name. Set "outputFlat" to put all generated files in the directory relative to the current working directory defined by "output".'
     }),
+    outputGlobalTypes: flags.string({
```
zionts: Not doing a full review, but I think that `outputGlobalTypes` is a strange name (and reads like a boolean) for a string that is meant to specify a file location to write global type information to. Perhaps something like `globalTypesFilePath`?

danilobuerger:
@zionts In this repo the Changelog is done on Publish. See https://github.com/apollographql/apollo-cli/commits/master/CHANGELOG.md not during PRs.

packages/apollo-cli/src/commands/codegen/generate.ts
danilobuerger: The intention was that it is aligned with the other output options (`output` and `outputFlat`) for easier grouping if they all begin with `output...`. But I really don't care what the param is named :-)

martijnwalraven: I agree `outputGlobalTypes` feels a bit weird as a flag name. Maybe `globalTypesFile`?

danilobuerger: @martijnwalraven I amended the commit

rahuljaindream11:
Will this key work for swift target

PR #553, opened 2018-08-14
Initial query extraction using AST

### LINKED ISSUE #556: Having issues with Graphql Fragments
```
export const MESSAGE_FRAGMENT = gql`
	fragment MessageFragment on Message {
		id
		parentId
		body
		author {
			username
			avatar_url {
				url
			}
		}
	}
`


export const SHOW_CHANNEL_QUERY = gql`
	query ShowChannelQuery($channelId: ID!) {
		showChannel(channelId: $channelId) {
			id
			name
			slug
			messages {
				...MessageFragment
			}
		}
	}
	${MESSAGE_FRAGMENT}
`

```

Works fine with querying, but when I try to generate types I get a error saying unknown fragment.

evans:
This is a query extraction technique. We currently output map from sha to query string in a file, manifest.json. The output location could be customizable and probably should be stdout by default, so that a user can pipe it into a file should they choose.

fixes #556


packages/apollo-cli/src/commands/queries/extract.ts
```diff
@@ -0,0 +1,80 @@
+import 'apollo-codegen-core/lib/polyfills';
+import { Command, flags } from '@oclif/command';
+import * as Listr from 'listr';
+import { print } from 'graphql';
+import { sha512 } from 'js-sha512';
```
abernix: Was there a particular reason for using [`js-sha512`](https://nvm.im/js-sha512), a JavaScript implementation of the SHA-512 routine, rather than Node.js’ native [`crypto.createHash`](
https://nodejs.org/api/crypto.html#crypto_crypto_createhash_algorithm_options) (which is implemented in C/C++).

packages/apollo-cli/src/commands/queries/extract.ts
```diff
@@ -0,0 +1,80 @@
+import 'apollo-codegen-core/lib/polyfills';
```
abernix: What’s this polyfill relatated to? Might be worth putting a comment about its necessity/purpose since it’s imported in its entirety rather generically.

packages/apollo-cli/src/commands/queries/extract.ts
```diff
@@ -0,0 +1,80 @@
...
+import { Command, flags } from '@oclif/command';
+import * as Listr from 'listr';
+import { print } from 'graphql';
+import { sha512 } from 'js-sha512';
+import * as fs from 'fs';
+
+import { loadQueryDocuments } from 'apollo-codegen-core/lib/loading';
```
abernix: Longer term, is it confusing to leave the `loadQueryDocuments` functionality in `apollo-codegen-core`? (It doesn’t really sound like a utility library and we’re not doing code-gen.)

packages/apollo-cli/src/commands/queries/extract.ts
evans: I'm not sure why it's included myself. I'll dig around and see. My first guess is lack of native support for the functions in Node 8

packages/apollo-cli/src/commands/queries/extract.ts
evans: hmm we could probably make the judgement of apollo-codegen-core in general. Let's keep it in mind as we continue digging into the functionality

packages/apollo-cli/src/commands/queries/extract.ts
```diff
@@ -0,0 +1,80 @@
...
+          ctx.mapping = {};
+          (ctx.operations as Array<{document: string}>).forEach(({ document }) => {
+            ctx.mapping[sha512.update(document).hex()] = document;
+          });
+        },
+      },
+      {
+        title: 'Outputing manifest',
+        task: async ctx => {
+          fs.writeFileSync('manifest.json', JSON.stringify(ctx.mapping));
```
danilobuerger: This should really be an option before this gets merged. People will riot otherwise 😂

evans: 😆 Totally! What do you think about the default being output to stdout and then having an option to specify the file with `--output`? Or should it automatically send it to the manifest?

danilobuerger: That would be one option, but consider the rest of the cli that just takes the output path as the last arg without a flag.

evans:
This can be tested with a command similar to 

```bash
(cd ~/engine/frontend  && ~/apollo-cli/packages/apollo-cli/bin/run queries:extract --queries=**/*.jsx --queries=./imports/**/*.js && cat manifest.json | jq)
```

packages/apollo-cli/src/commands/queries/check.ts
```diff
@@ -1,46 +1,45 @@
...
 
-import { engineFlags } from "../../engine-cli";
-import { engineLink, getIdFromKey } from "../../engine";
-import { gitInfo } from "../../git";
-import { VALIDATE_OPERATIONS } from "../../operations/validateOperations";
-import { ChangeType } from "../../printer/ast";
-import { format } from "../schema/check";
-import { resolveDocumentSets } from "../../config";
-import { loadConfigStep } from "../../load-config";
+import { engineFlags } from '../../engine-cli';
```
abernix: Would it be possible to revert the Prettier changes present in 721429634a8a41768a86363f3c781fc8485eefca?  We should run it on the whole repository outside the scope of this PR if that's what we want, but for the sake of accurately reviewing this PR it should be kept as a separate concern.

packages/apollo-codegen-core/src/loading.ts
```diff
@@ -197,14 +92,19 @@ export function extractDocumentFromJavascript(
...
-  let matches = [];
+  let matches: string[] = [];
 
-  while ((match = re.exec(content))) {
-    const doc = match[1].replace(/\${[^}]*}/g, '');
+  try {
+    matches = extractDocumentsWithAST(content, options);
+  } catch (e) {
+    console.log('Extraction using AST failed with \n', e, '\nRetrying using regex');
+    const re = new RegExp(tagName + 's*`([^`]*)`', 'g');
```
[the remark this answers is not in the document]
evans: Removing the regex fallback makes sense. It will cause some configuration pain in some cases., i.e. actively excluding `.d.ts` files, since there are many that don't parse well. For the most part, I trust the parsers. Running it on spectrum and the engine frontend yielded good results.

evans:
The behavior of the `query:extract` now mirrors `schema:download` in terms of output location.

The PR also improves fragment inclusion for query checking, fixing #556. The changes also apply to mobile, since the fragment inclusions work for .gql and .graphql files.

evans:
Comments from @benjamn to optimize the solution:

1. use a traversal that does not construct `NodePath` objects, since we don’t need to know about the ancestry of the node
2. use a regular expression to find possible `gql` identifiers, and then ignore subtrees that don’t contain any
3. (harder) modify the parser so that just calls a callback when it hits a `gql`-tagged template string, and doesn’t generate AST nodes we don’t care about

Hopefully these are useful in the future 🤞

PR #564, opened 2018-08-29
[TS] Add support for calling an external module with the query source

DrOfAwesomeness:
This PR is heavily modeled after the excellent work done by @standy in PR #423. I've rewritten it to support the new lerna repository layout and modern TypeScript code generation. This is an attempt at solving the issues brought up in #308. I set up a quick example showing the code generated by this PR at https://github.com/DrOfAwesomeness/apollo-cli-pr-example.

/label typescript


DrOfAwesomeness:
@rricard and/or @shadaj (sorry, I'm not sure who exactly is best to ping for this), could you review this PR? My company needs this feature internally, and I'd like to avoid having to maintain a custom fork if possible.

Thanks!

DrOfAwesomeness:
Hey, has anyone had a chance to review this yet?

shadaj:
Hi @DrOfAwesomeness this looks really awesome! Could you put together a couple unit tests (likely snapshot based, you can see the CLI tests for an example) to make sure that this feature won't be broken by mistake in the future?

DrOfAwesomeness:
Thanks for taking a look at this @shadaj! I added three snapshot-based unit tests. Will those be sufficient?

packages/apollo-codegen-typescript/src/codeGeneration.ts
```diff
@@ -223,13 +231,24 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
     this.scopeStack = [];
   }
 
-  fileHeader() {
+  fileHeader(tagName?: string) {
```
shadaj: This is a bit confusing for me, since specifying `tagName` by itself does not actually guarantee that the `import` will be generated. Could this function be adjusted to always generate the `import ${tagName}` if `tagName` is defined and then also generate the `import codeGenerationModule` if that config value is specified? This would have to come with a change above to call `generator.fileHeader(codeGenerationModule && "gql")`.

Also, I think the logic to generate imports merits a function of its own, since it's less of a header and fits in more with the rest of the types being generated.

DrOfAwesomeness: Ah, I can see how that'd be confusing. I've split the import generation logic off into `generator.fileImports(tagName: string, codeGenerationModule: string)` and changed it so it'd always generate the `import ${tagName}` if `tagName` is defined.

DrOfAwesomeness:
Hey, I just wanted to check on this - is there anything else needed from me here, or are we just waiting on someone else to merge it? Not trying to push, just want to make sure this isn't waiting on me 😃

DrOfAwesomeness:
Hey @shadaj, just wanted to check in again - is there anything else I need to do for this to get merged?
Thanks!

shadaj:
Hey @DrOfAwesomeness I'm not officially running Apollo CLI anymore (my internship ended in August), so just waiting on someone else to also take a look at this! Nothing needed from you right now though 😃

jbaxleyiii:
@DrOfAwesomeness I think this is a really interesting discussion! I think we should go through an RFC on the apollo-feature-requests repo first. Would you open up an issue there for discussion?

martinek:
@jbaxleyiii I can't find this feature in apollo-feature-requests repo. Was this discussed ever again somewhere or is this feature available in codegen already?

PR #565, opened 2018-08-29
Support for self-signed certificates

### LINKED ISSUE #199: Support for self-signed certs?
When developing locally, it is important to support self-signed certs. I believe adding this is quite simple and can be disabled by default for security reason.

robertomg:
### Adds support for skipping certificate validation

Often developers use a self-signed certificate for local services. In development process is not uncommon to see those kind of https requests without a Root CA signed certificate. This PR tries to solve the issue reported on #199 when using `apollo-cli` for downloading the GraphQL schema.

`schema:download` has a new option `--insecure` or equally `-k`  for skipping certificate validation. These names were chosen following the popular unix command `curl` convention.

It is my first PR on the project so any kind of feedback is more than welcome.

Fixes #199

### DIFF
--- a/packages/apollo-cli/src/commands/schema/download.ts
+++ b/packages/apollo-cli/src/commands/schema/download.ts
@@ -35,7 +35,7 @@ export default class SchemaDownload extends Command {
       description:
         "The URL of the server to fetch the schema from or path to ./your/local/schema.graphql"
     }),
-    insecure: flags.boolean({
+    skipSSLValidation: flags.boolean({
       char: "k",
       description: "Allow connections to a SSL site without certs"
     }),
--- a/packages/apollo-cli/src/commands/schema/download.ts
+++ b/packages/apollo-cli/src/commands/schema/download.ts
@@ -35,6 +35,10 @@ export default class SchemaDownload extends Command {
       description:
         "The URL of the server to fetch the schema from or path to ./your/local/schema.graphql"
     }),
+    insecure: flags.boolean({
+      char: "k",
+      description: "Allow connections to a SSL site without certs"
+    }),
 
     ...engineFlags
   };
--- a/packages/apollo-cli/src/config.ts
+++ b/packages/apollo-cli/src/config.ts
@@ -11,6 +11,7 @@ export interface EndpointConfig {
   url?: string; // main HTTP endpoint
   subscriptions?: string; // WS endpoint for subscriptions
   headers?: Object; // headers to send when performing operations
+  skipSSLValidation?: boolean; // bypass the SSL validation on a HTTPS request
 }
 
 export interface SchemaDependency {
@@ -19,7 +20,6 @@ export interface SchemaDependency {
   engineKey?: string;
   extends?: string;
   clientSide?: boolean;
-  skipsSSLValidation?: boolean;
 }
 
 export interface DocumentSet {
--- a/packages/apollo-cli/src/config.ts
+++ b/packages/apollo-cli/src/config.ts
@@ -19,6 +19,7 @@ export interface SchemaDependency {
   engineKey?: string;
   extends?: string;
   clientSide?: boolean;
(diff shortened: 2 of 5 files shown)


packages/apollo-cli/src/config.ts
```diff
@@ -19,6 +19,7 @@ export interface SchemaDependency {
   engineKey?: string;
   extends?: string;
   clientSide?: boolean;
+  skipsSSLValidation?: boolean;
```
martijnwalraven: I think this should probably be part of [`EndpointConfig`](
https://github.com/apollographql/apollo-cli#endpoint-configuration), similar to endpoint-specific headers.

packages/apollo-cli/src/commands/schema/download.ts
```diff
@@ -35,6 +35,10 @@ export default class SchemaDownload extends Command {
       description:
         "The URL of the server to fetch the schema from or path to ./your/local/schema.graphql"
     }),
+    insecure: flags.boolean({
```
martijnwalraven: Wondering about the naming, because it's a bit confusing to have `insecure`, `skipsSSLValidation `, and `rejectUnauthorized` referring to the same option. 

`skipSSLValidation` seems the most descriptive (I prefer skip vs. skips, but that's no biggie). I understand wanting to follow `curl` in using `insecure`, but consistency between command flag and config is more important here I think.

packages/apollo-cli/src/fetch-schema.ts
```diff
@@ -56,15 +51,35 @@ export async function fromFile(
...
   { url, headers }: EndpointConfig,
-  projectFolder?: string
+  projectFolder?: string,
+  insecureEnabled?: boolean
 ): Promise<GraphQLSchema | undefined> => {
   if (!url) throw new Error("No endpoint provided when fetching schema");
   const filePath = projectFolder ? path.resolve(projectFolder, url) : url;
   if (fs.existsSync(filePath)) return fromFile(filePath);
 
+  const insecureOptionActive = insecureEnabled ? insecureEnabled : false;
```
martijnwalraven: I think we can avoid an extra variable here by using a default parameter value (`insecureEnabled?: boolean = false`).

packages/apollo-cli/src/load-schema.ts
```diff
@@ -17,7 +17,7 @@ export async function loadSchema(
 
   if (dependency.endpoint && dependency.endpoint.url) {
     try {
-      return await fetchSchema(dependency.endpoint, config.projectFolder);
+      return await fetchSchema(dependency.endpoint, config.projectFolder, dependency.skipsSSLValidation);
```
martijnwalraven: If we put `skipSSLValidation` under `endpoint` we can avoid passing in an extra parameter.

packages/apollo-cli/src/config.ts
robertomg: I think this is a great suggestion and actually I thought about implementing this way on the first place. 👍

packages/apollo-cli/src/commands/schema/download.ts
robertomg: I agree on this that giving a more self-explanatory name is more important than keeping old names. So I decided to change this (also to skip instead of skips). Thanks for the feedback

packages/apollo-cli/src/fetch-schema.ts
robertomg: Just discovered that checking a (`boolean | undefined`) variable on an if would only accept it if the variable is `true` so this line is not necessary anymore.

packages/apollo-cli/src/load-schema.ts
robertomg: Done

martijnwalraven:
Thanks for the quick response!

PR #578, opened 2018-09-04
Issue #215 - Make Swift Enums conform to Hashable

bclymer:
# What

Makes `enum`s conform to `Hashable`. The actual conformance is automatically synthesized by the Swift 4.1 compiler.

# Why

In response to https://github.com/apollographql/apollo-cli/issues/215

# Testing

In my local environment I was able to verify simply adding `Hashable` worked as expected, and it could be used in external code as if it was `Hashable`. As long as CI's tests pass, this should be good to go.

# Problems

Swift 4.1 only exists in Xcode 9.3, and not 9.2 or below. From what I can see, Apollo still supports "Xcode 9 and Swift 4", so that will no longer be completely true. However, Swift 4.1 is a non-source-breaking change, and having `enum`s automatically implement `Hashable` is a nice win.

- [x] feature
- [ ] blocking
- [ ] docs


bclymer:
@meteor-bot done.

PR #579, opened 2018-09-04
Make operationDefinition on GraphQLOperation static.

bclymer:
# What

This makes `operationDefinition` on `GraphQLOperation` `static`. 

# Why

Since it is already a constant, there are a couple reasons why making it `static` is nice.

1. It's not allocated each time the class is instantiated.
1. It allows accessing the string outside of the context of an instance.

The latter is the main reason I'm making this PR. We were previously using `requestString` from the older `apollo-codegen` to detect changes in the query in order to invalidate local caches. Ignoring the fact that we could have a better way of determining that, I think this property _should_ be `static` anyway. 

# Problems

This requires a change in apollo-ios as well. Here's my PR -- https://github.com/apollographql/apollo-ios/pull/363


designatednerd:
Closing per @bclymer's request [here](https://github.com/apollographql/apollo-ios/pull/363)

PR #582, opened 2018-09-07
Add schema:print command

evans:
The command receives a json file input and prints the schema

Future extensions could include printing the schema directly from Engine or adding flags to limit the output to certain fields/types/remove comments.


evans:
@jbaxleyiii It prints the SDL corresponding to the introspection.json. When I was working on Engine, I found that I wanted to look at the schema and couldn't understand the json dump. This was a nice way to explore the schema. I specifically ran something like: `npx apollo schema:print ~/engine/frontend/imports/shared/graphql/schema.json  | less`

PR #584, opened 2018-09-08
Possibility to force camel casing

tomitrescak:

### DIFF
--- a/packages/apollo-codegen-typescript/src/language.ts
+++ b/packages/apollo-codegen-typescript/src/language.ts
@@ -153,7 +153,12 @@ export default class TypescriptGenerator {
   }
 
   public nameFromScopeStack(scope: string[]) {
-    return scope.join("_");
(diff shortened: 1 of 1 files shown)


jbaxleyiii:
@tomitrescak what are you wanting to accomplish here?

tomitrescak:
Well, I would like to have class names as per recommendations. The package is awesome, but seeing class names with a weird snake case with lowercase letters breaks my convention heart. It’s just that I never seen classes such as My_snake_class. It would be great to have a parameter that decides this.

trevor-scheer:
@tomitrescak sorry for the delay on follow up, I'm trying to triage PRs and issues as I'm able but we're quite busy here.

This is more of a feature request than an actual PR - would you mind opening an issue as such?
Else, a PR with a more complete implementation would also be entertained, but I'd prefer we have the chance to align on any decisions first. Thanks!

tomitrescak:
Can be ingnored since the graphql-code-generator came to exist.

PR #591, opened 2018-09-13
Fix loadConfigFromFile

WaldoJeffers:
Fixes https://github.com/apollographql/apollo-cli/issues/549

Current implementation resolves the config file as a node module, making it impossible to actually pass a config file


shadaj:
@WaldoJeffers this looks great, but looks like linting is failing. Could you run `npm run lint-fix` from the base `apollo-cli` director?

WaldoJeffers:
@shadaj Sure thing :)

PR #600, opened 2018-09-20
Migrate stats query to druid, update how the data is consumed

trevor-scheer:
This PR cuts the stats query over to Druid. Also add watch support for better extension development.

### DIFF
--- a/packages/apollo-language-server/src/project.ts
+++ b/packages/apollo-language-server/src/project.ts
@@ -73,15 +73,14 @@ const fileAssociations: { [extension: string]: string } = {
 const engineStatsQuery = gql`
   query EngineSchemaStats($id: ID!) {
     service(id: $id) {
-      report(from: "-3600", to: "-0") {
-        usageStats {
-          types {
-            name
-            fields {
-              name
-              latencyHistogram {
-                serviceTimeP95: durationMs(percentile: 0.95)
-              }
+      stats(from: "-3600", to: "-0") {
+        fieldStats {
+          groupBy {
+            field
+          }
+          metrics {
+            fieldHistogram {
+              durationMs(percentile: 0.95)
             }
           }
         }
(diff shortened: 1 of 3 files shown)


jbaxleyiii:
@trevor-scheer this looks great!

packages/apollo-language-server/src/project.ts
```diff
@@ -201,27 +200,44 @@ export class GraphQLProject {
...
-                  fieldsMap.set(
-                    field.name,
-                    field.latencyHistogram.serviceTimeP95
-                  );
-                });
-
-                schemaEngineStats.set(typ.name, fieldsMap);
+            engineData.data!.service.stats.fieldStats.forEach(
+              (fieldStat: FieldStat) => {
+                // Parse field "ParentType.fieldName:FieldType" into ["ParentType", "fieldName", "FieldType"]
```
jbaxleyiii: Whats going on with this? How does this data come back? Is this a change we should add to the server?

packages/apollo-language-server/src/project.ts
```diff
@@ -201,27 +200,44 @@ export class GraphQLProject {
...
             );
 
             this.engineStats.set(schemaDef.engineKey, schemaEngineStats);
-
-            return;
-          } else {
-            return;
           }
+
+          return;
```
evans: Is this necessary?

packages/apollo-language-server/src/project.ts
trevor-scheer: Good question! I just had some discussion with Danielle and @evans about this and learned quite a bit. The shape of this query relates directly to Druid and the utility we use that generates resolvers for druid queries. So...I do agree that parsing this on the frontend isn't ideal but I don't think it justifies a change to the resolvers. Did I understand your question correctly?

trevor-scheer: @jbaxleyiii I'm looking to merge soon, can I resolve this or did you want to dig any deeper?

PR #605, opened 2018-09-28
Remove run query feature for now, to be re-introduced in the future.

trevor-scheer:
This PR is to remove the WebView (run query feature) from the VSCode extension. An issue to reintroduce the work will be opened after this is merged, and will point to the commit / PR for reference.


packages/apollo-language-server/src/languageProvider.ts
```diff
@@ -377,106 +372,4 @@ ${argumentNode.description}
...
-                definition.operation === "subscription"
-                  ? set.endpoint.subscriptions
-                  : set.endpoint.url,
-                set.endpoint.headers,
-                graphql.printSchema(set.schema!)
-              )
-            });
-          }
-        } else if (definition.kind === Kind.FRAGMENT_DEFINITION) {
-          const references = project.fragmentSpreadsForFragment(
```
shadaj: This part should probably be kept, it provides the `n references` codelens for fragments.

trevor-scheer: Nice catch and thanks for taking a look!

trevor-scheer:
Bumping this @jbaxleyiii - no actual hurry, but I'd like to merge this. @shadaj if you wanted to take another pass I would be 🙏  but if you're busy I understand!

trevor-scheer:
Thanks @shadaj!

PR #611, opened 2018-10-02
Restructure `apollo queries:extract` logic and manifest format.

abernix:
**Breaking changes:** This changes the output format of the manifest generated by `queries:extract` to be in a more evolvable format.  Existing users of the manifest produced by `manifest.json` should take care to observe the new manifest format, as indicated by the top-level presence of a `version` attribute (now `1`, previously `undefined`).

In general, this just moves some bits of often repeated portions of the CLI into a unified location where they can be shared by more modules as we use more and more portions in multiple places.

### DIFF
--- a/packages/apollo/src/commands/queries/check.ts
+++ b/packages/apollo/src/commands/queries/check.ts
@@ -5,11 +5,7 @@ import * as Listr from "listr";
 import { toPromise, execute } from "apollo-link";
 import { print, GraphQLError, DocumentNode } from "graphql";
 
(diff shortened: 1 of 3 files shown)


abernix:
@trevor-scheer 	Happy to assist/adjust if this breaks anything you currently have in-flight.

trevor-scheer:
@abernix @jbaxleyiii nothing in flight that would be affected AFAIK, taking a look now to be sure. Thanks for the heads up!

PR #613, opened 2018-10-04
Revert "Fall back to other methods of schema loading... (#533)

### LINKED ISSUE #560: Error message of schema loading unclear when from server without introspection
Regression of #533 

Error message is now:

```
$ apollo schema:download --endpoint ... schema.json
 ✔ Loading Apollo config
 ✔ Fetching current schema
 ✖ Saving schema to schema.json
   → Expected undefined to be a GraphQL schema.
Error: Expected undefined to be a GraphQL schema.
```

previously it was:

```
$ apollo schema:download --endpoint ... schema.json
 ✔ Loading Apollo config
 ✖ Fetching current schema
   → Response not successful: Received status code 400
   Saving schema to schema.json
Error: Response not successful: Received status code 400
```

Is it now unclear why the command failed.

abernix:
This reverts commit b48e7bf31ebb9f5cb8f4daa7830a8fb8c4c81d45.

As corroborated by the trending number of 👍 emotes on #560, this PR introduced a number of `try` `catch`es which silenced relatively useful information from their _tried_ operations.  I'm happy to help figure out a better approach but since this seems to have caused many more problems than it solved, we should consider a different approach.

cc @shadaj 

Closes #560

### DIFF
--- a/packages/apollo/src/load-schema.ts
+++ b/packages/apollo/src/load-schema.ts
@@ -7,28 +7,15 @@ export async function loadSchema(
   config: ApolloConfig
 ): Promise<GraphQLSchema | undefined> {
   if (dependency.schema) {
-    try {
-      return await fetchSchema(
-        { url: dependency.schema },
-        config.projectFolder
-      );
-    } catch {}
+    return await fetchSchema({ url: dependency.schema }, config.projectFolder);
+  } else if (dependency.endpoint && dependency.endpoint.url) {
+    return await fetchSchema(dependency.endpoint, config.projectFolder);
+  } else if (dependency.engineKey) {
+    return await fetchSchemaFromEngine(
+      dependency.engineKey,
+      config.engineEndpoint
+    );
+  } else {
+    return undefined;
   }
-
-  if (dependency.endpoint && dependency.endpoint.url) {
-    try {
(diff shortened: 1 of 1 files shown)


packages/apollo/src/load-schema.ts
```diff
@@ -7,28 +7,15 @@ export async function loadSchema(
   config: ApolloConfig
 ): Promise<GraphQLSchema | undefined> {
   if (dependency.schema) {
-    try {
-      return await fetchSchema(
-        { url: dependency.schema },
-        config.projectFolder
-      );
-    } catch {}
```
abernix: Throwing away all errors which occur from the `fetchSchema` method buries not only just the mere failure to acquire the schema in _some_ manner, but it also neglects to reveal very important information about HTTP transport errors, file system errors, GraphQL execution and more — all of which happen via `fetchSchema`!

shadaj:
This makes sense given the difficulty of tracing the root error. Maybe a future change could handle this better by aggregating the individual errors when no available sources work for the returned error? But this looks good otherwise!

packages/apollo/src/load-schema.ts
aodinok: Faced this yesterday, because of this catch - SSL certificate error was swallowed.

PR #614, opened 2018-10-04
Avoid re-introspecting within local `graphql` context and exclude client-only directives.

abernix:
~**Note:** This PR depends on #611 and #613 and merges into #611.  If #611 lands first, this should be re-targeted to land in `master`.~

---

The commits in this PR avoid re-running the introspection result from a GraphQL server as a client schema when not necessary.  This is particularly important to maintain an introspection result which is consistent with the version of `graphql` which is used on the actual GraphQL server being introspected.  Previously, this was re-running that introspection in the Apollo CLI’s `graphql` executor which produced a different result since there is no guarantee that the server is using `graphql@0.13.2`, the version currently used by the `apollo` CLI.

Additionally, thanks to @jbaxleyiii’s work in 88a248c, operations in the manifest now properly consider common dynamic transformations to operations like client-only directives (e.g. `@client`, `@connection`) and additions of `__typename` fields, as is common under normal Apollo Client use.

### DIFF
--- a/packages/apollo/src/config.ts
+++ b/packages/apollo/src/config.ts
@@ -3,7 +3,13 @@ import { fs, withGlobalFS } from "apollo-codegen-core/lib/localfs";
 
 import * as fg from "glob";
 import * as minimatch from "minimatch";
-import { GraphQLSchema, extendSchema, visit, buildASTSchema } from "graphql";
+import {
+  GraphQLSchema,
+  extendSchema,
+  visit,
+  buildASTSchema,
+  buildClientSchema
+} from "graphql";
 import { loadSchema } from "./load-schema";
 import { loadQueryDocuments } from "apollo-codegen-core/lib/loading";
 
@@ -232,7 +238,10 @@ export async function resolveSchema(
       )
     : referredSchema.clientSide
       ? buildASTSchema(loadAsAST())
-      : await loadSchema(referredSchema, config);
+      : await loadSchema(referredSchema, config).then(introspectionResult => {
+          if (!introspectionResult) return;
+          return buildClientSchema({ __schema: introspectionResult });
+        });
 }
 
 export async function resolveDocumentSets(
--- a/packages/apollo/src/config.ts
+++ b/packages/apollo/src/config.ts
@@ -238,9 +238,9 @@ export async function resolveSchema(
       )
     : referredSchema.clientSide
       ? buildASTSchema(loadAsAST())
-      : await loadSchema(referredSchema, config).then(introspectionResult => {
-          if (!introspectionResult) return;
-          return buildClientSchema({ __schema: introspectionResult });
+      : await loadSchema(referredSchema, config).then(introspectionSchema => {
+          if (!introspectionSchema) return;
+          return buildClientSchema({ __schema: introspectionSchema });
         });
 }
 
--- a/packages/apollo/src/fetch-schema.ts
+++ b/packages/apollo/src/fetch-schema.ts
@@ -81,17 +81,22 @@ export const fetchSchema = async (
     options.fetchOptions = { agent: agent };
   }
 
-  return toPromise(
-    // XXX node-fetch isn't compatible typescript wise here?
+  const { data, errors }: any = await toPromise(
     linkExecute(createHttpLink(options), {
       query: introspection,
       context: { headers }
     })
-  ).then(({ data, errors }: any) => {
-    if (errors)
-      throw new Error(errors.map(({ message }: Error) => message).join("\n"));
-    return buildClientSchema({ __schema: data.__schema });
-  });
+  );
+
+  if (!data) {
+    throw new Error("No data received from server introspection.");
+  }
+
+  if (errors) {
+    throw new Error(errors.map(({ message }: Error) => message).join("\n"));
+  }
+
+  return buildClientSchema({ __schema: data.__schema });
 };
 
 export async function fetchSchemaFromEngine(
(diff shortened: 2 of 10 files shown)


packages/apollo/src/fetch-schema.ts
```diff
@@ -18,47 +20,74 @@ import { EndpointConfig } from "./config";
...
 
-const introspection = gql(introspectionQuery);
+const introspection = gql(getIntrospectionQuery());
 
-export async function fromFile(
+async function buildIntrospectionSchemaInLocalGraphQLContext(
+  source: string | Source
+): Promise<IntrospectionSchema> {
+  const schema: GraphQLSchema = buildSchema(source);
+  const executionResult: ExecutionResult = await graphql(
```
martijnwalraven: I know this is unrelated to this PR, but I still find it confusing to import `execute` from `graphql-js` as `graphql`, while there is also a `graphql` export.

packages/apollo/src/config.ts
```diff
@@ -232,7 +238,10 @@ export async function resolveSchema(
       )
     : referredSchema.clientSide
       ? buildASTSchema(loadAsAST())
-      : await loadSchema(referredSchema, config);
+      : await loadSchema(referredSchema, config).then(introspectionResult => {
```
martijnwalraven: I think `introspectionResult` is actually `IntrospectionSchema`, so maybe the name should reflect that.

packages/apollo/src/fetch-schema.ts
```diff
@@ -18,47 +20,74 @@ import { EndpointConfig } from "./config";
...
+    schema,
+    getIntrospectionQuery()
+  );
+
+  if (executionResult.errors) {
+    console.error(executionResult.errors);
+    throw new Error("No data received during introspection query execution.");
```
martijnwalraven: Should this report the errors as part of the thrown `Error`?

packages/apollo/src/fetch-schema.ts
```diff
@@ -18,47 +20,74 @@ import { EndpointConfig } from "./config";
 import { engineLink, getIdFromKey } from "./engine";
 import { SCHEMA_QUERY } from "./operations/schema";
 
-const introspection = gql(introspectionQuery);
+const introspection = gql(getIntrospectionQuery());
 
-export async function fromFile(
+async function buildIntrospectionSchemaInLocalGraphQLContext(
```
martijnwalraven: I find the naming of this function a bit confusing. Maybe `buildIntrospectionSchemaFromSDL`?

packages/apollo/src/fetch-schema.ts
```diff
@@ -81,30 +110,35 @@ export const fetchSchema = async (
...
       context: { headers }
     })
-  ).then(({ data, errors }: any) => {
-    if (errors)
-      throw new Error(errors.map(({ message }: Error) => message).join("\n"));
-    return buildClientSchema({ __schema: data.__schema });
-  });
+  );
+
+  if (!data) {
```
martijnwalraven: Should the `data` and `error` checks be reversed, so we always report errors if there are any?

packages/apollo/src/fetch-schema.ts
abernix: This is actually no longer the case and this _really_ is `graphql` from `graphql`.  I did try to fix that myself by aliasing `execute` to `graphqlExecute` instead, but @jbaxleyiii 	took it a step further and just used `graphql`.

That said, what you're saying was still true when this logic still lived in [`schema/publish`](https://github.com/apollographql/apollo-cli/pull/614/files/c3054817ecde88f2ec4af6781c2368bb4c71e5e1#diff-1d42d3c5aff25031f2706605c8de0bb9L9) and [`schema/check`](https://github.com/apollographql/apollo-cli/pull/614/files/c3054817ecde88f2ec4af6781c2368bb4c71e5e1#diff-2364a0900ee1fb8e10783fc35eebddceL10).

packages/apollo/src/fetch-schema.ts
abernix: Fixed via d31f9a8361f18dee7b8186a6fc8c292c37aa3b7.

packages/apollo/src/fetch-schema.ts
abernix: Yes, I'll add that.  I actually started refactoring to use a more global helper implementation of a common pattern across this whole repo:

```js
errors.map(({ message }: Error) => message).join("\n"))
```

But I stopped short due to time.  Thanks for raising this very correct suggestion.

abernix: Fixed via 72bf1540ea244cff912c7d15919aa27ae7e9040a.

packages/apollo/src/fetch-schema.ts
abernix: Makes sense to me.  Fixed via 9be573a45302af8302cffb6de36f61e6be721b41.

packages/apollo/src/config.ts
abernix: Just saw this!  Fixed via 8e144a1120a63ae6a1aecfb73231b64eddc3f668.

jbaxleyiii: @martijnwalraven great catch on that! That difference (between `IntrospectionSchema` and `IntrospectionQuery` tripped me up a few times). Thanks @abernix for the change!

trevor-scheer:
Thanks for adding me @abernix, reading PRs like this will help me gain context much more quickly 😄

PR #618, opened 2018-10-05
Filter type definitions

trevor-scheer:
When building up an array of diagnostics in the language server, we should handle type definitions separately.

This PR narrows the scope of the `collectDiagnostics` function to be responsible for operations only.
This PR does not implement diagnostics for type defs, this just removes a false positive.

Still trying to figure out the best way to a good `watch` flow while working on the extension. Updated some of the scripts/dev config in the language server and extension packages. This simplifies it a bit - all that is required now is a symlink from `apollo-language-server/lib` to `apollo-vscode/server`. Still a WIP, but it feels like a step in the right direction. I'm _very_ open to feedback here.

Before:
screen shot 2018-10-05 at 11 20 38 am

After:
screen shot 2018-10-05 at 11 19 42 am

### DIFF
--- a/packages/apollo-language-server/src/__tests__/diagnostics.ts
+++ b/packages/apollo-language-server/src/__tests__/diagnostics.ts
@@ -1,0 +1,72 @@
+import { Source } from "graphql";
+import { loadSchema } from "apollo-codegen-core/lib/loading";
+import { GraphQLDocument } from "../document";
+import { collectExecutableDefinitionDiagnositics } from "../diagnostics";
+const schema = loadSchema(
+  require.resolve("../../../common-test/fixtures/starwars/schema.json")
+);
+const validDocument = new GraphQLDocument(
+  new Source(`
+    query HeroAndFriendsNames {
+      hero {
+        name
+        friends {
+          name
+        }
+      }
+    }`)
+);
+const invalidDocument = new GraphQLDocument(
+  new Source(`
+    query HeroAndFriendsNames {
+      hero {
+        nam         # Missing letter 'e'
+        friend {    # Missing letter 's'
+          name
+        }
+      }
+    }`)
+);
+const documentWithTypes = new GraphQLDocument(
+  new Source(`
+    type SomeType {
+      thing: String
+    }
+     enum SomeEnum {
+      THING_ONE
+      THING_TWO
+    }
+     query HeroAndFriendsNames {
+      hero {
+        name
+        friends {
+          name
+        }
+      }
+    }`)
+);
+describe("Language server diagnostics", () => {
+  describe("#collectExecutableDefinitionDiagnositics", () => {
+    it("returns no diagnostics for a correct document", () => {
+      const diagnostics = collectExecutableDefinitionDiagnositics(
+        schema,
+        validDocument
+      );
+      expect(diagnostics.length).toEqual(0);
+    });
+    it("returns two diagnostics for a document with two errors", () => {
+      const diagnostics = collectExecutableDefinitionDiagnositics(
+        schema,
+        invalidDocument
+      );
+      expect(diagnostics.length).toEqual(2);
+    });
+    it("returns no diagnostics for a document that includes type definitions", () => {
+      const diagnostics = collectExecutableDefinitionDiagnositics(
+        schema,
+        documentWithTypes
+      );
+      expect(diagnostics.length).toEqual(0);
+    });
+  });
+});
--- a/packages/apollo-language-server/src/diagnostics.ts
+++ b/packages/apollo-language-server/src/diagnostics.ts
@@ -2,7 +2,8 @@ import {
   GraphQLSchema,
   GraphQLError,
   FragmentDefinitionNode,
-  findDeprecatedUsages
+  findDeprecatedUsages,
+  isExecutableDefinitionNode
 } from "graphql";
 
 import { Diagnostic, DiagnosticSeverity } from "vscode-languageserver";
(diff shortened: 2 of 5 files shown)


packages/apollo-language-server/src/diagnostics.ts
```diff
@@ -11,25 +13,43 @@ import { GraphQLDocument } from "./document";
...
 
+import { TypeKind } from "apollo/lib/printer/ast";
 import { getValidationErrors } from "apollo/lib/validation";
 
-export function collectDiagnostics(
+/**
+ * Build an array of code diagnostics for all GraphQL operations in a document.
+ * This function explicitly ignores type definitions.
+ */
+export function collectOperationDiagnostics(
```
martijnwalraven: Note that this won't just collect operation definitions, but also fragment definitions (as it should). I think the right term to describe both is 'executable definitions'. So maybe `collectExecutableDefinitionDiagnositics` (I realize that's quite the mouthful)?

packages/apollo-language-server/src/diagnostics.ts
```diff
@@ -11,25 +13,43 @@ import { GraphQLDocument } from "./document";
...
   schema: GraphQLSchema,
   queryDocument: GraphQLDocument,
   fragments: { [fragmentName: string]: FragmentDefinitionNode }
 ): Diagnostic[] {
   const ast = queryDocument.ast;
   if (!ast) return queryDocument.syntaxErrors;
 
+  // I would prefer to use the TypeKind enum from graphql-js here,
+  // but the enum values aren't the same, i.e. "OBJECT" vs "ObjectTypeDefinition"
+  const typeKinds = Object.values(TypeKind);
```
martijnwalraven: You should be able to import `Kind` from `graphql` and check against those values. But you probably want to use utility methods like `isTypeSystemDefinitionNode` and `isTypeSystemExtensionNode` instead, so you don't have to enumerate all possible values.

packages/apollo-language-server/src/diagnostics.ts
```diff
@@ -11,25 +13,43 @@ import { GraphQLDocument } from "./document";
...
+  const astWithoutTypeDefinitions = visit(ast, {
+    enter(node: ASTNode) {
```
martijnwalraven: Note that `visit` also allows you to specify specific nodes to visit, so you don't have to perform a manual check. See [here](https://github.com/apollographql/apollo-cli/blob/c50691d8c8fdb0e2beea817052c28ccf5887ea8a/packages/apollo-language-server/src/project.ts#L469) for example. Or [here](https://github.com/graphql/graphql-js/blob/26c9874107e65f19576aae0a32638287820f68aa/src/language/visitor.js#L142) for the `visit` docs, which describe all usage patterns.

martijnwalraven: In this case however, a visitor might be overkill. You can simply loop over `ast.definitions` and either filter out type definitions and extensions, or only keep operation and fragment definitions.

packages/apollo-language-server/src/diagnostics.ts
martijnwalraven: Ah, it turns out there is also `isExecutableDefinitionNode`, which checks for operation and fragment definitions.

packages/apollo-language-server/src/diagnostics.ts
martijnwalraven: I think this could actually be simplified to `ast.definitions.map(isExecutableDefinitionNode)`!

trevor-scheer: I like that a lot! Thanks for the info and the links, I'll take a look.

packages/apollo-language-server/src/diagnostics.ts
trevor-scheer: I'd rather be on the more verbose side if it provides clarity. Thanks for the clarity 👍

packages/apollo-language-server/src/__tests__/diagnostics.ts
```diff
@@ -0,0 +1,81 @@
...
+        friends {
+          name
+        }
+      }
+    }`)
+);
+
+describe("Language server diagnostics", () => {
+  describe("collectExecutableDefinitionDiagnositics", () => {
+    test("Returns no diagnostics for a correct document", () => {
```
martijnwalraven: Thanks for taking the time to write tests, we should definitely do more of that!

In terms of style, I personally prefer what is sometimes called spec style, where you use `it` instead of `test` and the combined message reads as a description of behavior. That actually already works well here: `it("returns no diagnostics for a correct document"` would read as 'collectExecutableDefinitionDiagnositics returns no diagnostics for a correct document'.

martijnwalraven: We don't have real testing guidelines in place unfortunately, so that might be something we'd want to do. If you're looking for examples of tests, you could look [at this file](https://github.com/apollographql/apollo-cli/blob/master/packages/apollo-codegen-core/src/compiler/visitors/__tests__/conditionalFields.ts) and other tests that are part of codegen. One nice addition there is the use of a custom matcher `toMatchSelectionSet` (see [here](https://github.com/apollographql/apollo-cli/blob/master/packages/apollo-codegen-core/test-utils/matchers.ts)). Custom matchers can make both code and failure output a lot more readable (and therefore actionable). There's also a growing collection of [built-in matchers](https://jestjs.io/docs/en/expect), so you may want to have a look there.

trevor-scheer: Thanks for showing me these custom matchers - are you suggesting they have an application here? I looked into them a bit and it didn't seem quite what I was looking for, since I'm just confirming that validation messages come back. I could do a direct string comparison on the messages if you prefer the test be more specific than "confirm there are 2 errors in the result"?

trevor-scheer:
@martijnwalraven this should be ready for another / probably final pass 👍

trevor-scheer:
Bump if you find time for this tomorrow @martijnwalraven

martijnwalraven:
@trevor-scheer This looks great!

PR #624, opened 2018-10-09
Update graphql to version 14

trevor-scheer:
This PR is to upgrade our usage of the graphql library to v14 🎉 

@abernix: tests pass, but what needs to be done to make sure this isn't breaking anything around schema publishing? (IIRC, this was the concern you raised)


trevor-scheer:
Looks like quite a few typescript errors showed up that weren't popping in vscode - will address these soon, feel free to disregard until then.

abernix:
@trevor-scheer The test seemed to be spurious and went away when I re-ran it.  If you don't mind testing the workflow I was concerned about able to:

1. Run a GraphQL server
2. Try running a `schema:publish` against that running server, with the currently stable/published version of `apollo`.
3. Try running a `schema:check` against that same running server with no other schema changes with _this_ version.
4. Finally, try running a `schema:publish` against that same running server, again with no other schema changes.

I'm mainly curious if the:

* `id` changes in the output. (I suspect it will)
* If any diff is displayed. (I hope it won't)
* Any other weirdness.

trevor-scheer:
@abernix:
* Seems that `id` is the same across publishes, that's not what you suspected but I don't know what the desirable outcome is here. I imagine this is a best case scenario.
* No diff!

Steps:
1) Cloned fullstack workshop, installed stable cli, ran `schema:publish` for a new engine service
stableschemapublish
2) Sanity check `schema:check` with stable
stablecheckschema
3) `npm link` from current working branch, `npm link apollo` from workshop root
4) `schema:check` with working branch
v14schemacheck
5) `schema:publish` with working branch
v14schemapublish

6) Bonus Engine sanity check:
engineresult

daggerjames:
Please consider put graphql into peerDependencies. [reference](https://medium.com/@leeb/graphql-js-preparing-for-v14-0-0-839f823c144e)

daggerjames:
Also, please consider problem I described in #629, which would also prevent `apollo-cli` to get rid of duplicated `graphql` dependencies

abernix:
@daggerjames I hope we can help solve your problem, but `graphql` is not a _companion library_ ti the `apollo` CLI — it's directly depended upon.

Ref from your reference above:
> tool (like graphql-cli) then the normal dependencies field is still appropriate.

If you have `apollo` in your project, we presume it's specified within `devDependencies`.  With proper bundling, you will not end up with two copies, and unless you directly depend on `apollo` within your application (e.g. `import 'apollo', or `require('apollo')`), which we have no reason to believe you would need to, then there should be no overlaps in execution (for example, via [`npx`](https://npm.im/npx) or [`npm-scripts`](https://docs.npmjs.com/misc/scripts)).  Either way, let's please continue the discussion in another issue not in this PR!  Thanks!

abernix:
@trevor-scheer Full disclosure: I should have made two of my instructions more clear:

1. The server that's being introspected should be using `graphl@0.13.x` and the second `schema:publish` should also be from this branch.  Given that [`fullstack-workshop-server`](https://github.com/apollographql/fullstack-workshop-server) comes out of the box with `graphql@0.13.x`, unless you upgraded it to `graphql@14`, then things should have been fine.
2. The second `schema:publish` should be done from this branch, just like the previous `schema:check`.  Looking at your above example, I think you did this anyway, so should be okay in that regard.

That said, I was quite surprised that `id` was not different, so I gave it a try with an even more simple example, also using `fullstack-workshop-server` which I simply `npm start`ed, followed by:

image

So... is it possible that you've updated the version of `graphql` that your `fullstack-workshop-server`'s `graphql` to `14.x`?

package-lock.json
```diff
@@ -2128,7 +2128,7 @@
         },
         "strip-ansi": {
           "version": "3.0.1",
-          "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
+          "resolved": "http://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
```
abernix: One thing, could you — this one time — run:

```
npm cache clear --force
```

and then run `npm install` again and see if it changes these back to `https`?

PR #632, opened 2018-10-18
VSCode: Schema tag switching

trevor-scheer:
This PR:
* adds functionality to the VSCode extension for switching to / developing against different schema tags.
* adds the ability to publish a schema to engine with a tag via the CLI and the `--tag` flag.

For context:
Schemas can be "branched" in a similar way that one might make a branch in a git repo. This concept allows developers to publish different "branches" (tags) of a schema to engine. Publishing a new tag to engine makes the new introspection available to a client, in this case, VSCode.

The result here is that an extension user can switch tags on the fly and expect all the benefits that an introspection provides: autocomplete, syntax errors, and other relevant information _specific to that tag_.

oct-18-2018 15-47-34

### DIFF
--- a/packages/apollo-language-server/src/project.ts
+++ b/packages/apollo-language-server/src/project.ts
@@ -89,6 +89,16 @@ const engineStatsQuery = gql`
   }
 `;
 
+const schemaTagsQuery = gql`
+  query SchemaTags($id: ID!) {
+    service(id: $id) {
+      schemaTags {
+        tag
+      }
+    }
+  }
+`;
+
 export interface DocumentAndSet {
   doc: GraphQLDocument;
   set: ResolvedDocumentSet;
@@ -98,6 +108,7 @@ export class GraphQLProject {
   public config: ApolloConfig = (null as any) as ApolloConfig;
   private _onDiagnostics?: NotificationHandler<PublishDiagnosticsParams>;
   private _onDecorations?: (any: any) => void;
+  private _onSchemaTags?: (tags: Map<string, string[]>) => void;
 
   public isReady = false;
   public readyPromise: Promise<void> | undefined = undefined;
@@ -113,6 +124,7 @@ export class GraphQLProject {
     string,
     Map<string, Map<string, number>>
   > = new Map();
+  private schemaTags: Map<string, string[]> = new Map();
 
   constructor(
     config: ApolloConfig,
@@ -135,6 +147,7 @@ export class GraphQLProject {
 
     this.documentSets = undefined;
     this.engineStats.clear();
+    this.schemaTags = new Map();
     this.documentsByFile = new Map();
     this.setToResolved.clear();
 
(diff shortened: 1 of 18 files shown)


packages/apollo-language-server/src/project.ts
```diff
@@ -179,6 +200,38 @@ export class GraphQLProject {
     return !!this.setThatIncludes(filePath);
   }
 
+  async loadSchemaTags() {
+    await this.loadingHandler.handle(
+      `Loading available schema tags for ${this.config.name!}`,
+      Promise.all(
+        Object.values(this.config.schemas!).map(async schemaDef => {
+          if (schemaDef.engineKey) {
```
jbaxleyiii: should we not even go into loading if we don't have an engineKey for any service? And instead flag a missing value

packages/apollo-language-server/src/project.ts
```diff
@@ -241,12 +294,50 @@ export class GraphQLProject {
     );
   }
 
+  async updateSchemaTag(tag: string) {
+    await this.loadingHandler.handle(
+      `Loading queries and schemas for ${this.config.name!}`,
+      (async () => {
+        this.documentSets = await resolveDocumentSets(this.config, true, tag);
+        for (const set of this.documentSets) {
+          if (!set.schema!.getQueryType()!.astNode) {
```
jbaxleyiii: lots of type overrides here, are we sure we won't hit errors?

packages/apollo/src/config.ts
```diff
@@ -82,7 +82,7 @@ function loadSchemaConfig(
       obj.endpoint,
       !obj.engineKey && defaultEndpoint
     ),
-    engineKey: obj.engineKey,
+    engineKey: obj.engineKey || process.env.ENGINE_API_KEY,
```
jbaxleyiii: should `env` take precedent?

jbaxleyiii:
@trevor-scheer one thing else it would probably be better for this to be a command palate command instead of a click event on the status bar

packages/apollo-language-server/src/project.ts
trevor-scheer: a8f44782656fd1993482baee82c58f280c3a11f8

packages/apollo-language-server/src/project.ts
trevor-scheer: d0a34efa1371fdbc080a1c931027c79d3cb8a898

trevor-scheer:
@jbaxleyiii command palette change: 6920ea2db34701687a0eb97c59e8b457d4e4b6fe

packages/apollo/src/config.ts
trevor-scheer: f25428ed1f60a5f66f4f3f906f97052a2047a35f

packages/apollo-language-server/src/project.ts
```diff
@@ -135,20 +151,35 @@ export class GraphQLProject {
...
-        console.error(error);
-      });
+    this.readyPromise = process.env.ENGINE_API_KEY
+      ? this.loadEngineData()
+          .then(() => {
+            this.scanAllIncludedFiles();
+            this._onSchemaTags && this._onSchemaTags(this.schemaTags);
+          })
+          .catch(error => {
+            console.error(error);
```
shadaj: Is this an error that should be send to the loading handler? Otherwise I think the the indicator would get stuck.

PR #645, opened 2018-10-27
VS Code extension build configuration fixes

martijnwalraven:
This PR fixes the build configuration so the VS Code extension can be packaged correctly with its dependencies.


trevor-scheer:
LGTM, this is great!

shadaj:
Awesome! Glad the `copy-apollo-libs.sh` hack could be removed as well.

PR #653, opened 2018-11-02
VSCode Loading State / Error Messages (on init)

trevor-scheer:
This PR improves messaging to the user while the extension boots up.

Note: the extension doesn't activate w/o an `apollo.config.js` file present, so assume one exists.

* Warn user if no .env file is found
* Warn user if no ENGINE_API_KEY is specified (when .env file is present)
* Only display one message while initial loading promises resolve.

* Error specifically if extension fails to scan files
* Error specifically if engine stats fail to load.

All errors and warnings are actionable, but feel free to suggest improvements on the copy.

### DIFF
--- a/packages/apollo-vscode/src/extension.ts
+++ b/packages/apollo-vscode/src/extension.ts
@@ -1,5 +1,5 @@
 import { join, resolve } from "path";
-import { readFileSync } from "fs";
+import { readFileSync, existsSync } from "fs";
 import {
   window,
   workspace,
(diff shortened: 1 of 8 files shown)


packages/apollo-vscode/src/extension.ts
```diff
@@ -74,7 +84,8 @@ export function activate(context: ExtensionContext) {
       fileEvents: [
         workspace.createFileSystemWatcher("**/apollo.config.js"),
         workspace.createFileSystemWatcher("**/package.json"),
-        workspace.createFileSystemWatcher("**/*.{graphql,js,ts,jsx,tsx}")
+        workspace.createFileSystemWatcher("**/*.{graphql,js,ts,jsx,tsx}"),
```
stevenpetryk: Even though `.graphql` is the standard, I bump into a lot of cases where people use `.gql`. Perhaps it would be worthwhile to support both?

trevor-scheer: This is out of context for the PR, would you mind submitting an issue or PR for it? Thanks!

trevor-scheer:
Outdated

PR #656, opened 2018-11-05
filter objects in getter to only include known typenames

koenpunt:
This is a solution to the problem described here:

> We have generated a Swift API for a schema that includes an interface.
>
> When we now add an additional type that implements this interface to the server, the app crashes. This can of course be fixed by regenerating the api, but this doesn't help for apps already running on a user's device.
> https://github.com/apollographql/apollo-ios/issues/400

I'm not sure if this is the only place where this filtering has to be done, but this at least fixes it for our use case.

### DIFF
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -819,7 +819,9 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
   /// The friends of the character, or an empty list if they have none
   public var friends: [Friend?]? {
     get {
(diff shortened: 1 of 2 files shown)


koenpunt:
Ping ✨

koenpunt:
Anyone?

trevor-scheer:
@koenpunt apologies for the lack of response here. Unfortunately, this is code I don't know how to review and we don't currently have anyone available who does.

As a side note, we're hiring a mobile engineer!
https://www.apollographql.com/careers/positions/#Mobile-Engineer

koenpunt:
It took some time, but I'm glad it's merged now! 🚀

designatednerd:
D'oh! This is causing [some issues with our sample project](https://travis-ci.org/apollographql/apollo-ios/jobs/568034798), looking into a fix

PR #660, opened 2018-11-07
Fix unterminated string in readme

steverice:


steverice:
Duplicates https://github.com/apollographql/apollo-tooling/pull/659

Issue #669, opened 2018-11-08
apollo-vscode: Cannot use local schema file


stnwk:
Hi everyone!

I was just trying to figure out how to configure the new **VSCode Extension**, but I simply couldn't manage to do so.

That's my situation:
- installed apollo 2.0.5 - reloaded everything
- not using Apollo Engine
- cannot use an actual api endpoint (because of auth)
- local schema file available in JSON
- All my graphql files are under src/**

How do I successfully configure this in my `apollo.config.js`? I've tried everything but somehow it nevers works. 

```json
module.exports = {
  service: {
    localSchemaFile: './internals/schema.graphql',
  },
};
```
^ This config does not output anything in the debug view.

```json
module.exports = {
  client: {
    service: {
      localSchemaFile: './internals/schema.graphql',
    },
  }
};

```
^ This config outputs `TypeError: Only absolute URLs are supported` - welp

Additionally I receive the following error although I don't wan't to use the Apollo Engine related stuff..
`
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: Unable to find ENGINE_API_KEY
`

Can someone help me? The docs weren't really helpful about this :(

jessedvrs:
When I take a look at this source code here: https://github.com/apollographql/apollo-tooling/blob/b85fa652f373eafccdbd01d8653e4b54d3388333/packages/apollo-language-server/src/schema/providers/index.ts#L25

Looks like the `endpoint` configuration option is set by default. That may be causing your TypeError. Can you try to change your config to:

```js
module.exports = {
  client: {
    service: {
      localSchemaFile: './internals/schema.graphql',
      endpoint: null
    },
    includes: [
      "client/src/**/*.{ts,tsx,js,jsx,graphql}"
    ]
  }
};
```
Replace `client/src` with the path to your source code. This setting defaults to `src/**/*.{ts,tsx,js,jsx,graphql}`.

Also, apollo.config.js must be located in your project root (as seen by vscode).

jpaas:
Same problem here. I tried it with `endpoint: null` but same result

gund:
@stnwk this error still happens for me!

My config:
```js
module.exports = {
  client: {
    service: {
      name: 'MYGQL',
      localSchemaFile: './src/generated/graphql.schema.json',
      includes: ['src/**/*.{graphql,ts,js}'],
      excludes: ['**/node_modules/**'],
    },
  },
};
```

[...]

PR #676, opened 2018-11-08
[apollo-language-server] Allow config.client.localSchemaFile

stnwk:
Imagine the following `apollo.config.js`:

```js
module.exports = {
  client: {
    name: 'Some name',
    localSchemaFile: './src/data/schema.json',
  },
};
```

This config comes from the following situation:
- You cannot provide an api endpoint for some reasons (maybe auth, whatever)
- However, you have the necessary `schema.json` locally

Currently, this is not possible.
This PR fixes this.

Fixes https://github.com/apollographql/apollo-tooling/issues/669


stnwk:
No idea what's wrong with circleci though 🤔

stnwk:
@martijnwalraven Can you take a look at this and give me a hint whether this is moving in the right direction? Thanks :)

martijnwalraven:
@stnwk Yes, sorry, I had pushed a fix for the `getServiceName` issue before I saw this, so it needs to be rebased. (The CI issues should be fixed.)

Using `FileSchemaProvider` is exactly what that was meant for, but we'll need to decide on the config format. I suspect we'll want to nest `localSchemaFile` under `service` instead of directly under `client`, but I'd like to get @jbaxleyiii's take on that as well.

PR #682, opened 2018-11-11
Make sure writtenFiles is propagates

bobvanderlinden:
Currently whenever `client:codegen` is used the command outputs the following:

```
 ✔ Generating query files with 'typescript' target - wrote undefined files
```

`undefined` is because `writtenFiles` is `undefined`. `writtenFiles` is `undefined` because `write` doesn't return any value. `generate` returns the number of written files, so that should be returned in `write` as well.


jbaxleyiii:
@bobvanderlinden Thanks for this! 💯

PR #692, opened 2018-11-13
Small bug fix: Correctly incorporate column offset in decorations

patrys:
This fixes the case where error markers are placed incorrectly when an error refers to a range in the first line of inlined GraphQL.

Previously error locations (and thus editor decorations) would disregard the column offset. Error were reported like this:

```
foo = gql`query Foo { bar }`
            ^^^
```

This change makes the location add the column offset it referring to the first line:

```
foo = gql`query Foo { bar }`
                      ^^^
```


martijnwalraven:
Thanks a lot for tracking this down and fixing it, tests are especially appreciated!

PR #705, opened 2018-11-14
Don't swallow errors from Engine

trevor-scheer:
Throw errors we receive from Engine so we can provide the right feedback to users and handle the errors.

Tested for `client:check` and `client:push` using invalid operations against a published schema.

### DIFF
--- a/packages/apollo-language-server/src/engine/index.ts
+++ b/packages/apollo-language-server/src/engine/index.ts
@@ -131,7 +131,7 @@ export class ApolloEngineClient extends GraphQLDataSource {
       }
       // use error logger
       if (errors) {
-        throw new Error(errors.map(error => `${error.message}`).join("\n"));
+        throw new Error(errors.map(error => error.message).join("\n"));
       }
       if (!data) {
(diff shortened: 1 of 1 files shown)


packages/apollo-language-server/src/engine/index.ts
```diff
@@ -130,9 +130,9 @@ export class ApolloEngineClient extends GraphQLDataSource {
         );
       }
       // use error logger
-      // if (errors) {
-      //   throw new Error(errors);
-      // }
+      if (errors) {
+        throw new Error(errors.map(error => error.message).join("\n"));
```
abernix: Can we DRY this up a bit by creating a helper?  This seems like it's going to be a very common pattern that's going to perpetuate if we don't nip it in the bud right now.

At the very least, this might be a `mapGraphQLErrorsToString(errors)` (and be `errors.map(error => error.message).join("\n")`), but I'd even go so far to say we could have a `assertNoGraphQLErrors(errors);` function which would be all three lines here?

PR #707, opened 2018-11-15
Allow fragments only for codegen

### LINKED ISSUE #684: Fragments-only set of files are not supported
Hi,

When trying to generate types for a set of files that only contains fragments, `apollo client:codegen` fails with 
```
Error: No document sets found to generate code for.
```
I suppose that it's related to the following line in `apollo/lib/commands/client/codegen.js`:
```
if (!operations.length) {
  throw new Error("No document sets found to generate code for.");
}
```
As a workaround I can add any query and then codegen works

trevor-scheer:
Only throw an error if we have neither operations _or_ fragments.

Tested this on a demo project with fragments only to confirm the fix.

Fixes #684

### DIFF
--- a/packages/apollo/src/commands/client/codegen.ts
+++ b/packages/apollo/src/commands/client/codegen.ts
@@ -166,7 +166,7 @@ export default class Generate extends ClientCommand {
 
                 if (!operations.length && !fragments.length) {
                   throw new Error(
-                    "No document sets found to generate code for."
+                    "No operations or fragments found to generate code for."
                   );
                 }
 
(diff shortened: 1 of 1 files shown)


packages/apollo/src/commands/client/codegen.ts
```diff
@@ -164,7 +164,7 @@ export default class Generate extends ClientCommand {
                 const operations = Object.values(this.project.operations);
                 const fragments = Object.values(this.project.fragments);
 
-                if (!operations.length) {
+                if (!operations.length && !fragments.length) {
                   throw new Error(
                     "No document sets found to generate code for."
```
martijnwalraven: Maybe we could also change the error message, because 'document sets' are not a user concept and might be confusing.

trevor-scheer: Agreed. I'll go with `document sets` -> `operations or fragments`

Issue #680, opened 2018-11-09
`--no-addTypename` does not work


artemave:
```
$ yarn apollo client:codegen src --target=typescript --no-addTypename
$ /Users/artem/projects/inlens-web/node_modules/.bin/apollo client:codegen src --target=typescript --no-addTypename
 ›   Error: Unexpected argument: --no-addTypename
```

Looking at the code, the reason appears to be that `addTypename` flag declaration is missing `allowNo: true` option (or should not default to `true`):

![image](https://user-images.githubusercontent.com/23721/48290883-c78c4c80-e474-11e8-8782-6d60894507b4.png)

https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo/src/commands/client/codegen.ts#L47,L50

npm module version: 2.0.9

trevor-scheer:
Resolved in https://github.com/apollographql/apollo-tooling/pull/758

PR #713, opened 2018-11-16
addTypename flag should be false by default (#680)

milosdanilov:
By following documentation --addTypename argument should be set to false by default

### DIFF
--- a/packages/apollo/src/commands/client/codegen.ts
+++ b/packages/apollo/src/commands/client/codegen.ts
@@ -44,6 +44,10 @@ export default class Generate extends ClientCommand {
         "Type of code generator to use (swift | typescript | flow | scala), inferred from output",
       required: true
     }),
+    localSchemaFile: flags.string({
+      description:
+        "Path to your local GraphQL schema file (introspection result or SDL)"
+    }),
     addTypename: flags.boolean({
       description: "Automatically add __typename to your queries",
       default: true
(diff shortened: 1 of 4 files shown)


milosdanilov:
@zamiang Nope. you can not pass a value to addTypename flag. It is just checked for presence of that particular flag.

milosdanilov:
@martijnwalraven or @jbaxleyii please review this.

jbaxleyiii:
@milosdanilov I think that `addTypename` should be true since its the default for Apollo Client. With Oclif, we should be able to add a negation ability to this flag via [the flag config](https://oclif.io/docs/flags#docsNav) which we can document in the flag description.

From the oclif docs:
```
// flag with no value (-f, --force)
  force: flags.boolean({
    char: 'f',
    default: true,                // default value if flag not passed (can be a function that returns a boolean)
    // boolean flags may be reversed with `--no-` (in this case: `--no-force`).
    // The flag will be set to false if reversed. This functionality
    // is disabled by default, to enable it:
    // allowNo: true
  }),
```

packages/apollo/src/commands/client/codegen.ts
```diff
@@ -50,7 +50,7 @@ export default class Generate extends ClientCommand {
     }),
     addTypename: flags.boolean({
       description: "Automatically add __typename to your queries",
-      default: true
+      default: false
```
jbaxleyiii: Let's keep this as `true` but add `allowNo: true` and update the description to tell users they can use `--no-addTypename` to disable this feature

milosdanilov: Ok, I agree. Didn't know that `true` is default for Apollo Client. Will update the description also, because documentation was a little off, especially for this part.

Should i create new PR only for this change, and close this one?

PR #715, opened 2018-11-16
pass directives from schema modules through buildServiceDefinition

tgerk:
Directive definitions were discarded when merging types from modules in `buildServiceSchema`.  The fix is nearly trivial, because `Kind.DIRECTIVE_DEFINITION` is a unique case that is neither a schema definition nor a type definition.

### DIFF
--- a/packages/apollo-tools/src/buildServiceDefinition.ts
+++ b/packages/apollo-tools/src/buildServiceDefinition.ts
@@ -35,11 +35,15 @@ export function buildServiceDefinition(
   const errors: GraphQLError[] = [];
 
   const typeDefinitionsMap: {
-    [name: string]: (TypeDefinitionNode | DirectiveDefinitionNode)[];
+    [name: string]: TypeDefinitionNode[];
   } = Object.create(null);
 
   const typeExtensionsMap: {
     [name: string]: TypeExtensionNode[];
+  } = Object.create(null);
+
+  const directivesMap: {
+    [name: string]: DirectiveDefinitionNode[];
   } = Object.create(null);
 
   const schemaDefinitions: SchemaDefinitionNode[] = [];
@@ -50,10 +54,7 @@ export function buildServiceDefinition(
       module = { typeDefs: module };
     }
     for (const definition of module.typeDefs.definitions) {
-      if (
-        isTypeDefinitionNode(definition) ||
-        definition.kind === Kind.DIRECTIVE_DEFINITION
-      ) {
+      if (isTypeDefinitionNode(definition)) {
         const typeName = definition.name.value;
 
         if (typeDefinitionsMap[typeName]) {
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Thanks for looking into this! I don't think this is the right solution though, because directives live in their own namespace, and adding them to `typeDefinitionsMap` clashes with similarly named type definitions.

I think the solution is to keep a separate list of `directiveDefinitions`, similar to what `buildASTSchema` does (see [here](https://github.com/graphql/graphql-js/blob/64b194c6c9b9aaa1c139f1b7c3692a6ef851928e/src/utilities/buildASTSchema.js#L146-L148)).

Also, it seems this PR needs to be rebased, because it includes unrelated commits made by others.

tgerk:
Pardon the confusion, I have rebased my branch from the tagged release to master branch.

I've also segregated the directives from types as you pointed out.

packages/apollo-tools/src/buildServiceDefinition.ts
```diff
@@ -174,6 +183,15 @@ export function buildServiceDefinition(
       });
     }
 
+    const directiveExtensions = Object.values(directivesMap).flat();
+
+    if (directiveExtensions.length > 0) {
```
trevor-scheer: I don't think we're necessarily _extending_ the schema with directives - this might make more sense to include where we first build the schema up above:

```javascript
let schema = buildASTSchema({
  kind: Kind.DOCUMENT,
  definitions: [...typeDefinitions, ...directiveDefinitions]
});
```

I'd like to hear @martijnwalraven 's opinion on this though as I may be overlooking something.

martijnwalraven: @trevor-scheer Yep, this definitely makes sense, and it's even required if you're defining schema directives because you need those definitions when first constructing the schema. Regardless, `extendSchema` has to re-create the schema so it might be better to avoid that extra pass.

packages/apollo-tools/src/buildServiceDefinition.ts
```diff
@@ -65,6 +70,14 @@ export function buildServiceDefinition(
         } else {
           typeExtensionsMap[typeName] = [definition];
         }
+      } else if (definition.kind === Kind.DIRECTIVE_DEFINITION) {
+        const typeName = definition.name.value;
```
martijnwalraven: This isn't really a `typeName` but a `directiveName`. If we're using a map, we should probably also enforce a directive to only be defined once and report an error otherwise. That would actually make sense I think, but [`buildASTSchema`](https://github.com/graphql/graphql-js/blob/64b194c6c9b9aaa1c139f1b7c3692a6ef851928e/src/utilities/buildASTSchema.js#L146-L148) doesn't do that either (it just keeps an array of directives) and that results in 'last wins'.

tgerk:
```suggestion
        const directive = definition.name.value;
```

packages/apollo-tools/src/buildServiceDefinition.ts
```diff
@@ -65,6 +70,14 @@ export function buildServiceDefinition(
...
+
+        if (directivesMap[typeName]) {
+          directivesMap[typeName].push(definition);
```
tgerk:
```suggestion
          directivesMap[directiveName].push(definition);
```

tgerk:
Changes have been made to variable names and to behavior as suggested, including tests.
+ duplicated directive declarations within a module or in different modules are detected and reported
+ directives and type definitions are used together in constructing the resulting schema

martijnwalraven:
Thanks for opening this PR and making the changes! (The CI failures are unrelated I think.)

PR #727, opened 2018-11-19
Fix "go to definition" when a local schema file is provided

### LINKED ISSUE #726: [VSCode] "Go to Definition" is broken when using a localSchemaFile
When I configure my `apollo.config.js` with a `localSchemaFile` and select "go to definition" on a query field, I get the following error:
image

I would expect it to navigate to the definition of the field in the local schema file.

My config:
```js
module.exports = {
  client: {
    service: {
      name: 'local',
      localSchemaFile: './app/graphql/schema.graphql',
    },
    includes: ['frontend/javascripts/**/*.{ts,tsx,js,jsx,graphql}'],
  },
};
```

rylanc:
Fixes #726


jbaxleyiii:
@rylanc awesome, thanks for doing this!

I think long term we may want to actually link to the local schema file, but this would be tricky given our support of `.graphql` and  `.json` to load the schema

PR #732, opened 2018-11-21
Add @types/listr to apollo package devDeps

evans:
I ran into issues with referencing the apollo package from a typescript
file, since the listr types are not present in the package.json on
publish. Unfortunately lerna doesn't ensure that all used dependencies
show up in the package.json on publish. Maybe there is a flag we can use
to ensure that we don't run into these sort of annoying issues in the
future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
   },
   "dependencies": {
     "@apollographql/apollo-tools": "file:packages/apollo-tools",
+    "@types/listr": "^0.13.0",
     "apollo": "file:packages/apollo",
     "apollo-codegen-core": "file:packages/apollo-codegen-core",
     "apollo-codegen-flow": "file:packages/apollo-codegen-flow",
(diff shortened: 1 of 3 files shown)


evans:
@abernix That's totally true! Thank you for looking into it. Seems like https://docs.npmjs.com/cli/install#description only applies to dependencies a single level down, specifically:

> With the --production flag (or when the NODE_ENV environment variable is set to production), npm will not install modules listed in devDependencies.
>
> NOTE: The --production flag has no particular meaning when adding a dependency to a project.

I'll make the change!

trevor-scheer:
@evans terribly sorry this has gone ignored for so long 😞is this still something we're interested in merging? Let's catch up about this tomorrow if so, I'm curious about the details!

zionts:
@evans this PR seems to have bitrotten. Are you aware of whether or not this is still a problem? If so, should we continue on this branch or open up an issue to track it?

evans:
Looks like the dependency still exists in the root `package.json` and not inside of the package.json for `apollo`, so this is still a barrier to usage as a required module.

My current understanding is that `apollo` is a cli tool first and so the usage of it as a required module isn't pressing. I think that adding `@types/listr` will increase download time(albeit by a very small percentage), which is in conflict with making a cli tool that is efficiently downloaded on every ci/cd run.

I believe a work around is to install `@types/listr` driectly in the consuming project, so I'm comfortable closing this pull request.

Thank you for the bump @zionts!

Issue #728, opened 2018-11-20
Self signed certs


maarekj:
The `--insecure` option has disappeared again.
see #199

PR #735, opened 2018-11-21
Restore skipSSLValidation flag

trevor-scheer:
This restores the skipSSLValidation flag so a schema introspection can be downloaded via http.

For reference: https://github.com/apollographql/apollo-tooling/pull/565
I ran into errors trying to use this^ implementation with `https`. Using `http.Agent` worked for me, though based on the config from the previous implementation I'm curious if I'm overlooking something.
>TypeError [ERR_INVALID_PROTOCOL]: Protocol "http:" not supported. Expected "https:"

Fixes: #728

### DIFF
--- a/packages/apollo-language-server/src/schema/providers/introspection.ts
+++ b/packages/apollo-language-server/src/schema/providers/introspection.ts
@@ -10,10 +10,8 @@ import {
   IntrospectionQuery,
   parse
 } from "graphql";
-import { Agent, AgentOptions } from "https";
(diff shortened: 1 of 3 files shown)


raudabaugh:
I am seeing the same error as @codename- 
cc @trevor-scheer @jbaxleyiii

sebastianseilund:
Same issue here. [See workaround here](https://github.com/apollographql/apollo-tooling/issues/928#issuecomment-467609141)

PR #740, opened 2018-11-26
Associate `*.gql` files with graphql language in GraphQLProject

brabeji:
Codegen wasn't working for me with my files ending in ".gql". As my IDE plugin recognizes "*.gql" files as graphql, this is very confusing.

I see two possible further improvements:
1. Warn when including a file without language association
2. Fallback to `graphql` language


jbaxleyiii:
@brabeji ah great! Thanks for adding this!

Issue #737, opened 2018-11-25
fix : Uncomment test suite


andykenward:
The apollo commands service tests have been commented out for the summit.

Can this https://github.com/apollographql/apollo-tooling/commit/04452e584e572906b36cfb6654def6fa4a9b8de2 be reverted / fixed?

As versions have been released without these tests running.

jbaxleyiii:
@andykenward we are working to get these test suites back working asap! @JakeDawkins is rewriting the file mocking system that was a hack within codegen (and not the new language server) to make it work. Should be back on line this week!

JakeDawkins:
This is an ongoing project that gets worked on as features are worked on, so I'm going to close this issue, but it's not abandoned :)

PR #743, opened 2018-11-27
CLI Testing

JakeDawkins:


packages/apollo/package-lock.json
```diff
@@ -576,6 +576,12 @@
         "brace-expansion": "^1.1.7"
       }
     },
+    "mock-fs": {
```
jbaxleyiii: are we using mock-fs anymore?

packages/apollo/src/commands/client/__tests__/generate-new.test.ts
```diff
@@ -0,0 +1,268 @@
+// jest.mock("apollo-codegen-core/lib/localfs", () => {
```
jbaxleyiii: lets remove this comment and also rename this file to what we want it to be in the final version.

jbaxleyiii:
fixes https://github.com/apollographql/apollo-tooling/issues/737

packages/apollo/package-lock.json
JakeDawkins: Nope!

PR #758, opened 2018-11-28
Allow --no-addTypename in 

milosdanilov:
Should be able to set --addTypename to false with --no-addTypename  in apollo client:codegen command.

Follow up for PR #713 
By agreement of @jbaxleyiii

This should resolve Issue #680


jbaxleyiii:
@trevor-scheer I think you are correct! @milosdanilov thanks for doing this!

jbaxleyiii:
@JakeDawkins @trevor-scheer I think we should probably move this up to the `ClientCommand` at some point so the same option is available for client:check / publish, but our current stance of config first for those is the safer route for now.

PR #769, opened 2018-11-30
[vscode] Use `tagName` from config to extract documents.

alloy:
Now other tags (such as Relay’s `graphql` tag) will actually be recognized with a `apollo.config.js` file like the following:

```js
module.exports = {
  client: {
    service: {
      name: "local",
      localSchemaFile: "./data/schema.graphql",
    },
    includes: ["src/**/*.{ts,tsx,graphql}"],
    excludes: ["**/node_modules", "**/__tests__"],
    tagName: "graphql",
  }
}
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,14 @@
       "stopOnEntry": false,
       "sourceMaps": true,
       "outFiles": ["${workspaceRoot}/packages/vscode-apollo/lib/**/*.js"]
-    }
+    },
+    {
+      "name": "Attach to TS Server",
+      "type": "node",
+      "request": "attach",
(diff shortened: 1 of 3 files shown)


jbaxleyiii:
@alloy 👋 thanks for doing this! Does this mean the language server / extension will support Relay projects now? If so that would be really really awesome!

alloy:
@jbaxleyiii Yup, it does, that’s what we use 👍 Navigating with the extension is feeling pretty sweet thus far!

Btw, can you offer any insight into how long it would take for this to make it into a published release?

jbaxleyiii:
@alloy that's *awesome*! Release should be out today!

PR #787, opened 2018-12-02
Add a note about client schemas

### LINKED ISSUE #786: Possible regression: no longer able to extend with client schema
As first noted in #731, the `--clientSchema` CLI option has been removed. However, there is no documentation on how to reimplement the desired behavior, especially when using `--localSchemaFile`.

I see that @bsunderhus got it working in his setup somehow, but looking at the source I'm not sure how. Has this ability been removed?

I've resorted to just applying the client query myself, with this old script I had laying around, in case anyone else finds this necessary:

```js
#!/usr/bin/env node

const fs = require("fs");
const path = require("path");

const { introspectionFromSchema } = require("graphql/utilities");
const { makeExecutableSchema } = require("graphql-tools");
const { fileLoader, mergeTypes } = require("merge-graphql-schemas");

// Make sure unhandled errors in async code are propagated correctly
process.on("uncaughtException", error => {
  console.error(error);
  process.exit(1);
});

process.on("unhandledRejection", error => {
  throw error;
});

async function introspectSchema(input, output) {
  const schemas = [].concat(...input.map(i => fileLoader(i)));
  const typeDefs = mergeTypes(schemas, {
    all: true
  });

  const schema = await makeExecutableSchema({
    typeDefs,
    resolverValidationOptions: { requireResolversForResolveType: false }
  });
  const introspection = await introspectionFromSchema(schema);
  const json = JSON.stringify(introspection, null, 2);
  fs.writeFileSync(output, json);
}

const input = [
  path.resolve(__dirname, "../../../schema/*.graphql"),
  path.resolve(__dirname, "../src/schema/*.graphql")
];

const output = path.resolve(
  __dirname,
  "../src/schema/__generated__/schema.json"
);

// Generate an introspection JSON format from remote GraphQL server merging
// with any local GraphQL schemas
introspectSchema(input, output, true);
```

mike-marcacci:
I updated the CLI docs and apollo readme to note that they are now
specified in the --includes glob.

Fixes #786.

### DIFF
--- a/packages/apollo/README.md
+++ b/packages/apollo/README.md
@@ -71,8 +71,7 @@ OPTIONS
 
   --header=header                        Additional headers to send to server for introspectionQuery
 
-  --includes=includes                    Glob of files to search for GraphQL operations. This should be used to find queries
-                                         *and* any client schema extensions
+  --includes=includes                    Glob of files to search for GraphQL operations
 
   --key=key                              The API key for the Apollo Engine service
 
@@ -126,8 +125,7 @@ OPTIONS
 
   --header=header                            Additional headers to send to server for introspectionQuery
 
-  --includes=includes                        Glob of files to search for GraphQL operations. This should be used to find
-                                             queries *and* any client schema extensions
+  --includes=includes                        Glob of files to search for GraphQL operations
 
   --key=key                                  The API key for the Apollo Engine service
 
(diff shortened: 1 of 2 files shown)


trevor-scheer:
Hey Mike, thanks for adding this. The README is updated automatically when we publish the package, so if you don't mind reverting the changes to that file, this should be all set.

Looks like this might need a `prettier` run as well, though I think that should be in a precommit hook?

mike-marcacci:
Hi @trevor-scheer! I must have missed your comment. Fixed and fixed; prettier wasn't run because I never actually ran `npm install` (since it was a docs-only change) and so git hooks weren't set up.

trevor-scheer:
Thanks @mike-marcacci! 🏆

PR #794, opened 2018-12-04
[RFC] Add CLI parameter to pass additional Swift framework imports

arifken:
**What:**
This PR exposes an `importFramework` parameter to the `client:codegen` command that allows developers to add framework imports to the code generated swift files. 

**Use Case:**
Currently, the only import that is added to generated Swift files is for Apollo, but we found that it would be useful to be able to import other frameworks in cases where custom scalars are defined in modules outside of where the code generated Swift file is being added

**Is there a workaround?**
The only workaround currently is to edit the code generated swift files after they are generated and add additional imports

Note: This PR is built against the `apollo@1.8.3` tag, since there seems to be an issue currently with running swift code generation in master

### DIFF
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -47,12 +47,17 @@ export interface Options {
   customScalarsPrefix?: string;
 }
 
+export interface SwiftGeneratorOptions {
+  importFrameworks?: string[];
+}
+
 export function generateSource(
   context: CompilerContext,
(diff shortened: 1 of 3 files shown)


zionts:
@arifken since it's been more than a year 😨 since this was open, I'm going to close this out, but if this is still an issue, please let me know! We would love to put some time into understanding what it takes to enable this functionality and discuss it over an issue or a PR. Please feel free to reopen if you're interested in picking this back up, and feel free to `@` to make sure it gets some attention

PR #800, opened 2018-12-04
[client config] Allow exclusion of validation rules.

alloy:
We’re enjoying the extension on our Relay project very much!

The only thing that’s not working well right now is in operation variables validation. Relay Modern has [the `@argumentDefinition` directive](https://facebook.github.io/relay/docs/en/fragment-container.html#argumentdefinitions), which defines ‘local’ variables. These don’t need to be declared at the root of the operation, which currently leads to the many false positives seen below.

This change makes it possible to define rules to exclude in your project config. Ours now looks like this:

```js
module.exports = {
  client: {
    service: {
      name: "local",
      localSchemaFile: "data/schema.graphql",
    },
    excludeValidationRules: [
      // Defaults of vscode-apollo. Not sure why, maybe we can play with this.
      "NoUnusedFragments",
      "KnownDirectives",
      // Added because Relay has ‘local’ variables (defined with the
      // `@argumentDefinitions` directive) and these don’t need to be defined on
      // the operation.
      "NoUndefinedVariables",
    ],
    includes: ["src/**/*.{ts,tsx,graphql}"],
    excludes: ["**/node_modules", "**/__tests__"],
    tagName: "graphql",
  },
}
```

screen shot 2018-12-01 at 00 29 47

### DIFF
--- a/packages/apollo-language-server/src/errors/validation.ts
+++ b/packages/apollo-language-server/src/errors/validation.ts
@@ -20,17 +20,13 @@ import { ToolError, logError } from "./logger";
 export function getValidationErrors(
   schema: GraphQLSchema,
   document: DocumentNode,
-  fragments?: { [fragmentName: string]: FragmentDefinitionNode }
+  fragments?: { [fragmentName: string]: FragmentDefinitionNode },
+  excludeRules: string[] = ["NoUnusedFragments", "KnownDirectives"]
 ) {
-  const specifiedRulesToBeRemoved = [
-    NoUnusedFragmentsRule,
-    KnownDirectivesRule
-  ];
-
   const rules = [
     NoAnonymousQueries,
     NoTypenameAlias,
-    ...specifiedRules.filter(rule => !specifiedRulesToBeRemoved.includes(rule))
+    ...specifiedRules.filter(rule => !excludeRules.includes(rule.name))
   ];
 
   const typeInfo = new TypeInfo(schema);
--- a/packages/apollo-language-server/src/config.ts
+++ b/packages/apollo-language-server/src/config.ts
@@ -76,6 +76,15 @@ export interface ClientConfigFormat extends ConfigBase {
   tagName?: string;
   // stats window config
   statsWindow?: StatsWindowSize;
+
+  /**
+   * Rules that should be excluded from the validation rules that are in the GraphQL specification.
+   *
+   * By default the `NoUnusedFragments` and `KnownDirectives` rules are excluded.
+   *
+   * @see https://github.com/graphql/graphql-js/blob/master/src/validation/specifiedRules.js
+   */
+  excludeValidationRules?: string[];
 }
 
 export const DefaultClientConfig = {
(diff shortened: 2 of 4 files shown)


packages/apollo-language-server/src/errors/validation.ts
```diff
@@ -20,17 +20,13 @@ import { ToolError, logError } from "./logger";
 export function getValidationErrors(
   schema: GraphQLSchema,
   document: DocumentNode,
-  fragments?: { [fragmentName: string]: FragmentDefinitionNode }
+  fragments?: { [fragmentName: string]: FragmentDefinitionNode },
+  excludeRules: string[] = ["NoUnusedFragments", "KnownDirectives"]
 ) {
-  const specifiedRulesToBeRemoved = [
-    NoUnusedFragmentsRule,
```
alloy: @martijnwalraven Do you know offhand why this one was needed? 1c22f0680c8f8be9c89782c402fd7413fe8d0e91

jbaxleyiii:
@alloy Thanks for opening this PR! I've been thinking about this over the weekend as well.

I'm wondering if instead of `excludeValidationRules` we could instead allow users to provide their own actual validation rules? One of the reasons why we made the config format javascript was so that teams could provide much more control over their environment. We were inspired very much by the power and flexibility of danger here :)

One thought was to export `defaultValidationRules` from `apollo-language-server` (or `apollo`) and allow users to set `validationRules: ValidationRule[]` so that you could not only remove defaults, but also write and configure you own rules! This could allow you to still validate those arguments, enforce certain fields (like the `id` field if present) or other options.

@martijnwalraven curious your thoughts on this as well

alloy:
I’m always in favour of allowing more control over one’s tools!

As a matter of fact, I was at first inclined to implement this exclusion configuration using function instances, as the existing code did. However, I quickly realised that this would be tough in unintuitive ways, because the language server loads `graphql-js` code from a copy in its own `node_modules` directory. The same problem, in the opposite direction, may also exist with code loaded in the user’s config file that’s dependent on the user’s project.

All of this can probably be fixed, but that’s a way larger scope and in the meantime the extension is overloading my coworkers with false positives, hence the simple iteration on existing code.

alloy:
Bump

trevor-scheer:
Hey @alloy, we should be taking a closer look at this shortly after the new year - the issue is high on our backlog 👍

alloy:
Thanks and have great holiday! 👌

orta:
Poke on this one 👍  ( we removed our fork of this extension from the store, so this is in a limbo from our side too )

orta:
Poke again folks ❤️

JakeDawkins:
@alloy @orta Okay, soo I've thought about this and discussed it with some others, and I think I have a solution.

For reasons described in #1056, I think we should add a complete override for validations, not just the ability to exclude existing ones. 

The work in this PR is a great guide, but not quite what we'd be looking for, so I'll close it, and gladly look at future PRs for this!

Feel free to also discuss it/ask questions/throw things at me in that issue 😄

PR #809, opened 2018-12-06
Ability to download schema from engine within service projects

JakeDawkins:
## Issue

- You couldn't use `apollo service:download` from within a `service` (server) project.
- The schema provider looks up schemas differently for client and service projects, since their configs are different shapes.
  - originally, the schema provider only supported getting a schema from a local schema files or an endpoint. 
  - the default endpoint that would get used if neither a schema file/endpoint option was present was `http://locahost:4000` which caused some users to see connection errors to localhost:4000 while trying to run introspection.

## Fixes

- Removed default endpoint for service projects
- Added ability for the `EngineSchemaProvider` to accept a service config file, and use the `service.name` to fetch from Engine (same as `client.service`)
- ~Made the schema lookup by `service.name` the one with highest precedence, since using a `--key` should always overwrite local config files with endpoints.~

### DIFF
--- a/packages/apollo-language-server/src/schema/providers/engine.ts
+++ b/packages/apollo-language-server/src/schema/providers/engine.ts
@@ -1,8 +1,6 @@
 // EngineSchemaProvider (engine schema reg => schema)
 import { NotificationHandler } from "vscode-languageserver";
-
 import gql from "graphql-tag";
-
 import {
   GraphQLSchema,
   buildClientSchema,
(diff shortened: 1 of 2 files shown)


packages/apollo-language-server/src/schema/providers/index.ts
```diff
@@ -24,6 +24,10 @@ export function schemaProviderFromConfig(
   config: ApolloConfig
 ): GraphQLSchemaProvider {
   if (isServiceConfig(config)) {
+    if (config.service.name) {
```
jbaxleyiii: I think we should hit this branch last since you may have a name and things like a local schema file. The name is arbitrary right now (in that we want to use it for debugging, client name, etc) and we need to clean up the "engine linking" side of things.

packages/apollo-language-server/src/schema/providers/index.ts
```diff
@@ -24,6 +24,10 @@ export function schemaProviderFromConfig(
   config: ApolloConfig
 ): GraphQLSchemaProvider {
```
jbaxleyiii: We should also have a nice error if this function can't find the needed information to find a service

packages/apollo-language-server/src/schema/providers/engine.ts
```diff
@@ -15,29 +21,49 @@ import {
...
-    const [id, tag = "current"] = parseServiceSpecificer(client.service);
-    const { data, errors } = await this.client.execute({
+    // we know that client.service is a string, since it's checked above
+    const serviceName = client
+      ? (client.service as string)
+      : service
+      ? service.name
+      : null;
+    // TODO more clear error
+    if (!serviceName) throw new Error(`Unable to find service name for Engine`);
```
jbaxleyiii: Let's make this something like:

`Unable to find a service name that would link to a service in Engine. This may be because we couldn't find an ENGINE_API_KEY in the environment, no `--key` was passed in, or the service name wasn't set in the apollo.config.js. For more information about configuring Apollo projects, see the guide here (url to apollo config page in docs shortened to a bitly style link)`

packages/apollo-language-server/src/schema/providers/engine.ts
```diff
@@ -15,29 +21,49 @@ import {
...
+      this.engineClient = new ApolloEngineClient(
+        engine.apiKey,
+        engine.endpoint
+      );
     }
 
-    const [id, tag = "current"] = parseServiceSpecificer(client.service);
-    const { data, errors } = await this.client.execute({
+    // we know that client.service is a string, since it's checked above
+    const serviceName = client
```
trevor-scheer: Can we split the logic for this nested ternary up into the `if / else if` above?
`isClientConfig / isServiceConfig`

packages/apollo-language-server/src/config.ts
```diff
@@ -93,13 +93,6 @@ export interface ServiceConfigFormat extends ConfigBase {
   localSchemaFile?: string;
 }
 
-export const DefaultServiceConfig = {
```
trevor-scheer: Totally agree with this change, but I'm concerned about people currently depending on this behavior.

Is it possible to either:
1) deprecate this with messaging and not remove it just yet? or...
2) add some error messaging with instructions for people who currently _do_ depend on this behavior so they have an easy time resolving the issue on upgrading.

JakeDawkins: Do you think that's necessary? The behavior isn't documented in any way currently, and IMHO shouldn't be a thing to begin with.

trevor-scheer: That's fair, I'm ok with leaving this out 👍

packages/apollo-language-server/src/schema/providers/engine.ts
```diff
@@ -15,29 +21,54 @@ import {
...
-    const { engine, client } = this.config;
+    const { engine, client, service } = this.config;
 
-    if (typeof client.service !== "string") {
-      throw new Error(
-        `Service name not found for client, found ${client.service}`
-      );
+    let serviceName;
+
+    if (isClientConfig(this.config)) {
```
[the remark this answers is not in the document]
JakeDawkins: This one is a bit tricky. The isClientConfig check doesn't explicitly check for `config.client`. `client` is a required field for `ClientConfig`, but not `ServiceConfig` and since `this.config` is a union between the two, theoretically, client _could_ be null

packages/apollo-language-server/src/schema/providers/index.ts
```diff
@@ -13,6 +13,7 @@ import {
 import { IntrospectionSchemaProvider } from "./introspection";
 import { EngineSchemaProvider } from "./engine";
 import { FileSchemaProvider } from "./file";
+import { createDecipher } from "crypto";
```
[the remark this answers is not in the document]
JakeDawkins: ugh, yep 😞

packages/apollo-language-server/src/schema/providers/engine.ts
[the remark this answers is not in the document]
trevor-scheer: Interesting, I see what you mean. I wonder if the `isClient` and `isService` fields are extraneous. Should be just checking for presence of client or service field I think.

Out of scope though 😛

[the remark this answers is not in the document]
martijnwalraven: I think `isClientConfig` and `isServiceConfig` guarantee `client` or `service` are non-null, because they acts as type guards (`config is ClientConfig` or `config is ServiceConfig` respectively). So if you move the destructuring within the conditionals, you shouldn't need `!s`.

PR #826, opened 2018-12-12
Setup CI with azure pipelines

JakeDawkins:
* Setup CI with azure pipelines to test windows w/ node 8 & 10
* Resolve Windows FS Issues
* Convert "string" URIs to actual Uri objects
* Don't manually build URIs, hand off to Uri constructor instead
* QoL debug script and snapshot update
* Updates snapshots


packages/vscode-apollo/create-server-symlink.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+const symlinkDir = require("symlink-dir");
+const path = require("path");
+
+symlinkDir("../../apollo-language-server", "./node_modules/")
```
trevor-scheer: Is it ok that this is an async function? Seems this could be fragile if not synchronous.

packages/vscode-apollo/create-server-symlink.js
```diff
@@ -0,0 +1,9 @@
...
+  .then(result => {
+    console.log(result);
+  })
+  .catch(err => console.error(err));
```
trevor-scheer: Should this throw on errors? I think if something goes wrong it may not make sense to continue.

packages/apollo-env/clone-types.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
trevor-scheer: Is this file not needed anymore?

azure-pipelines.yml
```diff
@@ -0,0 +1,77 @@
...
+      vmImage: "ubuntu 16.04"
+
+    steps:
+      - task: NodeTool@0
+        inputs:
+          versionSpec: "10.x"
+        displayName: "Install Node.js"
+
+      - script: |
+          npm ci
```
trevor-scheer: Should we make an npm script similar to what we do for `npm run circle`?

`"azure": "npm ci && npm run build && npm test"`

This would clean up some duplication and give devs an easy way to run exactly the commands the pipeline does without having to look anywhere besides package.json

packages/apollo-env/clone-types.js
```diff
@@ -0,0 +1,62 @@
+/**
```
trevor-scheer: I'm not sure where this belongs, but I don't think `apollo-env` is quite right. We might want a top-level `scripts` folder for these cross-platform scripts that we'll need to leverage.

Is typescript an option here?

packages/apollo-env/clone-types.js
```diff
@@ -0,0 +1,62 @@
...
+    // if it's in a nested dir, we need to create the full path to the file
+    // before creating the file
+    makeNestedDir(path.dirname(finalPath));
+
+    fs.copyFileSync(file, finalPath);
+    console.log(`copied: ${file} -> ${finalPath}`);
+  }
+};
+
+const files = walk("./src", /.*\.d\.ts/);
```
trevor-scheer: Is this the equivalent of `glob.sync("src/**/*.d.ts")`?

Nothing wrong here AFAICT, but this^ may be easier to grok as a dev just out of familiarity and we don't have to rewrite the thing.

This is just a nitpick, and if it would require a rewrite of your implementation around it then maybe not worth it.

JakeDawkins:
@kaylangan Also, is it possible to set specify a version of `npm` without having to run a script to install manually? It looks like the version installed with the `NodeTool` on linux is 6.x, but on windows it's 5.x which doesn't support `npm ci`

packages/vscode-apollo/create-server-symlink.js
JakeDawkins: I'm not 100% sure. I haven't seen it error _yet_ because of it though 😬

packages/apollo-env/clone-types.js
JakeDawkins: I'm not sure it'd fit outside anywhere though. It's something specifically used for env and in no other packages. I'd feel better hoisting it up a level if it was used anywhere else

kaylangan:
> @kaylangan Also, is it possible to set specify a version of `npm` without having to run a script to install manually? It looks like the version installed with the `NodeTool` on linux is 6.x, but on windows it's 5.x which doesn't support `npm ci`

Thanks for bringing this to our attention! It appears that specifying version 8.x for the NodeTool task actually installs different versions of Node depending on the OS. For Windows it seems to be installing 8.10.0 and for Linux it's installing 8.14.1. NPM 5.7 is the min version that supports "npm ci" which didn't ship until after 8.12 (?). I need to follow up with my team to understand why the task acts differently depending on the OS. Until then, I'd suggest specifying version "8.14" or ">8.14" which will eliminate needing the extra install step.

JakeDawkins:
@kaylangan Thanks!!

packages/vscode-apollo/create-server-symlink.js
abernix: @trevor-scheer Your concern here is very reasonable.  If any other instruction were after this, and not within the `then`-able, this would need to be `await`ed.  This works because:

1. There is nothing after the call to `symlinkDir`.
2. This is a stand-alone module invoked directly by `node` (That is to say, it's not traversed _via_ another file/module but directly by the interpreter).
3. The Node.js interpreter (`node`) evaluates this set of instructions (which populates the the event loop) and then proceeds to process the event loop until it's empty before exiting (`exit()`).

packages/vscode-apollo/create-server-symlink.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
```
trevor-scheer: Can we get rid of use strict?

packages/vscode-apollo/create-server-symlink.js
```diff
@@ -0,0 +1,11 @@
...
+const symlinkDir = require("symlink-dir");
+const path = require("path");
```
trevor-scheer: This is unused

packages/vscode-apollo/create-server-symlink.js
```diff
@@ -0,0 +1,11 @@
...
+
+symlinkDir("../../apollo-language-server", "./node_modules/")
+  .then(result => {
+    console.log(result);
+  })
+  .catch(err => {
```
trevor-scheer: We should be able to remove this catch if we're only rethrowing the error and not doing anything with it

packages/apollo/src/git.ts
```diff
@@ -48,7 +48,8 @@ export const gitInfo = async (): Promise<GitContext | undefined> => {
 
   let committer;
   let branch = ciBranch || prBranch;
-  let remoteUrl = slug;
+  // BUILD_REPOSITORY_ID is for azure pipelines
```
ngbrown: This doesn't seem to be right.

When using `apollo service:push` from an Azure repo on my local machine, the git path is correctly shown in the Apollo Engine schema history interface.  When using `apollo service:push` from an Azure pipeline, the following error is shown in the Apollo Engine schema history, with a yellow triangle:

```
The git context submitted to Engine has an invalid remoteUrl of "befddf0a-9669-46cd-af07-aba537276ebe"
```

The [Azure pipeline variables docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables) also don't mention a `Build.Repository.Id`, but do mention `Build.Repository.Uri`, which would be useful for a remoteUrl.

JakeDawkins: @ngbrown the format for remoteUrl seems to be different, depending on where in that file you're looking. For example, where you highlighted, it's something in the format of (`apollographql/apollo-tooling`). Later on in the file though, it's redefined with the actual remote url. That seems a bit odd. I'll look into  it and see if using the URL is correct. I wouldn't want to break anything that works currently

JakeDawkins: I opened an issue #1064 to track this until I find the answer, since it's a bit easier than doing it here :)

PR #829, opened 2018-12-13
VSCode Testing Suite

trevor-scheer:
This PR sets up the testing framework for the VSCode extension 🎉

There's a lot of hacky business in making this work that I'm not excited about. By default, VSCode is prescriptive about Mocha. We chose to explore Jest since that's what we use across the repo, and I found this which was very helpful:
https://github.com/Unibeautify/vscode/tree/master/test

I also found running VSCode's bin/test to be difficult from a standard npm script. The result of that is `vscode-test-runner.ts`.

Very open to feedback on this, as there are a few things I'm not all that excited about myself 👨‍🎨


packages/vscode-apollo/src/statusBar.ts
```diff
@@ -1,36 +1,37 @@
...
-  commands
-} from "vscode";
-import { LanguageClient } from "vscode-languageclient";
+import { window, StatusBarAlignment } from "vscode";
 
 export default class ApolloStatusBar {
-  private statusBarItem: StatusBarItem = window.createStatusBarItem(
-    StatusBarAlignment.Right
-  );
+  private _statusBarItem = window.createStatusBarItem(StatusBarAlignment.Right);
```
JakeDawkins: Why did you go with this convention here? Just for consistency across the project or what? Do you think it's necessary given we're already declaring it private with TS?

packages/vscode-apollo/src/testRunner/jest-config.ts
```diff
@@ -0,0 +1,20 @@
+import { resolve } from "path";
```
JakeDawkins: Is this file something we could hoist to the root, to make it more obvious where to find it?

packages/vscode-apollo/src/statusBar.ts
trevor-scheer: The convention is useful when you're using private variables with a getter (See `get statusBarItem` below). The value of keeping this private is debatable, but it does protect the internal instance of the `statusBarItem` from being set externally.

packages/vscode-apollo/src/testRunner/jest-config.ts
trevor-scheer: I'm tempted to leave it here next to its related files, since it resolves the environment and setup framework scripts which make up the whole testing suite. I'm open to other viewpoints - but this isn't meant to be consumed in the way that running jest normally would since it's imported and passed into the `jest.runCLI` function.

packages/vscode-apollo/src/statusBar.ts
```diff
@@ -1,36 +1,37 @@
...
 
-  constructor(context: ExtensionContext, client: LanguageClient) {
-    this.statusBarItem.text = "Apollo GraphQL $(rss)";
-    this.statusBarItem.show();
+  static loadingStateText = "Apollo GraphQL $(rss)";
+  static loadedStateText = "ApolloGraphQL $(rocket)";
+
+  constructor() {
+    this._statusBarItem.text = ApolloStatusBar.loadingStateText;
```
jbaxleyiii: I find it a little odd to refer to the `ApolloStatusBar` here. I'd say either make them static constants like `LOADING_STATE_TEXT` at the top of the file, or reference the class statics as `this.loadintStateText`

packages/vscode-apollo/package.json
```diff
@@ -113,5 +115,8 @@
         "category": "Apollo"
       }
     ]
+  },
+  "devDependencies": {
+    "jest-environment-node": "^23.4.0"
```
jbaxleyiii: I don't totally understand why this is needed here?

packages/vscode-apollo/src/statusBar.ts
```diff
@@ -1,36 +1,37 @@
...
+    this._statusBarItem.show();
 
     // this.statusBarItem.command = "apollographql/showOutputChannel";
-    // context.subscriptions.push(this.statusBarItem);
   }
 
-  public showLoadedState() {
-    if (!window.activeTextEditor) {
-      this.statusBarItem.hide();
+  get statusBarItem() {
```
jbaxleyiii: If we have a getter like this that just returns the internal private value (and does no work on it), I'd vote for just making it public

packages/vscode-apollo/src/statusBar.ts
```diff
@@ -1,36 +1,37 @@
...
+    hasActiveTextEditor: boolean;
+  }) {
+    if (!hasActiveTextEditor) {
+      this._statusBarItem.hide();
       return;
     }
 
-    this.statusBarItem.text = "Apollo GraphQL $(rocket)";
-    this.statusBarItem.show();
+    this._statusBarItem.text = ApolloStatusBar.loadedStateText;
```
jbaxleyiii: Same comment about statics vs `this`

packages/vscode-apollo/src/testRunner/jest-config.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const config = {
+  // This seems to have no effect, though the need may arise when running more tests
```
jbaxleyiii: What does `this` refer to here? Can we outline this file in a larger comment explaining the what and the why?

packages/vscode-apollo/src/testRunner/jest-vscode-environment.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+// Using import here results in `jest_environment_node_1.default`
+// The `.default` in this case doesn't exist. Interop issue with the built
+// version of the package?
+const NodeEnvironment = require("jest-environment-node");
+import * as vscode from "vscode";
+
+class VsCodeEnvironment extends NodeEnvironment {
+  constructor(config: any) {
```
jbaxleyiii: we shouldn't need the constructor + super since the super is the only action and is called with the config

packages/vscode-apollo/src/testRunner/jest-vscode-environment.ts
```diff
@@ -0,0 +1,43 @@
...
+    super(config);
+  }
+
+  public async setup() {
+    await super.setup();
+    this.global.vscode = vscode;
+
+    // For some reason this seems to be required for the Jest output to be streamed
+    // to the Debug Console.
+    process.stdout.write = this.logger;
```
jbaxleyiii: yikes haha. Not sure of the impact here, but we should probably reset these values in the teardown?

packages/vscode-apollo/src/testRunner/jest-vscode-environment.ts
```diff
@@ -0,0 +1,43 @@
...
+    process.stderr.write = this.logger;
+  }
+
+  public async teardown() {
+    this.global.vscode = {};
+    return await super.teardown();
+  }
+
+  public runScript(script: any) {
```
jbaxleyiii: Since this is just a proxy to the super, it shouldn't be needed?

packages/vscode-apollo/src/statusBar.ts
trevor-scheer: I thought the same with the class static and `this.loadingStateText` first, but it's a TS error!
image

The purpose for this is so that if the text changes, the value is updated for the test as well. I can either leave as is or go with the first option and export those constants.

packages/vscode-apollo/src/statusBar.ts
trevor-scheer: 👍

packages/vscode-apollo/src/testRunner/jest-config.ts
trevor-scheer: `This` refers to `runInBand` - it was here via prior art.
...it doesn't even exist in the docs so I'm just going to remove it.

To your second point, this is just a jest config - I can add an @see that points to the documentation. Additionally, I can add some comments about the less common stuff - `testEnvironment` and `setupTestFrameworkScriptFile` if you think that's useful?

packages/vscode-apollo/src/testRunner/jest-vscode-environment.ts
trevor-scheer: 👍 I'd be interested in exploring this with you at some point, maybe there's a better way to go about this.

packages/vscode-apollo/src/statusBar.ts
trevor-scheer: Resolved with 2afa864d200f34299fe9077d2313f3ba7814f031

packages/vscode-apollo/package.json
jbaxleyiii: I think we've tried to keep them all at the root if possible!

PR #837, opened 2018-12-17
Update README to point to the docs for apollo.config.js format

daniman:
I recently updated our Apollo config docs at apollographql.com/docs to cover the config format API, so this PR removes the duplication in `tooling`'s REAME of that API and links to our docs for details instead.


trevor-scheer:
Thanks for the update @daniman!

PR #840, opened 2018-12-18
Support status bar command and service stats logging

JakeDawkins:
## Goals
- [x] Add an action to the action bar item at the bottom of the VSCode editor that opens up the output for the Apollo extension and prints out basic service info
- [x] Add support on the language server for reporting that service info
- [x] Supports multiple projects open in the workspace (like in a monorepo). Stats printed are for the project associated with the open file. If no file is open, it just prints extension version

**Client projects when loaded successfully:**
screen shot 2019-01-10 at 4 34 45 pm

**Error when projects can't load stats (either because of misconfiguration or other language server/extension errors):**
screen shot 2019-01-10 at 4 35 20 pm

Note: Service projects currently aren't supported for stats, so clicking the status bar icon just prints the extension version and opens the output pane

### DIFF
--- a/packages/apollo-language-server/src/fileSet.ts
+++ b/packages/apollo-language-server/src/fileSet.ts
@@ -45,12 +45,18 @@ export class FileSet {
   allFiles(): string[] {
     return this.includes
       .flatMap(include =>
-        glob.sync(include, { cwd: this.rootPath, absolute: true })
+        glob.sync(include, {
+          cwd: this.configPath || this.rootPath,
+          absolute: true
+        })
       )
       .filter(
         filePath =>
           !this.excludes.some(exclude =>
-            minimatch(relative(this.rootPath, filePath), exclude)
+            minimatch(
+              relative(this.configPath || this.rootPath, filePath),
+              exclude
+            )
           )
       );
   }
--- a/packages/apollo-language-server/src/fileSet.ts
+++ b/packages/apollo-language-server/src/fileSet.ts
@@ -1,18 +1,22 @@
-import { relative } from "path";
+import { relative, dirname } from "path";
 import minimatch = require("minimatch");
 import * as glob from "glob";
 import { invariant } from "@apollographql/apollo-tools";
+import URI from "vscode-uri";
 
 export class FileSet {
+  private configPath?: string;
   private rootPath: string;
   private includes: string[];
   private excludes: string[];
 
   constructor({
+    configPath,
     rootPath,
     includes,
     excludes
   }: {
+    configPath?: string;
     rootPath: string;
     includes: string[];
     excludes: string[];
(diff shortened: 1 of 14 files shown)


packages/apollo-language-server/src/project/client.ts
```diff
@@ -83,6 +83,7 @@ export class GraphQLClientProject extends GraphQLProject {
   private _onSchemaTags?: NotificationHandler<[ServiceID, SchemaTag[]]>;
 
   private fieldStats?: FieldStats;
+  protected lastLoadDate?: number;
```
trevor-scheer: Don't think you need this one since it exists in the base.

packages/apollo-language-server/src/project/base.ts
```diff
@@ -106,6 +108,8 @@ export abstract class GraphQLProject implements GraphQLSchemaProvider {
 
   protected abstract initialize(): Promise<void>[];
 
+  abstract getProjectStats(): any;
```
trevor-scheer: Does it make sense to enforce a type here? Or only at the concrete implementation? Here makes sense to me but you've also got some more context than I do.

packages/vscode-apollo/src/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+    return;
+  } else if (stats.type === "client") {
+    client.outputChannel.appendLine("✅ Service Loaded from Engine!");
+    client.outputChannel.appendLine(`🆔 Service ID: ${stats.serviceId}`);
+    client.outputChannel.appendLine(`🏷 Schema Tag: ${stats.tag}`);
+    client.outputChannel.appendLine(
+      `📈 Number of Types: ${stats.types.total} ${
+        stats.types.client
+          ? `(${stats.types.client} client ${
+              stats.types.client > 1 ? "types" : "type"
```
abernix: Fly-by observation: Can `stats.types.client` be `0`?  If so — and to be defensive, in case it is in the future — perhaps this could be:

```suggestion
              stats.types.client === 1 ? "type" : "types"
```

...in order to avoid reading `0 type`, rather than `0 types`.

packages/vscode-apollo/src/extension.ts
```diff
@@ -86,6 +88,28 @@ export function activate(context: ExtensionContext) {
...
   client.onReady().then(() => {
+    commands.registerCommand("apollographql/showStats", () => {
+      const fileUri = window.activeTextEditor
+        ? window.activeTextEditor.document.uri.toString()
+        : null;
+
+      // if no editor is open, vscode returns something like output:extension-output-%234
+      // if an editor is open, this is something like file://Users/...
+      // this check is just for either a / or a \ in a fileUri
+      const fileOpen = fileUri && /.*[\/\\].*/gm.test(fileUri);
```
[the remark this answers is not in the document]
JakeDawkins: What about on windows? Wouldn't the uri be a different format?

packages/vscode-apollo/src/utils.ts
abernix: I like the idea of showing `0 client types`, for explicitness.  I could probably be convinced otherwise though.

packages/apollo-language-server/src/fileSet.ts
```diff
@@ -1,18 +1,22 @@
-import { relative } from "path";
+import { relative, dirname } from "path";
 import minimatch = require("minimatch");
 import * as glob from "glob";
 import { invariant } from "@apollographql/apollo-tools";
+import URI from "vscode-uri";
```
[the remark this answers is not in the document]
abernix: @JakeDawkins 	As long as it matches here, I think that's good enough.  We probably have a lot of variety, but I like `URI`.

packages/vscode-apollo/src/extension.ts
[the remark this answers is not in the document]
abernix: @JakeDawkins 	The first option is fine, but I'd argue that a file-based URI ([by definition](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier)) should always start with `file://` — regardless of the remaining portion.

PR #896, opened 2019-01-16
Fixed config failing to load .env files for editors

JakeDawkins:
## Issues

1. The CLI would load a `.env` file containing an `ENGINE_API_KEY`, but the editor integration would not. 
2. If a `type` was not passed into the loadConfig (which happens when run from the editor integration), and the config was a client config, we got into a weird state where all the defaultConfig's were getting merged together.

## Causes

**For Issue 1:**
* The editor and CLI were both providing a different format for `configPath` (where to look for a `.env`). The editor was passing the dir only (`/User/Jake/my-project`) whereas the CLI was passing the filename (`/User/Jake/my-project/apollo.config.js`).
* The CLI would load the `.env` because it was looking at `configPath.dir` (the parent), but when the config tried to process the same way, it would use the parent of the cwd, giving us a path one level too high (`/User/Jake/` in this example)

**For Issue 2:**
* The editor doesn't know the `type` of the config. The CLI does, because it's run with `client:*` or `service:*`.

## What was done

**For Issue 1:**

* fixed cli to only pass in the parent dir of config when using config flag

**For Issue 2:**

* infer config type when not explicitly passed to loadConfig

**Misc:**

* fix improper symlink paths
* replace last occurence of rm -rm with rimraf

### DIFF
--- a/packages/apollo-language-server/src/config.ts
+++ b/packages/apollo-language-server/src/config.ts
@@ -1,7 +1,7 @@
 import * as cosmiconfig from "cosmiconfig";
 import { LoaderEntry } from "cosmiconfig";
 import TypeScriptLoader from "@endemolshinegroup/cosmiconfig-typescript-loader";
-import { parse, resolve } from "path";
+import { resolve } from "path";
 import { readFileSync, existsSync } from "fs";
 import { merge } from "lodash/fp";
 
@@ -282,7 +282,7 @@ export const loadConfig = async ({
   let engineConfig = {},
     nameFromKey;
   const dotEnvPath = configPath
-    ? resolve(parse(configPath).dir, ".env")
+    ? resolve(configPath, ".env")
     : resolve(process.cwd(), ".env");
 
   if (existsSync(dotEnvPath)) {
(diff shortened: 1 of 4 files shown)


packages/apollo-language-server/src/config.ts
```diff
@@ -282,7 +282,7 @@ export const loadConfig = async ({
   let engineConfig = {},
     nameFromKey;
   const dotEnvPath = configPath
-    ? resolve(parse(configPath).dir, ".env")
+    ? resolve(configPath, ".env")
```
trevor-scheer: `parse` import is unused now

packages/apollo-language-server/src/config.ts
```diff
@@ -297,6 +297,23 @@ export const loadConfig = async ({
   }
 
   const resolvedName = name || nameFromKey;
+
+  // The CLI passes in a type when loading config. The editor extension
+  // does not. So we determine the type of the config here, and use it if
+  // the type wasn't explicitly passed in.
+  let resolvedType;
```
trevor-scheer: Is this a TS error (implicit any)?
`let resolvedType: "client" | "service";`

PR #897, opened 2019-01-17
[AT-10] Add helpful errors / messaging around config files

trevor-scheer:
This PR adds new warning states and communication within the extension in order to help users resolve config issues more easily and know when there's a problem.

noconfig

Also included with this PR is a bit of a shift in how errors are handled. After speaking with James about it, we both agreed that this would be a scalable approach that will work well with clients consuming the language server or using its parts programmatically.
* Internals of the language server should throw errors, as opposed to catching them and logging at any point.
* The server itself should catch these errors at the top level and message them out via the connection, so they can be consumed by any client listening on that connection.
* Clients using the language server parts programmatically should catch and handle these errors in an appropriate manner based on the requirements of the client.

### DIFF
--- a/packages/apollo-language-server/src/project/client.ts
+++ b/packages/apollo-language-server/src/project/client.ts
@@ -94,7 +94,7 @@ export class GraphQLClientProject extends GraphQLProject {
       // the URI of the folder _containing_ the apollo.config.js is the true project's root.
       // if a config doesn't have a uri associated, we can assume the `rootURI` is the project's root.
       rootURI: config.configDirURI || rootURI,
-      includes: config.client.includes,
+      includes: [...config.client.includes, ".env", "apollo.config.js"],
       excludes: config.client.excludes
     });
 
--- a/packages/apollo-language-server/src/project/client.ts
+++ b/packages/apollo-language-server/src/project/client.ts
@@ -140,6 +140,11 @@ export class GraphQLClientProject extends GraphQLProject {
     };
   }
 
+  updateConfig(config: ClientConfig) {
+    this.config = config;
+    return this.initialize();
+  }
+
   onDecorations(handler: (any: any) => void) {
     this._onDecorations = handler;
   }
--- a/packages/apollo-language-server/src/project/client.ts
+++ b/packages/apollo-language-server/src/project/client.ts
@@ -140,11 +140,6 @@ export class GraphQLClientProject extends GraphQLProject {
     };
   }
 
-  updateConfig(config: ClientConfig) {
-    this.config = config;
-    return this.initialize();
-  }
-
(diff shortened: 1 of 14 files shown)


packages/vscode-apollo/src/extension.ts
```diff
@@ -3,90 +3,63 @@ import {
...
+import { LanguageClient } from "vscode-languageclient";
 import {
-  printStatsToClientOutputChannel,
-  printNoFileOpenMessage
+  printNoFileOpenMessage,
+  printStatsToClientOutputChannel
 } from "./utils";
-const { version, referenceID } = require("../package.json");
+
+const { version } = require("./package.json");
```
JakeDawkins: Looks like this path got messed up in rebase or something :) 

```suggestion
const { version } = require("../package.json");
```

packages/vscode-apollo/package.json
```diff
@@ -40,7 +40,7 @@
   ],
   "icon": "images/icon-apollo-teal-400x400.png",
   "activationEvents": [
-    "workspaceContains:**/apollo.config.js"
+    "workspaceContains:**/package.json"
```
JakeDawkins: I'm not sure about this. This would try to load the extension on _any_ project with a package.json, even ones we don't care about.

Is there any other way to know if a project is _supposed_ to be a graphql project, other than with the inclusion of an Apollo config?

JakeDawkins:
@trevor-scheer since you're already working on errors, can you add a more detailed error message under `Apollo-language-server/src/schema/providers/index.ts > line 54`? 😄

Like `"No schema provider was created, because the project type was unable to be resolved from your config. Please add either a client or service config. For more information, please refer to https://bit.ly/2ByILPj"`

packages/apollo-language-server/src/project/client.ts
```diff
@@ -140,6 +140,11 @@ export class GraphQLClientProject extends GraphQLProject {
     };
   }
 
+  updateConfig(config: ClientConfig) {
```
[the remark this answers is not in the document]
trevor-scheer: Yep! Good call :)

packages/vscode-apollo/package.json
trevor-scheer: I decided this was necessary in order to launch the extension and "listen" for the addition of an apollo config if none exists. This way, a project can go from no config -> config w/o reloading VSCode, and expect the extension to pick up on the changes and initialize the project.

I'm open to other ideas here!

JakeDawkins: Is it possible for the error where there' no config to be a toast-style notification that disappears after a few seconds then? This one you have to dismiss I believe right?

JakeDawkins:
:yolo:

PR #915, opened 2019-01-22
[WIP] Allow Typescript global types in global NS

steelbrain:
This PR adds an option that allows users to define all the GQL enums in the global namespace so they don't have to import those types everywhere.

This was already possible for flow with their `[libs]` config, but TS requires the file itself to specify its scope. I've tested and have tried it, it works.

TODO: 

- [ ] Squash commits
- [ ] Add Tests


zionts:
Hi @steelbrain , thanks for taking the time to work on this! Since it has been almost a year 😮 since this was opened up in WIP mode, I'm going to close it out for now, but if you'd like to pick the work back up, please feel free!

PR #921, opened 2019-01-24
Apollo Init - Config builder

trevor-scheer:
Create a new command for the Apollo CLI to help users create their config files. This is a step-by-step set of interactive prompts that ultimately creates the config (and .env entry) based on a user's responses.

Hopefully, this simplifies and "train tracks" the process of building a config file so that users can land on the config they need w/o any hassle. 🎉 🎉

initsmall


JakeDawkins:
I'm going to close this for now, since it's probably a little stale and I don't have the bandwidth to maintain this at the moment. I _do_ think this would be a good thing to revive and implement for the next version of the config though!

PR #934, opened 2019-01-28
[AT-36] Improve client:check output

trevor-scheer:
* Update CLI to replace checkOperations with validateOperations mutation.
* Leverage new mutation to provide file specific data for validation errors.
* Display messages grouped by query, with a clickable link to each query

image

**Sidenote, this ^ feels pretty good but we can improve**
As far as flow-y errors go, I know we can get there but as I explored the idea, it started to feel like a solid amount of work in and of itself. The short explanation is that we can currently get close, but since we don't store the raw document in the data model, we lose the source location data we would need to give a correct location.

Some more explanation: what we do now is pass around an actual `Document` and `print()` it when it's time to pass the string to the backend - we can't use this because the source formatting is lost. Sooo, when we want to perfect this, we'll need to pass that information around as well and do a little bit of math with the validation results that come back from the server.


jbaxleyiii:
This is so incredible

packages/apollo/src/diff/index.ts
```diff
@@ -36,21 +40,21 @@ export interface SchemaChange {
...
+      break;
   }
-
-  return {
-    type: color(change.type),
-    code: color(change.code),
-    description: color(change.description)
-  };
-};
+  return `    [${index}] ${color(type)}    ${description}`;
```
JakeDawkins: What's the benefit of adding the index to the error log?

trevor-scheer: I was already on the fence about leaving this in, so I pulled it since you're asking 😛
It really doesn't add anything useful.

PR #941, opened 2019-01-29
Add Issue and PR Templates

JakeDawkins:
Just to make our lives a little easier, I figured I'd copy and modify the issue/PR templates from other repos that we have :)


JakeDawkins:
Secretly I just wanted there to be a checkbox for CHANGELOG entries, but while I'm at it, I may as well change _everything_

PR #943, opened 2019-01-30
Improve Scala codegen to make results feel more like case classes

shadaj:
The change to use traits left some codebases pretty source incompatible, as they depended on APIs that are automatically generated for case classes. This adjusts the codegen to generate similar APIs so that query results have the feel of being case classes without the performance overhead.

### DIFF
--- a/packages/apollo-codegen-scala/src/language.ts
+++ b/packages/apollo-codegen-scala/src/language.ts
@@ -31,7 +31,6 @@ export function packageDeclaration(
 ) {
   generator.printNewlineIfNeeded();
   generator.printOnNewline(`package ${pkg}`);
-  generator.popScope();
 }
 
 export function objectDeclaration(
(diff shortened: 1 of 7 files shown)


packages/apollo-codegen-scala/src/language.ts
```diff
@@ -120,12 +119,12 @@ export function methodDeclaration(
     .join(", ");
 
   generator.printOnNewline(
-    `def ${methodName}(${paramsSection})` + (closure ? " = " : "")
+    `def ${methodName}(${paramsSection})` + (closure ? " =" : "")
```
trevor-scheer: Just double checking that the space removal was intentional, since it looks like it could be accidental 🤷‍♂️

shadaj: Yeah, this is intentional. While reviewing snapshots I noticed that there was an extra space between the `=` and the following `{` so this fixes it.

shadaj:
Huh, the Azure tests are failing now, but seems unrelated to the changes in this PR.

PR #953, opened 2019-01-31
Support validation parameters for service:check

trevor-scheer:
This PR adds functionality for 3 new flags in the service:check command:

* validationPeriod - the duration (from now) to validate the service against
* queryCountThreshold - minimum number of requests within the requested time window for a query to be considered
* queryCountThresholdPercentage - number of requests within the requested time window for a query to be considered, relative to total request count.

Also included in this PR - better error messaging:
* Reorders error throwing in `EngineClient`, so that errors reported from the service will be thrown
with priority over the generic "noServiceError" when data isn't present.

TODO:

### DIFF
--- a/packages/apollo/src/commands/service/check.ts
+++ b/packages/apollo/src/commands/service/check.ts
@@ -2,9 +2,9 @@ import { flags } from "@oclif/command";
 import { table } from "heroku-cli-util";
 import { introspectionFromSchema } from "graphql";
 import chalk from "chalk";
-
 import { gitInfo } from "../../git";
 import { ProjectCommand } from "../../Command";
+import { validateHistoricParams } from "../../utils";
 
 // TODO These types can/should be generated from the engine schema when we
 // dogfood codegen inside of the apollo-tooling repo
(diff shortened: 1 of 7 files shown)


packages/apollo/package.json
```diff
@@ -60,6 +60,7 @@
     "heroku-cli-util": "^8.0.9",
     "listr": "^0.14.1",
     "lodash": "^4.17.10",
+    "moment": "^2.24.0",
```
jbaxleyiii: Can we use something lighter than moment like date-fns?

packages/apollo/src/commands/service/check.ts
```diff
@@ -69,3 +93,57 @@ export default class ServiceCheck extends ProjectCommand {
     return;
   }
 }
+
+// Validates flag variables and:
+//   1) returns an input object for the checkSchema query (HistoricQueryParameters)
+//   2) or throws an error
+export function validateHistoricParams({
```
jbaxleyiii: This probable deserves to live on its own outside of this service command. I think we should try to only have the command class when possible in these files (I know I added something else earlier on haha)

packages/apollo/package.json
trevor-scheer: Looked for anything to support durations with the ease-of-use + granularity that ISO 8601 gives us and no love, + discussed offline. Going to leave as is. If having moment gets to heavy, we can write a duration parser.

PR #961, opened 2019-02-02
Pin core-js version in apollo-env

bfirsh:
The API was changed in 3.0.0-beta.12 ([released 2 hours ago](https://github.com/zloirock/core-js/releases)) causing this error in my application:

```
Error: Cannot find module 'core-js/proposals/array-flat-and-flat-map'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:603:15)
    at Function.Module._load (internal/modules/cjs/loader.js:529:25)
    at Module.require (internal/modules/cjs/loader.js:657:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/ben/p/miles/node_modules/apollo-env/src/polyfills/array.ts:1:1)
    at Module._compile (internal/modules/cjs/loader.js:721:30)
    at Module._compile (/Users/ben/p/miles/example/node_modules/pirates/lib/index.js:83:24)
    at Module._extensions..js (internal/modules/cjs/loader.js:732:10)
    at Object.newLoader [as .js] (/Users/ben/p/miles/example/node_modules/pirates/lib/index.js:88:7)
    at Module.load (internal/modules/cjs/loader.js:620:32)
```

It seems sensible to pin to a specific beta version while the API is moving around. When 3.0.0 is released, then it could be switched back to `^3.0.0`.

### DIFF
--- a/packages/apollo-env/package.json
+++ b/packages/apollo-env/package.json
@@ -20,7 +20,7 @@
     "node": ">=8"
   },
   "dependencies": {
-    "core-js": "^3.0.0-beta.3",
+    "core-js": "3.0.0-beta.3",
     "node-fetch": "^2.2.0"
   }
 }


koenpunt:
We can assume that the moved dependencies will not move back, so I think it's a good idea to pin to the current latest beta, and update the imports. 

Using `patch-package` I applied this patch:

```patch
diff --git a/node_modules/apollo-env/lib/polyfills/array.js b/node_modules/apollo-env/lib/polyfills/array.js
index 9d5bb1f..7dc6721 100644
--- a/node_modules/apollo-env/lib/polyfills/array.js
+++ b/node_modules/apollo-env/lib/polyfills/array.js
@@ -1,4 +1,5 @@
 "use strict";
 Object.defineProperty(exports, "__esModule", { value: true });
-require("core-js/proposals/array-flat-and-flat-map");
+require("core-js/es/array/flat");
+require("core-js/es/array/flat-map");
 //# sourceMappingURL=array.js.map
diff --git a/node_modules/apollo-env/lib/polyfills/object.js b/node_modules/apollo-env/lib/polyfills/object.js
index f302992..c215afb 100644
--- a/node_modules/apollo-env/lib/polyfills/object.js
+++ b/node_modules/apollo-env/lib/polyfills/object.js
@@ -1,4 +1,4 @@
 "use strict";
 Object.defineProperty(exports, "__esModule", { value: true });
-require("core-js/proposals/object-from-entries");
+require("core-js/es/object/from-entries");
 //# sourceMappingURL=object.js.map
```

JakeDawkins:
Thanks so much for this @bfirsh I'll publish this now

JakeDawkins:
Released in `apollo@2.4.3`!

PR #967, opened 2019-02-05
apollo: remove unused schemaDiff

evans:
This removes unused and undocumented code that is a holdover from
apollo@1 and tech debt.  Since `apollo` is a commandline tool, this code
should not be used outside of Apollo's internal Platform tools, so I
suggest that the code be removed without a major version bump

TODO:


trevor-scheer:
Thanks for the cleanup 🎉

PR #1007, opened 2019-02-11
[AT-65] Added warning when no files found in fileset

### LINKED ISSUE #702: VSCode - Show error when no files are found
Extension can load with no issues but provide no feedback if the `{client: include: '**'}` isn't specified correctly.

### LINKED ISSUE #883: VS Code - autocomplete not working in all files
When `apollo.config.js` is thus:
```
module.exports = {
    client: {
        name: 'graphitti schema',
        service: 'graphitti-schema',
    }
};
```
and `.env` set appropriately to point to Apollo Engine service, VS Code shows both autocomplete and metrics working in `src/see.graphql` file:
works

But when `apollo.config.js` is updated to include files in other directories:
```
module.exports = {
    client: {
        name: 'graphitti schema',
        service: 'graphitti-schema',
        includes: [__dirname+ '/client/**/*.js', './client/**/*.graphql', './src/**/*.graphql']
    }
};
```

VS Code continues to show metrics for the included js files, not does not show autocomplete or type information on hover:

bug

What am I missing?

### LINKED ISSUE #1002: [apollo-graphql] vscode gql tag functionality broken when relative globs used in includes config
Auto-complete and type hints do not work in `gql` tagged template literals when the includes config setting in apollo.config.js file has relative globs in it (globs that start with `./`).  Some of this is documented in issue #690.

**Intended outcome:**

When I set my includes glob to `includes: ['./examples/**/*.{ts,tsx}'],` I expect that project files with `gql` tagged template literals that match that glob work with apollo-graphql extension auto-complete and type hint hover functionality in the graphql query. And when clicking the status bar `ApolloGraphQL` button when in a file that matches the glob, a status message similar to below is displayed.
```
------------------------------
🚀 Apollo GraphQL v1.4.5
------------------------------
✅ Service Loaded!
🆔 Service ID: OSTK GraphQL Demo1
🏷 Schema Tag: current
📈 Number of Types: 45 (0 client types)
------------------------------
```

**Actual outcome:**

apollo-graphql extension functionality doesn't work in vscode. There are no type hints or auto-complete based on schema. Moreover, when clicking the status bar `ApolloGraphQL` button when in a file other than `apollo.config.js` the error message below is shown in the OUTPUT panel. 
```
------------------------------
🚀 Apollo GraphQL v1.4.5
------------------------------
❌ Service stats could not be loaded. This may be because you're missing an apollo.config.js file or it is misconfigured. For more information about configuring Apollo projects, see the guide here (https://bit.ly/2ByILPj).
```
Changing the includes config to `includes: ['examples/**/*.{ts,tsx}'],` (relative but without `./` at the beginning) makes it work as expected.

Open an apollo-graphql based project that has client.service.localSchemaFile, client.service.url, or apollo engine set up and working properly. Set up a folder outside of the `src/` folder (this should skirt the default includes - in my case it was `examples/`) with a component the uses a `gql` tagged template literal. Try different includes config setting with and without `./` at the beginning to see the difference in behaviour

**Versions**
apollo@2.4.4
vscode@1.31.0
apollo-graphql@1.4.5

JakeDawkins:
Fixes #1002 
Fixes #702 
Fixes #883

## File count warnings

Whenever the VS Code extension opens a file, it tries to find a projectForFile. If the includes or excludes field in the Apollo Config file is misconfigured, it may never actually find the project for a given file.

The way the status bar/stats reporting command is setup, if you click the icon to report stats while having one of those files open, it will not find the associated project and warn that the project may be misconfigured. Which is okay, but I think we can do better.

On project creation, this PR checks the FileSet to see how many files are included. If there are 0 files in the FileSet (excluding `.env` and configs), we issue a warning but not an error. that's because there's no way to _know_ that there are supposed to be files there. Maybe the user is bootstrapping a new project and there just aren't any files yet. 

## Includes/Excludes fixes

Because of how we were using glob/minimatch on the includes/excludes, there were some issues:

1. includes/excludes that were relative (i.e. `./src/__tests__`) would not work with `FileSet.includesFile`. To fix, `includesFile` now uses the results of `allFiles` (#1002)
2. `excludes` wasn't properly being filtered out in `FileSet.allFiles`, and wouldn't allow proper filtering if the glob was relative (the same issue that originally was in `includesFile`

TODO:

screen shot 2019-02-11 at 5 11 07 pm
screen shot 2019-02-11 at 5 13 05 pm


packages/apollo-language-server/src/fileSet.ts
```diff
@@ -25,6 +27,26 @@ export class FileSet {
...
+
+    /**
+     * This function is used in the Array.filter function below it to remove any .env files and config files.
+     * If there are 0 files remaining after removing those files, we should warn the user that their config
+     * may be wrong. We shouldn't throw an error here, since they could just be initially setting up a project
+     * and there's no way to know for sure that there _should_ be files.
+     */
+    const filterConfigAndEnvFiles = (path: string) =>
+      !(
+        path.match(/.*(\.env|apollo\.config\.js).*/) ||
```
trevor-scheer: Just for simplicity, I'll suggest:
`path.includes("apollo.config") || path.includes(".env")`

Just a minor preference, and not a big deal if you prefer the regex.

We might not account for this in all places, but we should allow `apollo.config.ts`.

packages/apollo-language-server/src/fileSet.ts
```diff
@@ -25,27 +27,42 @@ export class FileSet {
...
-        filePath =>
-          !this.excludes.some(exclude =>
-            minimatch(relative(this.rootURI.fsPath, filePath), exclude)
-          )
-      );
+    // since glob.sync takes a single patttern, but we allow an array of `includes`, we can join all the
+    // `includes` globs into a single pattern and pass to glob.sync The `ignore` option does, however, allow
+    // an array of globs to ignore, so we can pass it in directly
+    const joinedIncludes = `{${this.includes.join(",")}}`;
+    return glob.sync(joinedIncludes, {
```
trevor-scheer: Awesome 😄

PR #1008, opened 2019-02-12
Fix Scala codegen emitting fields under names that don't match the schema

shadaj:
Fixes an oversight in the trait rewrite, since variable names in the generated traits determine the JS object that will be sent to Apollo but those variable names are transformed versions of the schema names, so there could be mismatches. This PR annotates the variables to tell Scala.js to emit those variables under the original schema names.

This can be included in a patch release, it doesn't have any breaking changes.


packages/apollo-codegen-scala/src/codeGeneration.ts
```diff
@@ -443,7 +443,7 @@ function dataContainerDeclaration(
     insideCompanion
   }: {
     name: string;
-    properties: Property[];
+    properties: (Property & ({ name: string } | { responseName: string }))[];
```
trevor-scheer: Would typing `properties` correctly here allow you to remove the `any`s below? Or maybe the typing here isn't wrong, but the `any` is just leftover as an oversight? I'll let you find out what works 😄

I _think_ you want

```javascript
(Property & {
  name?: string;
  responseName?: string;
})[];
```

shadaj: The typing here is a bit weird, I know that at least one of `name` and `responseName` will always be defined, but TypeScript isn't handling that too well. Ideally it would understand that in the situation where `name` is `undefined`, then `responseName` will be defined for sure. But instead neither of the properties show up on the object, so I did the `any` typing.

Now that I think of it, it's probably better just to define the type the way you did, so at least there won't be typos with grabbing values from the fields. I'll add a commit to change that in a bit.

trevor-scheer: I see what you mean - I actually see how your original typing is more correct, but I couldn't make TS happy either in a small repl. Let's keep as is since this is fine for all intents and purposes.

trevor-scheer:
Thanks @shadaj!

PR #1030, opened 2019-02-16
Clean up configuration code

trevor-scheer:
This PR brings some sanity to the configuration code. Functionally, it should be a no-op.

* Create a config _folder_, in order to break up `config.ts` into smaller, more manageable parts
* Move utility functions out of `config.ts` into `utils.ts`
* `loadConfig` lives in its own file now, along with related cosmiconfig code
* Remove some duplication
* Clean up functions with some minor TS improvements

> Want to add some tests for util functions, probably will get to that

TODO:

### DIFF
--- a/packages/apollo-language-server/src/config/utils.ts
+++ b/packages/apollo-language-server/src/config/utils.ts
@@ -35,7 +35,7 @@ export function getServiceName(config: ApolloConfigFormat) {
   if (config.service) return config.service.name;
   if (config.client) {
     if (typeof config.client.service === "string") {
-      return parseServiceSpecificer(config.client
+      return parseServiceSpecifier(config.client
         .service as ServiceSpecifier)[0];
     }
     return config.client.service && config.client.service.name;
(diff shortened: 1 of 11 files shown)


JakeDawkins:
I was debating working on this! I started some testing work around the Config a couple days ago. I may PR Onto this branch if you’re cool with that

packages/apollo-language-server/src/config/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export function parseServiceSpecificer(specifier: ServiceSpecifier) {
+  const [id, tag] = specifier.split("@").map(x => x.trim());
+  return [id, tag] as ServiceIDAndTag;
+}
+
+// take a config with multiple project types and return
+// an array of individual types
+export function projectsFromConfig(
```
JakeDawkins: I feel like this can just go inside the ApolloConfig base class. It's not used anywhere else, and I can't think of a use case outside right now. Maybe you can though 😀

trevor-scheer: ✅76e0ded

I see you did that on your PR so I just stole your code to make your life simpler 😛

JakeDawkins:
@trevor-scheer I merged my tests. I also changed the config filenames in the codegen tests as to not make vscode go crazy 😎

PR #1055, opened 2019-02-26
Fix config loading issues

JakeDawkins:
Fixes 2 bugs:
- VS Code not recognizing a file as being part of the `project` (first commit)
	- Changed it from using relative filepath (which is wrong) to use the full path as expected.
- In monorepos, if there are multiple configs, and one fails to load, it won't load _any_ configs.
	- Added warning for failed config including path to broken config.

screen shot 2019-02-26 at 3 09 20 pm

TODO:

### DIFF
--- a/packages/apollo-language-server/src/workspace.ts
+++ b/packages/apollo-language-server/src/workspace.ts
@@ -117,25 +117,26 @@ export class GraphQLWorkspace {
     const apolloConfigFolders = new Set<string>(apolloConfigFiles.map(dirname));
 
     // go from possible folders to known array of configs
+    let foundConfigs: ApolloConfig[] = [];
+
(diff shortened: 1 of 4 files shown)


packages/apollo-language-server/src/workspace.ts
```diff
@@ -117,25 +117,26 @@ export class GraphQLWorkspace {
...
+          const existingProjects =
+            this.projectsByFolderUri.get(folder.uri) || [];
+
+          this.projectsByFolderUri.set(folder.uri, [
+            ...existingProjects,
+            ...projectsForConfig
+          ]);
+        })
+        .catch(error => console.error(error))
+    );
```
trevor-scheer: Spoke offline, we recognized the need to still await all the promises before calling `this._onConfigFilesFound`.

PR #1062, opened 2019-02-28
Remove the default --tag flag from client commands

JakeDawkins:
`client:codegen` was defaulting to `--tag=current` when not specifying a flag. That's because there was a default set. It didn't matter if you set another tag in your config, because the CLI flags took precedence over any other configs.

This PR just removes the default _flag_, not the default _tag_. Hopefully this should be non-breaking in any way since the `current` tag fallback is already set other places as a last-resort item.

fallbacks: 
- https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-language-server/src/config/config.ts#L147
- https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-language-server/src/engine/operations/schemaTagInfo.ts#L4
- https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-language-server/src/schema/providers/engine.ts#L54

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - `apollo`
   - Use generated Typescript types via client:codegen [#1016](https://github.com/apollographql/apollo-tooling/pull/1016)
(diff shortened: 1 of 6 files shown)


trevor-scheer:
Woo generated type check!

PR #1066, opened 2019-03-03
Introduced `type` keyword among the ones that need to be escaped

espinhogr:
TODO:

Tests are only there for the general case and not for each single keyword


espinhogr:
I don't know why Generated Types Check is not passing as I haven't touched the typescript generation at all.

trevor-scheer:
@espinhogr thanks for the PR! Looks like it just needs a rebase against `master`.

espinhogr:
@trevor-scheer  I've just rebased to the master but it seems to fail in a test on Windows, is this something known or you think it's the change I've introduced?

trevor-scheer:
@espinhogr we've got something flaky in the tests, I re-ran and it looks good. Thanks again!

espinhogr:
@trevor-scheer  Brilliant, happy to help!

PR #1085, opened 2019-03-06
apollo: ensure targetUrl is defined when printing

evans:
The `targetUrl` can be undefined, so we need to ensure it's defined, otherwise we'll print the line without anything actionable 😰 


trevor-scheer:
Thanks @evans!

PR #1096, opened 2019-03-08
Bump CircleCI cache salt to hopefully solve peculiar failures.

abernix:
Right now each CircleCI test run is resulting in different failures.  For
example, each of these were run on the same branch one after each other with
no changes:

* https://circleci.com/workflow-run/60706f8f-1059-496b-a7af-2d24a20d546e
* https://circleci.com/workflow-run/358e33c1-a84a-41ab-9ebd-013507f088f6
* https://circleci.com/workflow-run/89788740-a884-4813-87b3-e4f997362ca9

This is a first step in debugging, and matches the seed pattern used in
other repositories in our organization.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,13 +22,13 @@ common_test_steps: &common_test_steps
     - restore_cache:
         keys:
           # When lock file changes, use increasingly general patterns to restore cache
-          - npm-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "package-lock.json" }}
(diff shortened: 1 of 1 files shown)


abernix:
This is a precautionary move and for debugging.

abernix:
I see clearly that it's there now, but I didn't think to check for a `CACHE_VERSION` environment variable as the seed since that deviates from all the other repositories in our organization!  While I'm intrigued by that pattern, CircleCI doesn't automatically re-run when we bump it in their settings.  🤔 	Maybe need to revisit this, but this resolved my concern for now.

PR #1104, opened 2019-03-11
Add graphqls as a supported file extension for VS Code

trevor-scheer:
The VS Code extension supports graphqls, but we don't list it as such.


trevor-scheer:
@abernix as requested, see 35f7e503c476480b11054f1df60b28ca7c6992b3.

This is a totally benign PR, on a branch that's current with master. After making my changes, I figured I'd see what happens with an `npm install` and here's the result!

PR #1108, opened 2019-03-12
Add ability to download schema to file

JakeDawkins:
Client projects could already download a schema from engine or an endpoint for codegen. This just exposes a command to download a schema to an output file, rather than just keeping it in memory.

I also updated the description for the service:download command and added an error to be more explicit that service:download is supposed to be used for `local` endpoints. Not registry downloads.

TODO:

Screen Shot 2019-03-12 at 3 38 21 PM

### DIFF
--- a/packages/apollo/src/commands/service/download.ts
+++ b/packages/apollo/src/commands/service/download.ts
@@ -1,6 +1,7 @@
 import { flags } from "@oclif/command";
 import { introspectionFromSchema } from "graphql";
 import { writeFileSync } from "fs";
+import chalk from "chalk";
 import { ProjectCommand } from "../../Command";
 
 export default class ServiceDownload extends ProjectCommand {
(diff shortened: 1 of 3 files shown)


packages/apollo/src/commands/service/download.ts
```diff
@@ -1,12 +1,12 @@
...
 import { introspectionFromSchema } from "graphql";
 import { writeFileSync } from "fs";
-
 import { ProjectCommand } from "../../Command";
 
 export default class ServiceDownload extends ProjectCommand {
   static aliases = ["schema:download"];
-  static description = "Download the schema from your GraphQL endpoint.";
+  static description =
+    "Download a schema from a locally running GraphQL endpoint.";
```
trevor-scheer: I think this is safe to leave as is - you can download from any endpoint. You just can't use this command to download from engine.

JakeDawkins: You _can_ but I added that just to clarify the _purpose_ of the command.

trevor-scheer: Do we consider it an improper use case to download the schema from a graph that isn't running locally? I actually don't know the answer to that question, but I'd imagine there's a valid use case in there.

I think downloading from a locally running server is just one use case, but really this command is set up to handle others as well.

alexpyoung:
Hey thanks for putting this together! Any thoughts if this might get released soon?

legopin:
Would it be possible to release the client download feature first while we deliberate on the service:download command?

It would be very helpful to aid in collaboration among frontend and backend devs for new features

JakeDawkins:
@alexpyoung @legopin `apollo@2.10.0` has this in it! Sorry for the extended delay :)

https://github.com/apollographql/apollo-tooling#apollo-clientdownload-schema-output

legopin:
@JakeDawkins thanks for providing this feature. Although I do want to confirm that this feature saves the introspection result of the schema instead of the schema in SDL form right? Is there a specific reason why one is favored over the other?

JakeDawkins:
Yep @legopin! This command saves the JSON encoded introspection result. We use this format because it's easier to consume by other tooling for things like codegen and more commonly fragment matchers in apollo-client. If needed, you could always generate sdl from this result, but most of the time, JSON is fine

PR #1114, opened 2019-03-13
Remove second parameter to `getOperationManifestFromProject`, always empty.

abernix:
The `getOperationManifestFromProject` can never have anything passed to it
besides an empty string, so we'll just make that the default behavior.

This differs slightly from the similarly named
`defaultEngineReportingSignature`, which does accept the `operationName`.

cc @trevor-scheer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## Upcoming
 
-## `apollo-graphql@0.2.0`
+## `apollo-graphql@0.2.1`
 
-- `apollo-graphql@0.2.0`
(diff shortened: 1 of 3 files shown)


abernix:
Just kidding, this is not the right fix.  We are currently _not_ passing the `operationName` here:

https://github.com/apollographql/apollo-tooling/blob/2a53ca12064b7b5a90a5fc2452c068857548804c/packages/apollo/src/utils/getOperationManifestFromProject.ts#L21

...we should be! Documents with multiple operations are individually registered with the operation registry.

PR #1121, opened 2019-03-15
apollo: fix remoteUrl in service:check [UI-258]

evans:
The remoteUrl can no longer be the slug, since we can't assume that the
feature is going to be used exclusively in GitHub and create an accurate
link the frontend. Additionally we add some logging in order to help
debug.

> Note: My first though was be to make the logging protected by a
verbose flag. Did I miss that functionality?

TODO:

### DIFF
--- a/packages/apollo/src/git.ts
+++ b/packages/apollo/src/git.ts
@@ -4,6 +4,7 @@ import * as ci from "env-ci";
 import { gitToJs } from "git-parse";
 import * as git from "git-rev-sync";
 import { pickBy, identity } from "lodash";
+import Command from "@oclif/command";
 
 const findGitRoot = (start?: string | string[]): string | void => {
   start = start || process.cwd();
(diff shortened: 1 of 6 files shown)


evans:
The `remoteUrl` regex matching in the test:

image

evans:
The azure pipelines fail with an unrelated error:
 
image

packages/apollo/src/git.ts
```diff
@@ -34,22 +35,22 @@ export interface GitContext {
   branch?: string;
 }
 
-export const gitInfo = async (): Promise<GitContext | undefined> => {
+export const gitInfo = async (
+  log: Command["log"]
```
trevor-scheer: Might want the type of log to be something a bit more generic, since we're passing in `console.log` for testing.

Paraphrasing and best guessing, but something like this?
`logger: (...args: string[]) => void`

This is a non-blocking comment.

evans: hmm they seem to have the same sort of signature `log(message?: string, ...args: any[]): void;` and I think we want to know when the types of the Command.log become incompatible

PR #1142, opened 2019-03-26
UI-307: Add `--markdown` option to `schema:check`

justinanastos:
TODO:


packages/apollo/src/commands/service/check.ts
```diff
@@ -27,6 +30,73 @@ const formatChange = (change: Change) => {
...
+  const { validationConfig } = diffToPrevious;
+
+  if (!validationConfig) {
+    throw new Error(
+      "checkSchemaResult.diffToPrevious.validationConfig missing"
+    );
+  }
+
+  // This will always return a negative number of days. Use `-` to make it positive.
+  const days = -moment()
```
evans: I found the formatting a bit hard to understand. Can we use [`Math.abs()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/abs)?

packages/apollo/src/Command.ts
```diff
@@ -214,10 +214,14 @@ export abstract class ProjectCommand extends Command {
   }
 
   async runTasks<Result>(
-    generateTasks: (context: ProjectContext) => ListrTask[]
+    generateTasks: (context: ProjectContext) => ListrTask[],
+    options?: Listr.ListrOptions | ((ctx: ProjectContext) => Listr.ListrOptions)
   ): Promise<Result> {
     const tasks = await generateTasks(this.ctx!);
```
evans: We could include a null check for `this.ctx` with an error that mentions the need to call `init` before any other function

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,5 +1,54 @@
...
+🔄 Validated your local schema against schema tag 'staging' on service 'engine'.
+🔢 Compared **0 schema changes** against operations seen over the **last day**.
+✅ Found **no breaking changes**.
+
+🔗 [View your service check details](https://engine-dev.apollographql.com/service/engine/checks?schemaTag=Detached%3A%20d664f715645c5f0bb5ad4f2260cd6cb8d19bbc68&schemaTagId=f9f68e7e-1b5f-4eab-a3da-1fd8cd681111&from=2019-03-26T22%3A25%3A12.887Z).
+"
+`);
+  });
+});
 // jest.mock("apollo-codegen-core/lib/localfs", () => {
```
evans: Since we're removing the tun off for summit, I recommend we comment/leave a TODO above the other tests that describes that these tests should be turned on, rather than deleted.

packages/apollo/src/commands/service/check.ts
justinanastos: Yes! That would be much more legible 👍

packages/apollo/src/Command.ts
justinanastos: That would be great because I didn't know that was necessary.

justinanastos: Also a small note: this line including `ctx!` wasn't from this PR. I updated it anyway :)

PR #1143, opened 2019-03-28
Add support for partial schemas in service:push

JakeDawkins:
TODO:


packages/apollo-language-server/src/providers/federation-info/endpoint.ts
```diff
@@ -0,0 +1,84 @@
...
+      );
+    }
+
+    this.info = data._service;
+  }
+
+  async resolveFederationInfo() {
+    if (!this.info) await this.getFederationInfo();
+
+    if (!this.info) throw new Error("No service info available");
```
trevor-scheer: I understand what's going on here, but there's a level of indirection around `this.info` that I think we could remove. If `getFederationInfo` just returned `data._service`, we could handle the `this.info` assignment here instead, and it'll be more obvious.

```js
if (!this.info) {
  this.info = ...
}
```

packages/apollo/src/commands/service/push.ts
```diff
@@ -28,6 +35,13 @@ export default class ServicePush extends ProjectCommand {
       {
         title: "Uploading service to Engine",
         task: async () => {
+          // handle partial schema uploading
+          if (flags.federated) {
+            const info = await (project as GraphQLServiceProject).resolveFederationInfo();
```
trevor-scheer: I think if we leverage the generic types for `this.runTasks` we _might_ be able to get rid of the cast here. See `ClientCheck` or `ServiceCheck` for an example.

packages/apollo-language-server/src/providers/federation-info/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    // }
+    if (config.service.endpoint) {
+      return new EndpointFederationInfoProvider(config.service.endpoint);
+    }
+    throw new Error(
+      "You must provide a service endpoint to use the federation info provider"
+    );
+  }
+
+  if (isClientConfig(config)) {
```
trevor-scheer: This is great 💯

packages/apollo-language-server/src/providers/federation-info/endpoint.ts
```diff
@@ -0,0 +1,84 @@
...
+  private async getFederationInfo() {
+    const { skipSSLValidation, url, headers } = this.config;
+    const options: HttpLink.Options = {
+      uri: url,
+      fetch,
+      ...(skipSSLValidation && { fetchOptions: { agent: new Agent() } })
+    };
+
+    const getFederationInfoQuery = `
+      query getFederationInfo {
```
trevor-scheer: Iiiiiinteresting...just thinking out loud. Seems we have a meta graphql problem here. Our type generation and VSCode are both going to try to validate this query against our schema 😂 

Might need an `// apollo-ignore` at some point.

JakeDawkins: Not in this case :) There's no `gql` so it's happily ignored!

PR #1148, opened 2019-03-29
Resolve nested dir creation codegen bug

trevor-scheer:
This resolves a bug in codegen where directory creation will fail / throw errors if we try to created a nested dir that doesn't exist. This case may only occur in a testing scenario, however this does allow us to re-enable our tests that exercise nested functionality.

TODO:

### DIFF
--- a/packages/apollo/__mocks__/apollo-cli-test.ts
+++ b/packages/apollo/__mocks__/apollo-cli-test.ts
@@ -4,6 +4,7 @@ import Nock from "@fancy-test/nock";
 import * as Test from "@oclif/test";
 export { expect } from "@oclif/test";
 import { mockConsole } from "heroku-cli-util";
+import { deleteFolderRecursive, makeNestedDir } from "../src/utils";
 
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "compile:clean": "tsc --build tsconfig.json --clean",
     "watch": "tsc --build tsconfig.json --watch",
     "test": "jest",
-    "test:ci": "npm run test -- --ci --maxWorkers=2",
+    "test:ci": "npm run test -- --ci --runInBand",
```
martijnwalraven: What's the reason we need to enable `--runInBand`? That can make running the tests slow, and more importantly is usually a sign tests aren't properly isolated.

trevor-scheer: @martijnwalraven I should've made more clear in the PR / commit messages that this is just exploratory. After initial work, tests were passing in circle and failing in azure, so I was trying a few things to try and identify the issue. I don't intend for any testing config changes to be final for this PR.

PR #1160, opened 2019-04-01
Add vue as supported language to extract gql

kestred:
Adds support for vue.  Uses the same extracter as `typescript`/`javacript`/ `___react`.

Just works by adding to list of supported languages.

### TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## Upcoming
 
+## `apollo-language-server`
+
 - apollo-language-server
   - Stop loadConfig from looking up the tree when a --config location is defined [#1059](https://github.com/apollographql/apollo-tooling/pull/1059)
   - Refactored/documented/tested loadConfig [#1059](https://github.com/apollographql/apollo-tooling/pull/1059)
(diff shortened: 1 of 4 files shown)


trevor-scheer:
@kestred this is great, thanks for adding this! If you can spare the effort, would you mind adding vue support here in the same way? If not, just let me know and I'll take care of it!

https://github.com/apollographql/apollo-tooling/blob/177f4274dfd386b9021f97b36406cbdc8c5dc753/packages/vscode-apollo/src/languageServerClient.ts#L42

Note the file watcher regex a few lines below as well.

kestred:
@trevor-scheer: added

kestred:
__Question/Aside:__ @trevor-scheer , is it the case that the vscode plugin (that package you had me update) will watch the files and automatically regenerate the types whenever they change?

trevor-scheer:
@kestred thanks for the quick turnaround. Merging this now!

Which types do you mean?

I can tell you we currently don't update any typescript types a la codegen.
The change I requested should just ensure that vscode notifies the language server of any changes to  files it should be interested in. The language server will then respond accordingly with hooks like this one:

https://github.com/apollographql/apollo-tooling/blob/177f4274dfd386b9021f97b36406cbdc8c5dc753/packages/apollo-language-server/src/server.ts#L127

PR #1178, opened 2019-04-08
UI-309 Update output for service:checks plan error

justinanastos:
Also fixes pluralization in the Listr status messages.

All errors thrown will cause Listr to look strange. I was unable to quickly diagnose where the error originates. It exists in prod also so I didn't try too hard to fix it.

TODO:

### DIFF
--- a/packages/apollo/src/commands/service/check.ts
+++ b/packages/apollo/src/commands/service/check.ts
@@ -13,13 +13,18 @@ import {
 import { ApolloConfig } from "apollo-language-server";
 import moment from "moment";
 import sortBy from "lodash.sortby";
+import stripANSI from "strip-ansi";
 
 function pluralize(
-  quantity: number | null,
+  quantity: string | number | null,
   singular: string,
   plural: string = `${singular}s`
 ) {
-  return `${quantity} ${quantity === 1 ? singular : plural}`;
+  // Strip ansi color from `quantity` if it's a string
+  const strippedQuantity =
+    typeof quantity === "string" ? parseInt(stripANSI(quantity), 0) : quantity;
+
+  return `${quantity} ${strippedQuantity === 1 ? singular : plural}`;
 }
 
 const formatChange = (change: Change) => {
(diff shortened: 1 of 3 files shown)


packages/apollo/src/commands/service/check.ts
```diff
@@ -13,13 +13,18 @@ import {
 import { ApolloConfig } from "apollo-language-server";
 import moment from "moment";
 import sortBy from "lodash.sortby";
+import stripANSI from "strip-ansi";
 
 function pluralize(
-  quantity: number | null,
+  quantity: string | number | null,
   singular: string,
   plural: string = `${singular}s`
```
evans: 😮 that's awesome!

packages/apollo/src/commands/service/check.ts
```diff
@@ -287,10 +287,6 @@ export default class ServiceCheck extends ProjectCommand {
                 // Save the output because we're going to use it even if we throw. `runTasks` won't return
                 // anything if we throw.
                 Object.assign(taskOutput, ctx);
-
-                task.title = `Validated local schema against tag ${chalk.blue(
```
evans: I think this is still valuable for the tense change

justinanastos: Good call.

justinanastos:
@evans 

> Also is there a way to test the screenshot below?

We _could_ refactor the functions that generate the messages out and then unit test them. I think we'd be better served spending a few hours to put together e2e tests. I think if we treat this like JavaScript that we can test all the output with relative ease!

justinanastos:
@evans I updated the tense change like you requested; that's the only change.

https://github.com/apollographql/apollo-tooling/pull/1178/files#diff-c61805bdfe85787383d4ae9927bd0774R291

justinanastos:
@evans 

>I agree with your assessment that end-2-end tests would be valuable. Do you think that it's worth tracking as a ticket in Jira?

I sure do

Issue #1180, opened 2019-04-10
Codegen is re-writting all generated types as many times as there are included graphql files


zapo:
**Intended outcome:**

Running `apollo client:codegen --includes="app/ui/**/*.gql" ...` should be writing generated types only once. 

**Actual outcome:**

In the following code extracted from codegen.ts command it seems like apollo codegen writes the types one time but also hooks on`project.onDiagnostics` to rewrite them as many time as there are "diagnostics".

```
  project.onDiagnostics(({ uri }) => {
    write();
  });

  const writtenFiles = write();
```

When building a new project in our current setup trying to debug very slow codegen I realized by console.log'ing `uri` in the above diagnostic hook that we end up having as many "diagnostics" as there are gql files included (not sure what "diagnostics" are used for in apollo projects). 
This ends up writing n+1 times all the generated types and the whole process takes up to 5min on our project instead of a few seconds (when commenting out the `project.onDiagnostics` hook or the `invalidate()` project calls that are responsible of filling diagnostics set).

**How to reproduce the issue:**

Here is the command we use (local schema file) and we aim to generate `__types__` folders alongside the queries:

```
apollo client:codegen --includes="app/ui/**/*.gql" --localSchemaFile=./schema.json --target=typescript --globalTypesFile=app/ui/__types__/graphql.ts __types__
```

I guess that the diagnostics hook is there for a reason and may be necessary in different setups (remote schema? apollo server?). If that's the case, could we provide a way to toggle off that hook in a setup like ours?

Thank you for your help!

**Versions**

This code is found in latest apollo version 2.8.2 and master.

zapo:
An other related weird issue with the onDiagnostics hook is that it persists SIGINT (Ctrl-C) and keeps writing things on disk until it went through all diagnostics.

zapo:
I realize onDiagnostics is used for watch mode and dropping it breaks it. I'll update the related PR accordingly.

eltonio450:
this is a huge problem for large projects heavily relying on fragments :(. Basically, the generation never ends, preventing it from using it in our CI. 

@zapo what  do you think we could do ? Is there a drawback to use your PR ?

Thank you :)

zapo:
Hey @eltonio450 we've been using my fork for about month at work without any issue. However note that I've only tested it in our setup which is basically running against a local schema.json. We run it in our CI once or in watch mode for devs.

[...]

PR #1183, opened 2019-04-12
Only write generated code once

zapo:
Fixes https://github.com/apollographql/apollo-tooling/issues/1180 by to only write generated code once by dropping the hook on `onDiagnostics`

This is probably not the right fix for this issue as I don't understand very well what diagnostics are meant for ... 

TODO:

- [x ] Rebase your changes on master so that they can be merged easily

### DIFF
--- a/packages/apollo/src/commands/client/codegen.ts
+++ b/packages/apollo/src/commands/client/codegen.ts
@@ -198,11 +198,6 @@ export default class Generate extends ClientCommand {
                 );
               };
 
-              // project.validationDidFinish(write);
-              project.onDiagnostics(({ uri }) => {
-                write();
-              });
-
               const writtenFiles = write();
 
               task.title = `Generating query files with '${inferredTarget}' target - wrote ${writtenFiles} files`;


smithcyr:
@zapo Is this fix still being worked on?

zapo:
@smithcyr Nope I was mostly waiting on comments to see if merge-able or not.

chrismohr:
I'd be pretty happy for this issue to be fixed. @trevor-scheer, @jbaxleyiii is there anything I can do to help get this reviewed and merged?

smoak:
@zapo It seems like option 1 might do the trick?

```js
if (watch) {
  project.onDiagnostics(..., => {
    write();
  });
}
```

I'm not sure what you mean by "would still have unnecessarily slow first codegen". The `write()` gets called the first time (we want this) and it's not "unnecessarily slow" for me with 145 graphql files to generate.

zapo:
@smoak It's been a long time but if I remember correctly from the time I debugged this `onDiagnostics` hook was running as many time as there are included query files on startup and then once per file change.
By guarding the onDiagnostics subscription to only occur on watch mode (option 1/) you're fixing the problem for single run but still have it in watch-mode.

I may be wrong and code might have changed since then though.

smoak:
@trevor-scheer, @jbaxleyiii any thoughts? this is currently affecting us and we are about to be forced to downgrade back to 1.x since our build times are now a lot longer.

trevor-scheer:
I don't want to be in the way on this but currently don't have the availability to diagnose and form an opinion on a solution. Tagging @JakeDawkins in case he does - this sounds like a very legitimate performance concern.

JakeDawkins:
I'll try my best to take a look at this in the next day or so! I'd really like to get this resolved

jhalborg:
Any news @JakeDawkins  ? :) This eats away at our daily output by quite a bit, it takes around 400 seconds to codegen for us

PR #1191, opened 2019-04-14
Add ability to skip codegen for inline documents with `// @codegen-ignore` (typescript only)

zapo:
This adds the ability to skip codegen for some inline documents with a `// @codegen-ignore` comment which leads document variable declaration in typescript only eg:

```
  // @codegen-ignore
  const document = gql`
    ...
  `;
```

This might be a bit weird/out of scope, but this is something our team would benefit from since we use apollo also for queries against a rest api (apollo-link-rest) for which we don't have schema, hence can't be codegen.

There is a much easy/obvious workaround which is to simply create a runtime alias for our tagName tagged template function (eg: gqlNoGen) so those don't get picked up by codegen but I wanted to try see if we could have a solution that doesn't involve changing runtime. 

To do so this updates parsing of template literals in typescript code to use typescript parser instead of regexp, which may be a bit more robust but requires adding typescript as runtime dependency. 

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Upcoming
 
 - `apollo`
-  - <First `apollo` related entry goes here>
+  - Add ability to ignore some inline documents in typescript codegen by prefixing with a `// @codegen-ignore` comment. [#1191](https://github.com/apollographql/apollo-tooling/pull/1191)
 - `apollo-codegen-core`
   - <First `apollo-codegen-core` related entry goes here>
 - `apollo-codegen-flow`
--- a/packages/apollo-language-server/package.json
+++ b/packages/apollo-language-server/package.json
@@ -41,6 +41,7 @@
     "moment": "^2.22.2",
     "recursive-readdir": "^2.2.2",
     "subscriptions-transport-ws": "^0.9.15",
+    "typescript": "^3.4.3",
     "vscode-languageserver": "^5.1.0",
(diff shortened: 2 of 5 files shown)


zionts:
@zapo thanks so much for taking the time and thought to put together this PR! This seems like a really valuable feature for the CLI to offer, and I'm curious to talk about how it can be supported in a more generic way, perhaps by leveraging the `includes` and `excludes` configuration options or by adding some kind of document-detection rules into the configuration.

Given this PR seems to have bitrotten some, what would you like to do to continue this work?

zapo:
Thanks for the feedback @zionts and sorry for late answer. I'm going to close this for now as I don't think this is the right way going forward. 

I believe includes/excludes is enough control for your tool to offer and the workaround of renaming the template literal processor can also be used for more control within a file (if not possible to isolate non-codegen queries in their own file for example)
I could also simply also hand-write a GQL schema to also generate types for my REST queries by passing it along my GQL schema in codegen params.

If really you want to support that feature:

I wonder if instead to parse the host language for comments we shouldn't instead parse for some comment in the GQL document itself (eg: `# @codegen-ignore`). 
That would avoid having to add a parser for each host language and leverage the existing graphql document parser in the tool.
If you want to leverage includes/excludes you would need a way to tag/name those documents since you may want to process only one of your queries for a given file. You can re-use query/mutation/fragment names but I feel that could become lengthy/unpractical:
eg:
```
excludes="src/module.ts:QueryName"
```

Issue #1167, opened 2019-04-04
[vscode plugin] autocomplete broken at v1.5.1


dontsave:
Just moved back to a graphql project and noticed that apollo no longer autocompletes queries in .graphql files or in gql (or custom) tags. There is no error in the console. The plugin loads fine and appears to load the schema fine from remote.

reverting to version `1.5.0` seems to solve the issue. The issue is also present in `1.5.2`

Apollo config:
```
module.exports = {
  client: {
    includes: [
      'src/**',
      'node_modules/gatsby-transformer-sharp/fragments.js',
      ],
      tagName: 'graphql',
      service: {
        name: 'ds.local',
        url: 'http://localhost:8800/___graphql',
       },
    },
};
```

VSCode info:

```
Version: 1.32.3 (user setup)
Commit: a3db5be9b5c6ba46bb7555ec5d60178ecc2eaae4
Date: 2019-03-14T23:43:35.476Z
Electron: 3.1.6
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Windows_NT x64 10.0.17763
```

ZenSoftware:
This seems to be a Windows specific issue.  Version 1.6.4 works fine on Linux, but not on Windows.  Reverting to 1.5.0 works on Windows.

jtmthf:
Also having this same issue but with macOS. Here is my config file

```js
module.exports = {
  client: {
    includes: ['src/**', 'node_modules/gatsby-*/**/*.js'],
    tagName: 'graphql',
    service: {
      name: 'gatsby'
      url: 'http://localhost:8000/___graphql',
    },
  },
};

```

dylanwulf:
I am also experiencing this issue, reverting back to 1.5.0 works for me. In my case, it seems that validation is also working incorrectly. It will validate the query once when VSCode starts up, but it will not re-validate when I change the query (unless I close and re-open VSCode)

JakeDawkins:
Autocomplete while typing should be working on windows with the latest release of #1213. It does look like there's an issue with the `ctrl+space` autocomplete though, so I'll leave this issue open until I can figure out what's going on there!

dylanwulf:
@JakeDawkins Looks like the problem with `ctrl+space` was fixed by #1361

JakeDawkins:
@dylanwulf great! I don't have the time today to verify this, so I'll close this issue and if anyone disagrees, let me know :)

[...]

PR #1213, opened 2019-04-27
Various fixes for Windows

JakeDawkins:
Fixes:
- [x] windows filepath support 
- [x] windows localSchemaFile support [spectrum discussion](https://spectrum.chat/apollo/apollo-tooling/vscode-does-not-seems-to-be-working-with-local-schema-file~7691b923-406c-4cfe-a554-77ad0b860068)
- [x] windows onHover not working
  - NOTE: it looks like it _does_ work, but only toward the end of the line. The offset it broken here somehow (maybe line endings)
- [x] windows autocomplete not working? #1167 

Summary: essentially there were 2 main issues
1. URIs weren't consistent. URI formats on windows varied between 3 or 4 different formats, and we weren't normalizing them to make sure we handled them all the same. I added a util where I saw breaks to handle them.
2. Line breaks on windows are 2 characters and we were only assuming 1 in checks inside the onHover handler, so some points on the line weren't being considered part of the token you were hovering over.

TODO:


berrylaw:
Which one ?

JakeDawkins:
@berrylaw still a work in progress. Right now, vscode on windows won't properly load because of an issue with matching paths. This fixes that (there's no GH issue for that one). There's also an onHover issue I need to look into before I can merge this

dylanwulf:
Is this going to include a fix for #1167 ? Or is that something different?

JakeDawkins:
@dylanwulf I believe these changes will fix #1167 but I can't confirm quite yet. I'm taking a look today and trying to wrap this up

JakeDawkins:
These changes were released in Apollo VS Code 1.6.9

dylanwulf:
@JakeDawkins Thank you! Just tried it out. The as-you-type autocomplete seems to be working correctly now, but ctrl+space is showing incorrect results. Should I make a new issue for that?

JakeDawkins:
Great! We're getting closer 🤞 

I'll just leave #1167 open for now and add a note until I can figure out what's going on there

PR #1241, opened 2019-05-07
Skip redundant modifiers

### LINKED ISSUE #1177: [Swift] Xcode 10.2 and use of namespace causes compiler warnings.
With Xcode 10.2, a warning is now emitted when a public class/struct exists within a public extension.

For example, using `apollo codegen:generate --target=swift --namespace="MyNamespace"` and the following generated code:

```swift
import Apollo

public extension MyNamespace {
  public final class MyQuery: GraphQLQuery {
  }

  public struct MyFragment: GraphQLFragment {
  }
}
```

Swift will emit the warning `'public' modifier is redundant for class declared in a public extension` for the `MyQuery` class and `MyFragment` struct.

The solution is to simply not include the `public` modifier for `MyQuery` when using a namespace.

This is not a language change, just a new warning. The fix should be compatible with all versions of Swift 4.

Note that it's quite common for Swift projects to treat warnings as errors, so this bug is a blocker for us.

**Versions**

Apollo 2.8.1

michaelnisi:
Resolves #1177. The goal is to prevent compiler warnings for redundant access-level modifiers when using `--namespace`.

It covers queries and fragments, according to basic testing at least, resolving the initial issue. Are there more types affected by this? @martijnwalraven 

### DIFF
--- a/packages/apollo-codegen-swift/src/__tests__/language.ts
+++ b/packages/apollo-codegen-swift/src/__tests__/language.ts
@@ -29,6 +29,26 @@ describe("Swift code generation: Basic language constructs", () => {
     `);
   });
 
+  it(`should generate a class declaration matching modifiers`, () => {
+    generator.classDeclaration(
+      { className: "Hero", modifiers: ["final"] },
+      () => {
+        generator.propertyDeclaration({
+          propertyName: "name",
+          typeName: "String"
+        });
+        generator.propertyDeclaration({ propertyName: "age", typeName: "Int" });
+      }
+    );
+
+    expect(generator.output).toBe(stripIndent`
+      final class Hero {
+        public var name: String
+        public var age: Int
+      }
+    `);
+  });
+
   it(`should generate a struct declaration`, () => {
     generator.structDeclaration({ structName: "Hero" }, () => {
       generator.propertyDeclaration({
@@ -40,6 +60,30 @@ describe("Swift code generation: Basic language constructs", () => {
 
     expect(generator.output).toBe(stripIndent`
       public struct Hero {
+        public var name: String
+        public var age: Int
+      }
+    `);
+  });
+
+  it(`should generate a namespaced fragment`, () => {
+    generator.structDeclaration(
+      {
+        structName: "Hero",
+        adoptedProtocols: ["GraphQLFragment"],
+        namespace: "StarWars"
+      },
+      () => {
+        generator.propertyDeclaration({
+          propertyName: "name",
+          typeName: "String"
+        });
+        generator.propertyDeclaration({ propertyName: "age", typeName: "Int" });
+      }
+    );
+
+    expect(generator.output).toBe(stripIndent`
+      struct Hero: GraphQLFragment {
         public var name: String
         public var age: Int
       }
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -168,10 +168,17 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
         throw new GraphQLError(`Unsupported operation type "${operationType}"`);
     }
 
+    const {
+      options: { namespace },
+      fragments
+    } = this.context;
+    const isRedundant = !!namespace;
+    const modifiers = isRedundant ? ["final"] : ["public", "final"];
+
     this.classDeclaration(
       {
         className,
-        modifiers: ["public", "final"],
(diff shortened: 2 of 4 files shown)


trevor-scheer:
Thanks for taking the time to put this together @michaelnisi!  The PR is greatly appreciated. I'm going to hand this off to @designatednerd for review since she's our new mobile architect 😄

packages/apollo-codegen-swift/src/__tests__/language.ts
```diff
@@ -46,6 +66,30 @@ describe("Swift code generation: Basic language constructs", () => {
...
+        generator.propertyDeclaration({
+          propertyName: "name",
+          typeName: "String"
+        });
+        generator.propertyDeclaration({ propertyName: "age", typeName: "Int" });
+      }
+    );
+
+    expect(generator.output).toBe(stripIndent`
+      struct Hero: GraphQLFragment {
```
designatednerd: I haven't been able to dig super-deep into what's happening here, but is the `adoptedProtocols` thing what shoves this into an extension rather than just into the big namespaced enum?

michaelnisi: Give me a minute to refresh my memory.

designatednerd: @michaelnisi Any recollection?

michaelnisi: First thing tomorrow morning.

michaelnisi:
Rebased and reviewed. These changes simply make the previously static `public` and `final` modifiers dynamic to prevent redundancy warnings #1177. 

The goal is to skip the `public` modifier in the namespaced case. It boils down to this.

For classes:

```ts
const isRedundant = !!namespace;
const modifiers = isRedundant ? ["final"] : ["public", "final"];
```

For structs: 

```ts
const isRedundant = adoptedProtocols.includes("GraphQLFragment") && !!namespace;
const modifier = isRedundant ? "" : "public ";
```

These changes cover queries and fragments.

packages/apollo-codegen-swift/src/__tests__/language.ts
michaelnisi: No, these changes are just for skipping the `public` modifier if we are in a namespace, see below.

michaelnisi: @designatednerd Lost interest? 🙂

designatednerd: Nope lost the changes in my flood of email 😄 will take another look

designatednerd:
@trevor-scheer I'll leave this to you to review the typescript on Monday but the swift stuff is ready to ship!

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -168,10 +168,17 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
...
+      fragments
+    } = this.context;
+    const isRedundant = !!namespace;
+    const modifiers = isRedundant ? ["final"] : ["public", "final"];
+
     this.classDeclaration(
       {
         className,
-        modifiers: ["public", "final"],
+        modifiers: modifiers,
```
trevor-scheer: Nit: since the variable name matches the key, the second `modifiers is redundant.
```suggestion
        modifiers,
```

michaelnisi: Thanks

michaelnisi:
Rebased and ready to go. 

However, there’s a new (unrelated?) failing CI check for Node.js 10:

```
Summary of all failing tests
 FAIL  packages/apollo/src/commands/client/__tests__/generate.test.ts (6.301s)
  ● client:codegen › extracts queries with a custom tagName provided in the config

    TypeError [ERR_INVALID_ARG_TYPE]: The "warning" argument must be one of type Error or string. Received type object
``` 

Let me know if this is a new requirement the Swift generator is neglecting and if I can help.

trevor-scheer:
@michaelnisi those are some flaky tests that we've been tolerating for quite some time now. All was well after a re-run. Thanks for the contribution! 🎉

gsabran:
This should only apply if the classes and structs are defined within an extension, which I think is equivalent to `outputIndividualFiles == true`. Else the code generated now looks like

```swift
public enum MyNameSpace {
    struct Hero: GraphQLFragment {
        public var name: String
        public var age: Int
    }
}
```

and the struct is internal.

Issue #1044, opened 2019-02-23
apollo client:codegen Option for different TypeScript enum output


rautenrieth-da:
Currently, the following GraphQL type
```
enum Color {
  RED
  BLUE
}
```

generates the following TypeScript type
```
export enum Color {
  RED = "RED",
  BLUE = "BLUE",
}
```

I would like to have the option to generate the following type instead:
```
export type Color = "RED" | "BLUE";
```

This definition provides the same safety, but is more efficient and works better with TypeScript structural typing.

Unrelated: consider generating `const enum` instead of `enum`.

mike-marcacci:
Related:
- #193 - [TS] Use Typescript string enum for enums
- #247 - Use TypeScript string enums

slorber:
+1, having an enum is annoying for me. I don't want to import the enum and to Enum.Value, I just want to use string literals, it's as safe as enums and works nicely

RyanCavanaugh:
#247 was a mistake IMO. String enums should only really be used when you want the underlying string value to be opaque, and that isn't a feature that GraphQL can provide. This would maybe be a good idea if you could produce *stable* key names, but the key names are going to change just as often as the string values themselves, so there's no point.

Jessidhia:
String enums still have a slight advantage in that you can find all usages of it. String constants actually can be searched by "find all usages", which was a nice surprise (:tada:), but typescript can't distinguish "intentional" uses from "coincidental" uses of a given string constant -- the string enums work to disambiguate them, especially if there are multiple different enums in your schema that happen to partially overlap in some of the values; it would be good to know _which_ enum is being referred to when the value appears in code.

I think this might be mitigated by the planned "unique types" feature, but I haven't seen if it can be made to work for this use case.

rohit-smpx:
> ```
> export type Color = "RED" | "BLUE";
> ```

These also work better if you have a js codebase and are using `checkjs`, there is no way (AFAIK) to import and use the typescript enum.

grantila:
I've fixed this in #1750 🙄

PR #1253, opened 2019-05-09
feat: support importing existing enums instead of generating them

ForbesLindesay:
This is useful because TypeScript does not see two identical enums in different files as being the same type.  If you already have a library that's expecting an enum from a different package, this lets you explicilty refer to that in place of the auto-generated enum.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - `apollo-codegen-swift`
   - <First `apollo-codegen-swift` related entry goes here>
 - `apollo-codegen-typescript`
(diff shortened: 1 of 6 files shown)


ForbesLindesay:
This should also resolve #1044

zionts:
Hi @ForbesLindesay — Thanks for putting the time into putting this PR together, and I'm sorry for the delay on getting back to you on it. Since so much has changed in the project, I'm going to close this out for now, but feel free to reopen if you're interested in dealing with the conflicts!

PR #1263, opened 2019-05-13
Update package.json

siddharth90:
TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
   - <First `apollo-codegen-typescript` related entry goes here>
 - `apollo-env`
   - Add new utils and predicates [#1251](https://github.com/apollographql/apollo-tooling/pull/1251)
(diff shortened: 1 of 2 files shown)


trevor-scheer:
Hey @siddharth90, thanks for the PR! This change is slightly more nuanced than just a version update in the package file, so I've opened this PR https://github.com/apollographql/apollo-tooling/pull/1267.

We had an issue in the past with an update that broke things for us, requiring us to pin the version indefinitely: https://github.com/apollographql/apollo-tooling/pull/961

But since you've raised my awareness to the official 3.x (stable) release, we can clean things up a bit on our side 😄

siddharth90:
Thanks for taking it into consideration and making the PR trevor! :)
Yes we were facing security issues as that particular version had a dependency on some other module version and it was flagged for security risk. Hence i looked up and saw they have a beta version that we could use now. I was sure there shouldve been a reason you had pinned the version to beta before. But thanks for taking this up now !

PR #1274, opened 2019-05-17
remove service:info command

evans:
This removes the `apollo service:info` command. Currently `apollo service:info` is not in the documentation. Additionally the information for a service is inside of the Apollo UI, so this command adds unnecessary noise.

Additionally, the command has only been called [once in the last
year](https://engine-staging.apollographql.com/service/engine/metrics?from=2018-04-17T23%3A25%3A00.000Z&query=5b4a5cb5ae6ed5af6c3b0422a31bda82cd12ca20&queryName=SchemaTagInfo&range=custom&schemaTag=prod&search=SchemaTagInfo&to=2019-05-17T23%3A25%3A00.000Z), so I suggest we make this a non-breaking change.


evans:
The command was used for internal platform usage according to @jbaxleyiii, so we can remove it! 🎉

PR #1286, opened 2019-05-20
Fix __typename addition for InlineFragments

trevor-scheer:
We need operation registration (client:push) `__typename`
adding to match how Apollo Client does this in order for query 
hashes to match.

In order to achieve parity here, I've added the `__typename` field
to InlineFragments as well. While the `__typename` on an interface
may be redundant (see new tests for example), it's also a safer
change for me to recommend than anything within Apollo Client.

TODO:

### DIFF
--- a/packages/apollo-language-server/src/utilities/graphql.ts
+++ b/packages/apollo-language-server/src/utilities/graphql.ts
@@ -136,8 +136,15 @@ export function withTypenameFieldAddedWhereNeeded(ast: ASTNode) {
       }
     },
     leave(node: ASTNode) {
-      if (!(node.kind === "Field" || node.kind === "FragmentDefinition"))
+      if (
+        !(
+          node.kind === "Field" ||
+          node.kind === "FragmentDefinition" ||
+          node.kind === "InlineFragment"
+        )
+      ) {
         return undefined;
+      }
       if (!node.selectionSet) return undefined;
 
       if (true) {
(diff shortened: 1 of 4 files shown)


shehaaz:
Thank you for the quick fix @trevor-scheer! Will this me approved and merged today? :)

packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -136,8 +136,15 @@ export function withTypenameFieldAddedWhereNeeded(ast: ASTNode) {
       }
     },
     leave(node: ASTNode) {
-      if (!(node.kind === "Field" || node.kind === "FragmentDefinition"))
+      if (
+        !(
+          node.kind === "Field" ||
+          node.kind === "FragmentDefinition" ||
+          node.kind === "InlineFragment"
```
abernix: I know you're just perpetuating the pattern that was already here, but thoughts on replacing these string literals (e.g. `"Field"`, `"FragmentDefinition"`, etc.) with the `graphql-js` "`Kind`" constants, for example, [`Kind.FRAGMENT_DEFINITION`](https://github.com/graphql/graphql-js/blob/8c96dc8276f2de27b8af9ffbd71a4597d483523f/src/language/kinds.js#L28).

trevor-scheer: Definitely. Good call!

trevor-scheer: Resolved in 6e61408

PR #1288, opened 2019-05-21
[language server] Allow configuration of validation rules.

### LINKED ISSUE #1056: Add support for custom validation rules in apollo-language-server
Inspired by #800 

> NOTE: I **won't** be personally building this anytime soon. I WILL, however, gladly review any pull requests that adds this feature 😄 

Right now, the `apollo-language-server` runs a set of validation rules against operations (great, right??). 

In certain circumstances (like if you're running a relay project), you may want to disable existing validations or add new ones (like requiring an `id` on every object type for caches).

Since the `apollo.config.js` is an executable file, not just a static configuration, we can actually add these validations to it to run against client projects!

Here's my proposed API:

```
// apollo.config.js
const { defaultValidationRules } = require('apollo/lib/defaultValidationRules')
const relayValidationRules = require('my-custom-validation-rules')

module.exports = {
  client: {
    validationRules: [
		relayValidationRules,
		...defaultValidationRules.filter(removeBitsThatMessUpRelay)
	]
  }
}
```

This way, you'd have _full_ control (for better or worse 😉) over validation rules, rather than just enabling/disabling rules from graphql-js

---

This would require changes to: 
- apollo-language-server
	- add validation overrides to the config/validation
- apollo
	- export the default validation rules for easy import

alloy:
Closes #1056

(I didn’t find any test coverage for the areas I touched, but am pretty sure my additions are backwards compatible.)

### Before

vscode-apollo before

### Config

vscode-apollo config

### After

vscode-apollo after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   - <First `apollo-graphql` related entry goes here>
 - `apollo-language-server`
   - update stats window types [#1292](https://github.com/apollographql/apollo-tooling/pull/1292)
+  - Allow configuration of validation rules [#1288](https://github.com/apollographql/apollo-tooling/pull/1288)
 - `apollo-tools`
   - <First `apollo-tools` related entry goes here>
 - `vscode-apollo`
--- a/packages/apollo-codegen-core/src/loading.ts
+++ b/packages/apollo-codegen-core/src/loading.ts
@@ -87,7 +87,7 @@ function extractDocumentsWithAST(
         // This currently ignores the anti-pattern of including an interpolated
         // string as anything other than a fragment definition, for example a
         // literal(these cases could be covered during the replacement of
-        // literals in the signature calcluation)
+        // literals in the signature calculation)
         finished.push(
           (path.value.quasi.quasis as Array<{
             value: { cooked: string; raw: string };
--- a/packages/apollo-language-server/src/config/config.ts
+++ b/packages/apollo-language-server/src/config/config.ts
@@ -72,13 +72,28 @@ export interface ClientConfigFormat extends ConfigBase {
   /**
    * Rules that will be applied when validating GraphQL documents.
    *
-   * If you wish to modify the default list of validation rules, import them from the apollo package:
+   * If you wish to modify the default list of validation rules, import them from the apollo package and
+   * assign your custom list:
    *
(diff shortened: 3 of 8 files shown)


alloy:
I've seen that CI failure before and it went away after a re-run by fixing a spelling error in a comment and force pushing. I can do the same again, but perhaps it’s better to highlight this gremlin?

alloy:
Oops, forgot to mention you before, @JakeDawkins.

package-lock.json
```diff
@@ -2327,17 +2327,15 @@
       "dependencies": {
         "@oclif/config": {
           "version": "1.13.0",
-          "resolved": "https://registry.npmjs.org/@oclif/config/-/config-1.13.0.tgz",
```
janicduplessis: Should probably revert the changes here, I guess its caused by using a diff npm version

alloy: Good call, thanks!

alloy:
I added the ability to specify a function that can be used to filter the default validation rules. The reason for this is that when using the extension on non-javascript projects, there is no good way to import the default validation rules from the Apollo package without adding a `package.json` file, the `apollo` dependency, and installing the node modules. This feels like overkill for such projects.

In our case, we have Ruby applications that we want to use the extension on, but we want to filter out the `NoAnonymousQueries` validation.

trevor-scheer:
This LGTM. There are a couple tasks I'd like to spin out of this.
* Export the rule set at the top level of the language server (and maybe `apollo`)
* Document the new config option (see referenced issue on this PR)

If you're up for both or either, just let me know - otherwise I'll take care of them. Thanks again!

trevor-scheer:
@alloy FYI, I cut a release but can't get the extension working (unrelated to your changes) so I'll need to get a fix together before I can get that packaged up and posted on the marketplace. Sorry about the delay! I'm aiming for tomorrow, but I'll keep you looped in either way.

trevor-scheer:
For context on the issue, see: https://github.com/apollographql/apollo-tooling/pull/1306
Extension is now released and up on the marketplace `vscode-apollo@1.7.1` 🎉

Thanks again!

alloy:
Awesome, thanks!! 🙏

Yeah I can do those other tasks 👍

Issue #1272, opened 2019-05-16
Improve error messaging around client:{extract|push}


trevor-scheer:
When we parse an invalid query during query extraction, we throw it out and silently succeed. We should inform the user that we found what we think is a query (in gql tag), but we couldn't parse it or it's invalid. We should also fail the command.

mgeraci:
This sounds like a great improvement! I ran into a problem related to this where I added a query that failed (JS string interpolation in a query is invalid, it turns out) but my CI flow succeeded. This meant that new code was deployed with queries that weren't in the Operation Registry. Luckily, I was updating code behind a feature flag, but this could have led to forbidden operation errors in production! So having the command fail on an invalid query would be a huge improvement.

PR #1290, opened 2019-05-22
apollo client:push update output to include more details

evans:
This updates the operation registration command to use the new resolver
that returns more information, namely added operations and invalidation
messages associated with the operation. The `apollo client:push` command
then outputs this information

Happy path:
image

Sad path:
image

TODO:

### DIFF
--- a/packages/apollo/src/commands/client/push.ts
+++ b/packages/apollo/src/commands/client/push.ts
@@ -9,6 +9,11 @@ import {
 } from "apollo-graphql";
 import { pluralize } from "../../utils";
 import chalk from "chalk";
+import {
+  GraphQLClientProject,
+  ApolloConfig,
+  graphqlTypes
+} from "apollo-language-server";
 
 export default class ClientPush extends ClientCommand {
   static description =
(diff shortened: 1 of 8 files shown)


trevor-scheer:
Fixes #1272? Would be great to get @mgeraci 's feedback on this since he suggested this idea to me oh-so-recently.

packages/apollo/src/commands/client/push.ts
```diff
@@ -1,70 +1,195 @@
...
+import URI from "vscode-uri";
+import { getOperationManifestFromProject } from "../../utils/getOperationManifestFromProject";
 import {
-  getOperationManifestFromProject,
-  ManifestEntry
-} from "../../utils/getOperationManifestFromProject";
-import { ClientIdentity } from "apollo-language-server";
+  operationHash,
+  defaultOperationRegistrySignature
+} from "../../../../apollo-graphql/lib";
```
trevor-scheer:
```suggestion
} from "apollo-graphql";
```

evans:
@trevor-scheer That's a great idea! I'll see how to work that into the PR

packages/apollo/src/commands/client/push.ts
```diff
@@ -1,70 +1,195 @@
...
+                registrationSuccess
+              } = await project.engine.registerOperations(variables);
 
-    this.log(
-      `Successfully pushed ${operations.length} operations from the ${
-        clientIdentity.name
-      } client to the ${serviceName} service in Engine`
-    );
+              if (!registrationSuccess) {
+                if (invalidOperations) {
```
trevor-scheer: Is it possible for `!registrationSuccess && !invalidOperations`? I only ask in case it's a possible scenario, and if it has any semantics that we can communicate.

I don't imagine so, but I'd suggest that we throw or log something if that's the case.

packages/apollo/src/commands/client/push.ts
```diff
@@ -1,70 +1,195 @@
...
+                      );
+                  });
+                  task.title = `Failed to register operations, due to ${
+                    invalidOperations.length
+                  } invalid operations`;
+                  throw Error(invalidOperationsErrorMessage);
+                }
+              } else {
+                if (newOperations && newOperations.length) {
+                  task.title = `Added ${
```
trevor-scheer: Small nit: `Added 6 to Apollo Platform` reads a bit awkward. I'll offer some suggestions, but please feel free to disregard / use your discretion.

`Successfully {registered|pushed|added} 6 operations (optional: to the operation {registry|safelist})`

packages/apollo/src/commands/client/push.ts
evans: Thanks!

packages/apollo/src/commands/client/push.ts
evans: Right now it shouldn't be possible, though maybe in the future. I'll add an error message!

packages/apollo/src/commands/client/push.ts
evans: I like it!

mgeraci:
Sorry I didn't get in here before the merge, but this looks like a great set of improvements! Did it end up hitting #1272 — there are tons of great messaging improvements here, but I wasn't sure if this change also errors out if it finds an invalid operation. Either way, great work!

PR #1291, opened 2019-05-22
Pin graphql to the ~14.2.x range

trevor-scheer:
An incoming graphql release (14.3.1) is going to break at least
`apollo client:check` as it did in the past with `14.2.0` (confirmed this
manually with graphql's master branch). This dependency restriction
will prevent this for now, without us needing to find the root of
the issue. This is preventative until we have the resources
to investigate and resolve the actual cause of the problem.

For more details and full context, see:
https://github.com/graphql/graphql-js/commit/183ff32bee4bc23eb23657f79f414c2400ecb06a#r32971387

cc @abernix @martijnwalraven

TODO:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2313,7 +2313,7 @@
         "git-parse": "1.0.3",
         "git-rev-sync": "1.12.0",
         "glob": "7.1.4",
-        "graphql": "^14.2.1",
+        "graphql": "~14.2.1",
         "graphql-tag": "2.10.1",
         "heroku-cli-util": "8.0.11",
         "listr": "0.14.3",
@@ -2485,7 +2485,7 @@
         "cosmiconfig": "^5.0.6",
         "dotenv": "^8.0.0",
         "glob": "^7.1.3",
-        "graphql": "^14.2.1",
+        "graphql": "~14.2.1",
         "graphql-tag": "^2.10.1",
         "lodash.debounce": "^4.0.8",
         "lodash.merge": "^4.6.1",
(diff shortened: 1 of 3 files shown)


abernix:
I'm slightly worried that this will cause duplicated `graphql` versions (which is a problem because of `graphql`) for those that have already upgraded to 14.3.0, but we should be defensive for the time-being.  At least in the case that I've just described, there would  be very clear failures pretty quickly in the application, rather than less-likely-to-be-noticed surprise differences due to unexpected changes in `graphql`.

ryami333:
@trevor-scheer is there anything I can do to help move this along? Is this a result of a breaking change inadvertantly introduced by the `graphql-js` maintainers, and if so: are they aware of it? I could not find any mention of this in their recent issues so I don't suppose there's any upstream fix coming, which implies that the fault lies here instead.

trevor-scheer:
@ryami333, you're welcome to investigate and I'm happy to share what information I have with you. There's some history with trouble with recent patches of `graphql` on our end. It's a lot to explain, but most of the conversation happened here:
https://github.com/graphql/graphql-js/commit/183ff32bee4bc23eb23657f79f414c2400ecb06a#r32971387

`14.2.0` breaks the CLI, but the resulting fix which works for the CLI (`14.2.1`) was a breaking change in and of itself. This was pointed out after `14.3.0`, which resulted in the `14.3.1` patch.

Prior to the release of `14.3.1`, I set the version to `~14.2.1` which currently works for the CLI.

If you're interested in proposing a fix and have the time to investigate, you can reproduce by changing the graphql version to latest and running `apollo client:check`
See this comment: https://github.com/graphql/graphql-js/commit/183ff32bee4bc23eb23657f79f414c2400ecb06a#r33630399

I haven't had time to investigate the issue, but I'd be happy to help anyone who's interested in doing so. Unfortunately, this path was chosen as a way to minimize damage (`^14.2.1` would resolve to `14.3.1` and break the CLI), but you're correct in noting that the problem isn't solved.

PR #1294, opened 2019-05-23
client:push add check for parse failures

evans:
Currently the client project returns operation that parse correctly.
This strategy swallows parse failures. This updates client:push to
surface the unparseable operations.

Ideally this would be a more comprehensive solution that solves the
issues for extract as well as check, so some feedback on a better
solution is highly encouraged

TODO:


zionts:
@evans this seems like a really useful feature, but it seems like this branch has bitrotten :/ do you think that you'll pick this work back up, or should we create an issue to come back to it and close this PR out?

Issue #1298, opened 2019-05-24
VS Code extension - Syntax Error with multi-line placeholders in gql


henryqdineen:
Our prettier config's print-width setting is breaking some placeholders into multiple lines in our `gql` tagged template literals and causing errors in the VS Code plugin.

**Intended outcome:**

No error. Correctly parse the JS.

**Actual outcome:**

`Syntax Error: Unexpected $ GraphQL: Syntax [73, 3]`

**How to reproduce the issue:**
You need to have a placeholder that breaks onto a new line.
```js
const query = gql`
  {
    hero {
      ...Hero_character
    }
  }

  ${Hero.fragments
    .character}
`;
```
**Versions**
v1.6.9 of the VS Code extension

PR #1299, opened 2019-05-24
language server: allow template literal placeholders that span multiple rows.

henryqdineen:
Fixes: #1298

The fix is just to update the regex to handle newlines in a template literal placeholder. I think this code would be more robust if it used ASTs to extract the graphql documents but I'm assuming there was a reason to use regexes instead.

TODO:


trevor-scheer:
@henryqdineen, very much appreciated - and thanks for the tests! I pushed a little cleanup to the branch to save you some minor nits. Going to merge as soon as the tests pass 😄

PR #1308, opened 2019-05-30
AP-490: Finalize federated service check and add integration tests

justinanastos:
**Review this PR commit-by-commit**

This PR:

* Adds integration tests for `service:check`
    * Input: CLI commands being run as a user would run them
    * Background: Mock network calls with [`nock`](https://github.com/nock/nock)
    * Measure: Compare expected CLI output. This includes titles and status being modified as commands are run
* Add federated service check functionality
    * Replace `--federated` flag in place of implying federated behavior with `--serviceName`
    * Also adds integration test
    * `--markdown` and `--json` are not supported; but that's coming soon.

_When creating new responses, use nock's recorder functionality (https://github.com/nock/nock#recording). We made the request matching only check the query name so the code would be less brittle._

TODO:

- [x] Update CHANGELOG.md with your change (include reference to issue & this PR)
- [x] Make sure all of the significant new logic is covered by tests \o/
- [x] Document `nock` recorder functionality :)


packages/apollo/src/commands/service/check.ts
```diff
@@ -250,115 +257,201 @@ export default class ServiceCheck extends ProjectCommand {
   async run() {
     // @ts-ignore we're goign to populate `taskOutput` later
     const taskOutput: TasksOutput = {};
-
+    // nock.recorder.rec();
```
justinanastos: Remove commented out code.

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,22 +1,328 @@
...
+function mockNonFederatedFailure() {
+  mockIntrospectionQuery();
+
+  nock("https://engine-staging-graphql.apollographql.com:443", {
+    encodedQueryParams: true
+  })
+    .post("/api/graphql", () => true)
+    .reply(200, {
+      data: {
+        service: {
```
trevor-scheer: Can we use our generated types to type this data? That way we know to update it and test the right thing when the schema / types change!

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,22 +1,328 @@
...
+        it("vanilla", async () => {
+          captureApplicationOutput();
+          mockNonFederatedSuccess();
+          expect.assertions(2);
+
+          await expect(
+            ServiceCheck.run([...cliKeyParameter])
+          ).resolves.not.toThrow();
+
+          // Inline snapshots don't work here due to https://github.com/facebook/jest/issues/6744.
```
trevor-scheer: This issue is left `needs repro` and hasn't seen much activity. Can we give them directions on how to reproduce within our repo after this lands?

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -238,6 +380,130 @@ describe("service:check", () => {
...
+            ])
+          ).rejects.toThrow(
+            federatedServiceCompositionUnsuccessfulErrorMessage
+          );
+
+          // Inline snapshots don't work here due to https://github.com/facebook/jest/issues/6744.
+          expect(uncaptureApplicationOutput()).toMatchSnapshot();
+        });
+
+        it.skip("--markdown", async () => {
```
trevor-scheer: Should this be skipped?

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -238,6 +380,130 @@ describe("service:check", () => {
...
+            ])
+          ).rejects.toThrow(
+            federatedServiceCompositionUnsuccessfulErrorMessage
+          );
+
+          // Inline snapshots don't work here due to https://github.com/facebook/jest/issues/6744.
+          expect(uncaptureApplicationOutput()).toMatchSnapshot();
+        });
+
+        it.skip("--json", async () => {
```
trevor-scheer: Should this be skipped?

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -238,6 +380,130 @@ describe("service:check", () => {
...
+              "--serviceName=accounts",
+              `--endpoint=${localURL}/graphql`
+            ])
+          ).resolves.not.toThrow();
+
+          // Inline snapshots don't work here due to https://github.com/facebook/jest/issues/6744.
+          expect(uncaptureApplicationOutput()).toMatchSnapshot();
+        });
+
+        it.skip("--markdown", async () => {
```
trevor-scheer: Should this be skipped?

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -238,6 +380,130 @@ describe("service:check", () => {
...
+              `--endpoint=${localURL}/graphql`,
+              "--markdown"
+            ])
+          ).resolves.not.toThrow();
+
+          // Inline snapshots don't work here due to https://github.com/facebook/jest/issues/6744.
+          expect(uncaptureApplicationOutput()).toMatchSnapshot();
+        });
+
+        it.skip("--json", async () => {
```
trevor-scheer: Should this be skipped?

packages/apollo/src/commands/service/check.ts
```diff
@@ -271,6 +273,7 @@ export default class ServiceCheck extends ProjectCommand {
 
           const configName = config.name;
           const tag = flags.tag || config.tag || "current";
+          const serviceName = flags.serviceName;
```
trevor-scheer: I think `config.name` (above) and `flags.serviceName` are the same, but from different locations. I think we should resolve similarly to how we resolve tag, with the flag overriding the value from config if it's present.

trevor-scheer: I may be mistaken here, as there might be some implicit federation logic that differentiates this. If that's the case, could you please add a comment explaining what's going on here?

packages/apollo/src/commands/service/check.ts
```diff
@@ -285,7 +288,13 @@ export default class ServiceCheck extends ProjectCommand {
                 task.output = "Resolving schema";
                 taskOutput.shouldOutputJson = flags.json;
 
-                if (flags.federated) {
+                // If we're `federated`, then run composition validation. When we're using composition
+                // validation we'll receive a schema has that represents the composed schema.
+                if (serviceName) {
```
trevor-scheer: I think removing `flags.federated` here is a mistake, unless we're trying to remove that flag. I'll point back to the comment about this possibly being a bit too implicit. Does the presence of `serviceName` -> we're federated?

packages/apollo/src/commands/service/check.ts
```diff
@@ -296,79 +305,129 @@ export default class ServiceCheck extends ProjectCommand {
...
-                    throw new Error(`Federated service could not be composed due to the following errors:
-  ${errors && errors.map(err => (err && err.message) || "").join("\n")}
-                  `);
                   }
-                } else {
-                  schema = await project.resolveSchema({ tag: config.tag });
 
-                  await gitInfo(this.log);
+                  if (errors.length > 0) {
+                    const decodedErrors = errors.filter(Boolean).map(error => {
```
trevor-scheer: `isNotNullOrUndefined` from `apollo-env` solves this nicely for us at the type level 👍

packages/apollo/src/commands/service/check.ts
```diff
@@ -296,79 +305,129 @@ export default class ServiceCheck extends ProjectCommand {
...
+                      // We `filter` by `Boolean` above; this will never happen and is just here to appease
+                      // TypeScript.
+                      if (!error) {
+                        return {};
+                      }
+
+                      const match = error.message.match(
```
trevor-scheer: Maybe we should be able to request these capture groups in the shape we want from the backend, rather than parse the error message. I think this needs to be implemented on our end at the composition level in order for them to do that. This would be a great feature request to bubble up through infra into federation / composition!

packages/apollo/src/commands/service/check.ts
```diff
@@ -259,6 +263,10 @@ export default class ServiceCheck extends ProjectCommand {
     try {
       await this.runTasks<TasksOutput>(
         ({ config, flags, project }) => {
+          if (!isServiceProject(project)) {
```
trevor-scheer: I love this, but can we make the error message a bit easier to understand / more actionable from the user's perspective? They may not know what they've gotten themselves into. We should let them know we've found a client project, but need a service project. We've got a bitly link scattered throughout the repo for pointing users to the apollo config docs that would be helpful here: https://bit.ly/2ByILPj

packages/apollo/src/commands/service/check.ts
```diff
@@ -20,6 +21,7 @@ import {
 import moment from "moment";
 import sortBy from "lodash.sortby";
 import cli from "cli-ux";
+import { continueStatement } from "babel-types";
```
trevor-scheer: Accidental import (which you've probably already removed somewhere)

packages/apollo/src/commands/service/check.ts
```diff
@@ -273,110 +271,142 @@ export default class ServiceCheck extends ProjectCommand {
 
           const configName = config.name;
           const tag = flags.tag || config.tag || "current";
-          const serviceName = flags.serviceName;
+          const serviceName: string | undefined = flags.serviceName;
```
trevor-scheer: Is this type not inferrable?

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,22 +1,328 @@
...
+ * with a fake API key.
+ */
+const fakeApiKey = "service:engine:9YC5AooMa2yO11eFlZat11";
+
+/**
+ * Engine API key we're using.
+ *
+ * We should use the `fakeApiKey` whenever possible to ensure we can't accidentally access real APIs.
+ */
+const apiKey = process.env.ENGINE_API_KEY || fakeApiKey;
```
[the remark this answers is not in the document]
justinanastos: Totally right and totally intentional. If we want to refresh the mock data we'll have to run this against the real APIs using a real `ENGINE_API_KEY`.

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,22 +1,328 @@
...
+
+/**
+ * An array that we'll spread into all CLI commands to pass the engine api key.
+ */
+const cliKeyParameter: string[] = apiKey ? [`--key=${apiKey}`] : [];
```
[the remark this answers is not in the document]
justinanastos: Totally right. This is a mistake from when I was trying to make this work without an API key. Great eyes!

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,22 +1,328 @@
...
+
+/**
+ * The original `console.log` being mocked.
+ *
+ * We save it so we can restore it after a test.
+ */
+let mockedConsoleLogOriginal: Console["log"] | null = null;
```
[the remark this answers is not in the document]
justinanastos: We could, but is that really what we want? I imagine we want to capture it when a test starts and release it when it the test ends. If we stored the "original", we'd be storing the value of `console.log` when this file was evaluated and then we'd have to restore it somewhere else.

I'm open to making this easier to understand, can you help me understand why that would be better?

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,22 +1,328 @@
...
+  stdout.start();
+}
+
+/**
+ * Reverse mocking of `console.log` and `stdout.write`. If they weren't mocked to begin with, this will do
+ * nothing and return null.
+ */
+function uncaptureApplicationOutput(): string | null {
+  // These will be `null` if we haven't mocked `console.log`.
+  if (!mockedConsoleLogOriginal || !mockedConsoleLogValues) {
```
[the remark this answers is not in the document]
justinanastos: Yes, I try to un-capture before and after each test incase something goes wrong.

packages/apollo/src/commands/service/__tests__/check.test.ts
justinanastos: I can _try_, but I'm not certain _where_ in the `ServiceCheck.run` the issue is coming from. I want to investigate that there's not some weird `fork` or multi-process logic going on.

packages/apollo/src/commands/service/__tests__/check.test.ts
justinanastos: Yes it should, I'll leave a comment as to why.

packages/apollo/src/commands/service/check.ts
justinanastos: `config.name` is the graph name (`engine`) and `serviceName` is the implementing service name (`accounts`). Great place to use some comments and more descriptive naming.

packages/apollo/src/commands/service/check.ts
justinanastos: Matt explicitly said he wants to use `--serviceName` and not `--federated`. Yes, `serviceName` means we're federated.

packages/apollo/src/commands/service/check.ts
justinanastos: Totally right! @JakeDawkins brought that up too. I'll make a ticket so I don't forget.

justinanastos: https://apollographql.atlassian.net/browse/AP-521

packages/apollo/src/commands/service/check.ts
justinanastos: 🤦‍♂

packages/apollo/src/commands/service/check.ts
justinanastos: Great point. I imagined this only being for development (to catch myself); I guess a user could get here too and needs a way to get out.

packages/apollo/src/commands/service/check.ts
justinanastos: Correct; flags is not typed.

packages/apollo/src/commands/service/__tests__/check.test.ts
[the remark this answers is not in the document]
trevor-scheer: Right, but if I'm running tests locally would I normally want to use `fakeApiKey` (per the comment)? I think if I ran this locally, _my_ api key would get loaded up just because I have the tooling repo configured for engine.

I think you're saying that's intentional, and if so then 👍. I was basing my comment on your comment about using `fakeApiKey` whenever possible.

packages/apollo/src/commands/service/__tests__/check.test.ts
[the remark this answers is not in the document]
trevor-scheer: I think you have a better understanding of some nuance here which I did not. My suggestion may have over-simplified the goal, and it was just or the sake of a bit less code / null checks. I don't think it's difficult to understand, so it sounds like leaving as-is is the right choice.

packages/apollo/src/commands/service/check.ts
trevor-scheer: Ok. We should document this in the `serviceName` flag `description` up above.

We should also remove the `federated` flag if it does nothing now. It's _sort of_ a breaking change, but since `serviceName` has been a requirement as well, nothing's actually breaking, we just don't _need_ it anymore.

packages/apollo/src/commands/service/check.ts
trevor-scheer: 🤗

packages/apollo/src/commands/service/check.ts
```diff
@@ -253,112 +260,198 @@ export default class ServiceCheck extends ProjectCommand {
...
+                      /^\[([^\[]+)\]\s+(\S+)\ ->\ (.+)/
+                    );
+
+                    if (!match) {
+                      // If we can't match the errors, that means they're in a format we don't recognize.
+                      // Report the entire string as the user will see the raw message.
+                      return { message: error.message };
                     }
+
+                    const [, service, field, message] = match;
```
evans: I'm not sure what the first value in the array is. Can we add a comment to describe what we're ignoring?

packages/apollo/src/commands/service/check.ts
```diff
@@ -253,112 +260,198 @@ export default class ServiceCheck extends ProjectCommand {
...
+                  // because we need to short-circuit the rest of the tasks in the service check.
+                  throw new Error(
+                    federatedServiceCompositionUnsuccessfulErrorMessage
+                  );
+                }
+              }
+            },
+            {
+              title: `Validating ${
+                serviceName ? "composed" : "local"
```
evans: I realized that local might not be accurate depending on the source/endpoint. Would it be reasonable to remove it?

packages/apollo/src/commands/service/check.ts
```diff
@@ -253,112 +260,198 @@ export default class ServiceCheck extends ProjectCommand {
...
-                    graphVariant: tag,
-                    implementingServiceName: flags.serviceName || info.name,
-                    partialSchema: {
-                      sdl: info.sdl
+                const info = await project.resolveFederationInfo();
+                if (!info.sdl) {
+                  throw new Error("No SDL found for federated service");
+                }
+
+                task.output = `Attempting to compose graph with ${chalk.blue(
```
evans: It would be great to have some diagnostic information about how many other services are included in the composition

packages/apollo/src/commands/service/check.ts
```diff
@@ -253,112 +260,198 @@ export default class ServiceCheck extends ProjectCommand {
...
+                  ...schemaCheckSchemaVariables
+                };
+
+                const { schema: _, ...restVariables } = variables;
+                this.debug("Variables sent to Engine:");
+                this.debug(restVariables);
+                if (schema) {
                   this.debug("SDL of introspection sent to Engine:");
                   this.debug(printSchema(schema));
+                }
```
evans: It would be helpful to log the schemaHash here.

Since it's debug mode, I'd also think that the command to introspect/download the full schema from manual inspection would complete the experience

packages/apollo/src/commands/service/__tests__/check.test.ts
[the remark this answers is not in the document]
evans: I'm not sure how we can refresh the mock data from this comment. Can we add something here and possibly the readme to explain how that process might work?

packages/apollo/src/commands/service/check.ts
```diff
@@ -253,112 +260,198 @@ export default class ServiceCheck extends ProjectCommand {
...
+
+                task.output = "Validating schema";
+
+                const variables: CheckSchemaVariables = {
+                  id: configName,
+                  tag: flags.tag,
+                  gitContext: await gitInfo(this.log),
+                  frontend: flags.frontend || config.engine.frontend,
+                  ...(historicParameters && { historicParameters }),
+                  // `variables` will either contain `schemaHash` or `schema`, not both.
```
evans: Is this an invariant we can enforce/pretty log in the client side?

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,22 +1,328 @@
...
+import ServiceCheck, {
   formatHumanReadable,
   formatMarkdown,
   formatTimePeriod
 } from "../check";
 import checkSchemaResult from "../../../../__fixtures__/check-schema-result";
 import { ChangeSeverity } from "apollo-language-server/lib/graphqlTypes";
 import chalk from "chalk";
+import nock = require("nock");
+import { stdout, stderr } from "stdout-stderr";
```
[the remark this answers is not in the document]
justinanastos: Great point. This would be something I'd love to lint for with `eslint-plugin-import`.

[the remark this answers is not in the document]
justinanastos: How could any mortal keep track of that 👀

packages/apollo/src/commands/service/__tests__/check.test.ts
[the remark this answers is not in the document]
justinanastos: Great suggestion :) And you're right Trevor; I'll remove the real API key and then document when to change that.

packages/apollo/src/commands/service/__tests__/check.test.ts
```diff
@@ -1,22 +1,328 @@
...
+}
+
+/**
+ * Convert a schema SDL to an introspection query result.
+ *
+ * @see https://blog.apollographql.com/three-ways-to-represent-your-graphql-schema-a41f4175100d
+ *
+ * @param schemaSdl string Schema in SDL form
+ */
+function sdlToIntrospectionQueryResult(schemaSdl: string) {
```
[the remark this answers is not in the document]
justinanastos: I messed with `introspectionFromSchema` and wasn't able to find out how to use it here. Can you help me understand what you mean?

packages/apollo/src/commands/service/__tests__/check.test.ts
justinanastos: Some of it yes, some of it no. I started this and it ended up being a pain because we have to add all the `__typename`s and we can't use string literals for enum values. If it's ok, I'd rather _not_ do this for now 🤷‍♂

packages/apollo/src/commands/service/check.ts
justinanastos: I want to point out that this is a breaking change, but the current state of the CLI doesn't allow anyone to run federated service checks at all, so it's not going to break anyone :)

packages/apollo/src/commands/service/check.ts
justinanastos: OMG why can `Boolean` _not_ do this 🤕

packages/apollo/src/commands/service/__tests__/check.test.ts
justinanastos: Yes, same as https://github.com/apollographql/apollo-tooling/pull/1308#discussion_r291298332:

>Yes it should because `--markdown` isn't implemented yet. I'll leave a comment here and at the other skip sites with the ticket that'll implementing this.

packages/apollo/src/commands/service/__tests__/check.test.ts
justinanastos: Yes, same as https://github.com/apollographql/apollo-tooling/pull/1308#discussion_r291298332:

>Yes it should because `--markdown` isn't implemented yet. I'll leave a comment here and at the other skip sites with the ticket that'll implementing this.

packages/apollo/src/commands/service/__tests__/check.test.ts
justinanastos: Yes, same as https://github.com/apollographql/apollo-tooling/pull/1308#discussion_r291298332:

>Yes it should because `--markdown` isn't implemented yet. I'll leave a comment here and at the other skip sites with the ticket that'll implementing this.

packages/apollo/src/commands/service/check.ts
justinanastos: You got it!

packages/apollo/src/commands/service/check.ts
justinanastos: That makes sense to me. Just to clarify:

For federated services we'll say: "Validating composed schema..."

For non-federated services we'll say: "Validating schema..."

Does that sound right?

packages/apollo/src/commands/service/check.ts
justinanastos: Unloaded question: why do you think this is important @evans ?

packages/apollo/src/commands/service/check.ts
justinanastos: I added the schema has; I don't know if we have a command to actually use it.

packages/apollo/src/commands/service/check.ts
justinanastos: We're enforcing it above, this comment is misplaced. I'm going to remove it.

packages/apollo/src/commands/service/check.ts
evans: Yeah! That sounds right

packages/apollo/src/commands/service/check.ts
evans: Good question! I think it's a nice to have, which would indicate how much work apollo is doing and a nice way for users to understand that the command is working as intended

packages/apollo/src/commands/service/__tests__/check.test.ts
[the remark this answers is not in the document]
trevor-scheer: We resolved this offline - the introspection result needed to be in JSON format

packages/apollo/src/commands/service/__tests__/check.test.ts
trevor-scheer: That's ok with me!

packages/apollo/README.md
```diff
@@ -752,6 +752,11 @@ node --inspect-brk=9002 packages/apollo/bin/run <command>
 
 If you're using VS Code, you can run the included "Attach to CLI Debugger" launch task and debug right from VS Code! Otherwise, you may use the (Chrome inspector)[https://nodejs.org/en/docs/guides/debugging-getting-started/] or other Node debugger of your choice.
 
+## Regenerating Mocked Network Data
```
trevor-scheer: Thank you for adding this!

packages/apollo/src/commands/service/__tests__/check.test.ts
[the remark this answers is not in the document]
trevor-scheer: I would support the addition of that 👍

At this point, I'm so familiar with the dependencies in this project and I knew that wasn't one of them 😬

PR #1343, opened 2019-06-12
buildSchemaFromSDL - support custom scalar resolvers

hannesj:
Fixes https://github.com/apollographql/apollo-server/issues/2832

TODO:


hannesj:
The build error seems to be from something completely different. Also, is there some reason, the functionality has been re-implememted and not used just from https://github.com/apollographql/graphql-tools/blob/master/src/generate/addResolveFunctionsToSchema.ts

hannesj:
Superseded by #1345

PR #1344, opened 2019-06-12
Add enum renaming ability to federated services.

JacksonKearl:
TODO:


JacksonKearl:
I question if the entire addResolversToSchema function isn't just a half-baked reimplementation of https://github.com/apollographql/graphql-tools/blob/master/src/generate/addResolveFunctionsToSchema.ts in graphql-tools. In particular my hacky implementation of renaming enums seems to be more thoughtfully implemented in https://github.com/apollographql/graphql-tools/blob/master/src/transforms/ConvertEnumValues.ts. Thoughts on if we should be pulling in `graphql-tools` and using its implementation?

JacksonKearl:
Superseded by #1345

PR #1345, opened 2019-06-13
Add support for custom enum resolvers and scalar types

hannesj:
As described in #1343 and #1344, it is better to use `addResolveFunctionsToSchema` from graphql-tools to add the resolvers to the new schema, as it is how the resolvers are added normally in apolo-server, so that both scalars and enums are handled properly. As a TODO, it might be sensible to use more functionality from https://github.com/apollographql/graphql-tools/blob/84e77ac7d7e06f149019c5a42dd2e86907e27ac1/src/makeExecutableSchema.ts when generating the new schema.

TODO:

### DIFF
--- a/packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
+++ b/packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
@@ -14,6 +14,7 @@ import {
   SchemaExtensionNode,
   OperationTypeNode,
   GraphQLObjectType,
+  GraphQLEnumType,
   isAbstractType,
   isScalarType,
   isEnumType
@@ -207,6 +208,8 @@ export function addResolversToSchema(
   schema: GraphQLSchema,
   resolvers: GraphQLResolverMap<any>
 ) {
+  const enumValueMap = Object.create(null);
+
   for (const [typeName, fieldConfigs] of Object.entries(resolvers)) {
     const type = schema.getType(typeName);
 
@@ -225,19 +228,29 @@ export function addResolversToSchema(
     }
 
     if (isEnumType(type)) {
-      let getValue = type.getValue.bind(type);
-      type.getValue = x => {
-        if (x in fieldConfigs) {
-          let old = getValue(x);
-          if (!old) {
-            return old;
-          }
-          old.value = (fieldConfigs as any)[x];
-          return old;
-        } else {
-          return getValue(x);
-        }
-      };
+      const values = type.getValues();
+      const newValues = {};
+      values.forEach(value => {
+        const newValue = (fieldConfigs as any)[value.name] || value.name;
+        (newValues as any)[value.name] = {
+          value: newValue,
+          deprecationReason: value.deprecationReason,
+          description: value.description,
+          astNode: value.astNode
+        };
+      });
+
+      // In place updating hack to get around pulling in the full
+      // schema walking and immutable updating machinery from graphql-tools
+      Object.assign(
+        type,
+        new GraphQLEnumType({
+          name: type.name,
+          description: type.description,
+          astNode: type.astNode,
+          values: newValues
+        })
+      );
     }
 
     if (!isObjectType(type)) continue;
--- a/packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
+++ b/packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
@@ -224,6 +224,22 @@ export function addResolversToSchema(
       }
     }
 
+    if (isEnumType(type)) {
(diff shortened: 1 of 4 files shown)


JacksonKearl:
Thanks! This feels much better. Ping @trevor-scheer to ask if this is a bad idea for any reason.

trevor-scheer:
Hey @hannesj, firstly thank you for the details and the contribution. Your analysis and approach are perfectly reasonable.

`graphql-tools` was intentionally left out as a dependency in this project as it's something that we're looking to move away from going forward (Apollo Server included, at some point). I _do_ think that we want to duplicate the missing logic to this function for now (re: abstract and enum types).

Apologies that this isn't documented or explained anywhere yet, but I'd like to add some additional context so this isn't coming out of nowhere.

GraphQL tools includes two major things:
- A way to build a schema from SDL + resolvers
- Schema stitching

We're currently working with maintainers of `graphql-js` to move most of the tools we wrote to build schemas into the core.

With Federation, we're moving away from stitching. Since many teams still rely on stitching, we want to make the `graphql-tools` package a community-owned package and are working on a transition plan for it.

hannesj:
Hi,

I rebased this based on feedback from you, leaving out the adding of the dependency, and using the previous commits.

Unfortunately it seems that the solution for renaming the enums was not enough, as the lastly added test now fails.

packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
```diff
@@ -216,6 +221,38 @@ export function addResolversToSchema(
...
+          value: newValue,
+          deprecationReason: value.deprecationReason,
+          description: value.description,
+          astNode: value.astNode
+        };
+      });
+
+      // In place updating hack to get around pulling in the full
+      // schema walking and immutable updating machinery from graphql-tools
+      Object.assign(
```
JacksonKearl: @trevor-scheer what do you think of this? At first I tried copying direct from grapql-tools, but it ended up requiring a lot more dependencies than I expected. (https://github.com/apollographql/graphql-tools/blob/master/src/transforms/ConvertEnumValues.ts#L10, which requires https://github.com/apollographql/graphql-tools/blob/master/src/transforms/visitSchema.ts#L38, which requires https://github.com/apollographql/graphql-tools/blob/master/src/stitching/schemaRecreation.ts#L27, etc.)

packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
```diff
@@ -216,6 +221,38 @@ export function addResolversToSchema(
...
+        type,
+        new GraphQLEnumType({
+          name: type.name,
```
trevor-scheer: Small nit: we can replace the "cloned" keys `name, description, astNode` with `...type.toConfig()`, which is a relatively new addition for convenience when we transform types, schema, etc.

packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
```diff
@@ -216,6 +221,38 @@ export function addResolversToSchema(
...
 
+    if (isScalarType(type)) {
+      for (const fn in fieldConfigs) {
+        (type as any)[fn] = (fieldConfigs as any)[fn];
+      }
+    }
+
+    if (isEnumType(type)) {
+      const values = type.getValues();
+      const newValues = {};
```
trevor-scheer: We can type this object and remove one of our `any`s:

```js
const newValues: { [key: string]: GraphQLEnumValue } = {};
```

packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
JacksonKearl: Bit confused, `type.toConfig()` doesn't provide `deprecationReason`, and `value` doesn't have a `.toConfig()`.

trevor-scheer: The values work up above is fine, to be more clear I'm suggesting this for building the new `GraphQLEnumType`. There's no `deprecationReason` on an enum type itself, just on its values:
https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L1202-L1224

```js
Object.assign(
  type,
  new GraphQLEnumType({
    ...type.toConfig(),
    values: newValues
  })
);

JacksonKearl: Ah that makes sense, thanks!

trevor-scheer:
Thank you both for this @hannesj and @JacksonKearl! 🎉

PR #1358, opened 2019-06-19
service:list in CLI

mayakoneval:
This PR adds `service:list` to the Apollo CLI tool, printing out the lists in a federated services like so:
image

Open questions:
- ~is copy / pasting the functions `uncaptureApplicationOutput` and `captureApplicationOutput` necessary / how do you share those testing helpers with the shared console log?~ (resolved - see comment below)

TODO:


packages/apollo/src/commands/service/__tests__/list.test.ts
```diff
@@ -0,0 +1,229 @@
...
+let mockedConsoleLogValues: string[] | null = null;
+
+/**
+ * Mock and capture `console.log` and `stdout.write`s. Return them in that order as a single string.
+ *
+ * This will emulate what the output of running the CLI would look like.
+ *
+ * Call `uncaptureApplicationOutput` to reverse the effects of this function.
+ */
+function captureApplicationOutput() {
```
mayakoneval: this and the following function are copied from `check.test.ts` I am not sure how to share them with a shared `console.log`

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+          );
+        }
+
+        /**
+         * Name of the graph being checked. `engine` is an example of a graph.
+         *
+         * A graph can be either a monolithic schema or the result of composition a federated schema.
+         */
+        const graphName = config.name;
+        const tag = flags.tag || "current"; //|| config.tag || "current";
```
mayakoneval: forgot to continue investigating this - looking today

justinanastos: I don't actually have a good answer for how to capture without copy/pasting. If the copy/paste works, then do that I guess 🤷‍♂

mayakoneval: This may be causing tests to fail ;(

packages/apollo/src/commands/service/__tests__/list.test.ts
zionts: hmmm... I think this would actually hugely benefit from passing in a custom logger to the entirety of the `apollo` CLI. Probably a feature we should log somewhere, if it's not already. That way you can just instantiate the tests with a `writeToMap` logger that stores the log output in an array under the test name or something.

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+  return result;
+}
+
+export default class ServiceList extends ProjectCommand {
+  static description = "List the services in a graph";
+  static flags = {
+    ...ProjectCommand.flags,
+    tag: flags.string({
+      char: "t",
+      description: "The published tag to list the services from"
```
zionts: nit: variant? IDK what our plan is for that migration tbh, but it's confusing that we're using graph everywhere and still using tag.

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+    })
+    // json: flags.boolean({
```
zionts: it this a TODO?

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+    try {
+      await this.runTasks<TasksOutput>(({ config, flags, project }) => {
+        if (!isServiceProject(project)) {
+          throw new Error(
+            "This project needs to be configured as a service project but is configured as a client project. Please see bit.ly/2ByILPj for help regarding configuration."
+          );
+        }
+
+        /**
+         * Name of the graph being checked. `engine` is an example of a graph.
```
zionts: kind of confusing example of a `graph`, but oh well 🤷‍♂

packages/apollo/src/commands/service/list.ts
zionts: what's up with this commented out line?

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+         *
+         * A graph can be either a monolithic schema or the result of composition a federated schema.
+         */
+        const graphName = config.name;
```
zionts: I'm confused. Shouldn't the `name` in the config for a service project be the service name, not the graph name?

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+              } = await project.engine.listServices({
+                id: graphName,
+                graphVariant: tag
+              });
+              const newContext: typeof ctx = {
+                implementingServices,
+                config
+              };
+
+              Object.assign(ctx, newContext);
```
zionts: I thought people preferred the spread operator thing now, but idk things

packages/apollo/src/commands/service/list.ts
mayakoneval: thats old :) whoops

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+        // Override `printHeader` so we don't print a header
+        printHeader: () => {},
+        // The default `printLine` will output to the console; we want to capture the output so we can test
+        // it.
+        printLine: line => {
+          result += `\n${line}`;
+        }
+      }
+    );
+    const serviceListUrlEnding = `/service/${graphName}/service-list`;
```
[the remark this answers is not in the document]
mayakoneval: fixed in new commits!

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+  frontendUrl
+}: {
+  implementingServices: ListServices_service_implementingServices | null;
+  graphName: string | undefined;
+  frontendUrl: string | undefined;
+}): string {
+  let result = "";
+  if (
+    !implementingServices ||
+    implementingServices.__typename === "NonFederatedImplementingService"
```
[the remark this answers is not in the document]
mayakoneval: at the moment, there is no way to get here, but since `implementingServices` can be either a federated or non federated implementing service at the moment (union of types) I decided to double check. Do you think it's not necessary?

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,173 @@
...
+    const targetUrl = frontendUrl
+      ? `${frontendUrl}${serviceListUrlEnding}`
+      : `https://engine.apollographql.com${serviceListUrlEnding}`;
+    result += `\n\nView full details at: ${targetUrl}`;
+  }
+  return result;
+}
+
+export default class ServiceList extends ProjectCommand {
+  static description = "List the services in a graph";
```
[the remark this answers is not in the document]
mayakoneval: Everywhere else, this is just named service (engine UI), so I am in favor of keeping them consistent.

packages/apollo/src/commands/service/list.ts
mayakoneval: Agreed, but the whole cli would have to make this change before this PR does!

packages/apollo/src/commands/service/list.ts
mayakoneval: I'm not sure I understand what a service project is, but as far as I know this is the graph name.

packages/apollo-language-server/src/engine/index.ts
```diff
@@ -77,6 +80,29 @@ export class ApolloEngineClient extends GraphQLDataSource {
...
+      variables
+    }).then(({ data, errors }) => {
+      // use error logger
+      if (errors) {
+        throw new Error(errors.map(error => error.message).join("\n"));
+      }
+
+      if (data && !data.service) {
+        throw new Error(
+          noServiceError(getServiceFromKey(this.engineKey), this.baseURL)
```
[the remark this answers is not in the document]
mayakoneval: yep! `Could not find service from Engine at ___. Please check your API key and service name`

packages/apollo-language-server/src/engine/index.ts
```diff
@@ -77,6 +80,29 @@ export class ApolloEngineClient extends GraphQLDataSource {
...
+        );
+      }
+
+      if (!(data && data.service)) {
+        throw new Error("Error in request from Engine");
```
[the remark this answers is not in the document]
mayakoneval: sure, this is something we can discuss. Right now this pattern is used with if not all, a lot of the cli.

packages/apollo-language-server/src/engine/operations/listServices.ts
```diff
@@ -0,0 +1,21 @@
+import gql from "graphql-tag";
+
+export const LIST_SERVICES = gql`
+  query ListServices($id: ID!, $graphVariant: String!) {
+    service(id: $id) {
+      implementingServices(graphVariant: $graphVariant) {
+        __typename
```
trevor-scheer: I think we can leave `__typename` out everywhere, should be taken care of for us by our tooling.
(auto-added in our requests, as well as our generated types)

packages/apollo/src/commands/service/__tests__/list.test.ts
trevor-scheer: @zionts, agree on both fronts - definitely a feature we should consider adding. In the meantime, @mayakoneval can you add comments in both places that this code is duplicated by necessity? Enough to provide some context for future implementors, as they may have to do the same thing.

packages/apollo/src/commands/service/__tests__/list.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  describe("integration", () => {
+    describe("should list services correctly for a federated service", () => {
+      it("vanilla", async () => {
+        captureApplicationOutput();
+        mockServiceListFederated();
+
+        expect.assertions(2);
+
+        await expect(
+          ServiceList.run([...cliKeyParameter])
```
trevor-scheer: Nit: you're spreading an array into an array, so should just be able to pass the array itself:
`ServiceList.run(cliKeyParameter)`

I think right now it's effectively:
```ServiceList.run([...[`--key=${apiKey}`]])```

File: packages/apollo/src/commands/service/__tests__/list.test.ts [hunk 15]
```diff
@@ -0,0 +1,229 @@
...
+
+    describe("should display non federated message for a non federated service", () => {
+      it("vanilla", async () => {
+        captureApplicationOutput();
+        mockServiceListNonFederated();
+
+        expect.assertions(2);
+
+        await expect(
+          ServiceList.run([...cliKeyParameter])
```
trevor-scheer: Same here

trevor-scheer: Consider a rename, since `cliKeyParameter` is itself an array. Maybe `cliOptions` or similar that represents a list of things we're passing.

packages/apollo/src/commands/service/list.ts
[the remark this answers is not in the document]
trevor-scheer: I think TS requires this check based on the generated types and I fully support this! It may read a little more clearly to refactor the logic, but this is just a nit.

```js
if (!(implementingServices && implementingServices.__typename === "FederatedImplementingService") {
  ...
}
```

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,168 @@
...
+  } else if (implementingServices.services.length === 0) {
+    result = "\nThere are no services on this federated graph";
+  } else {
+    // Create a sorted list of the services.
+    const sortedImplementingServices = sortBy<
+      ListServices_service_implementingServices_FederatedImplementingServices_services
+    >(implementingServices.services, [service => service.name.toUpperCase()]);
+
+    table(
+      [...sortedImplementingServices.map(formatImplementingService)].filter(
```
trevor-scheer: This is spreading an array into an array, which gives you an array. We can simplify like before 😄 

```js
sortedImplementingServices.map(formatImplementingService).filter(
  ...
)
```

packages/apollo/src/commands/service/list.ts
```diff
@@ -0,0 +1,168 @@
...
+        // it.
+        printLine: line => {
+          result += `\n${line}`;
+        }
+      }
+    );
+    const serviceListUrlEnding = `/graph/${graphName}/service-list`;
+    const targetUrl = frontendUrl
+      ? `${frontendUrl}${serviceListUrlEnding}`
+      : `https://engine.apollographql.com${serviceListUrlEnding}`;
```
trevor-scheer: Can we reference the constant / source of truth where this string actually lives? This way if we update one, they both change!

```js
export const DefaultEngineConfig = {
  endpoint: "https://engine-graphql.apollographql.com/api/graphql",
  frontend: "https://engine.apollographql.com"
};
```

Issue #1362, opened 2019-06-25
operationId is invalid for swift codegen


Codebear98:
- operationId mismatch with sever due to missing `\n` in apollo-codegen-swift.

You can find there is `\n` to separate fragments in `apollo-codegen-core`
```
  const sourceWithFragments = [
    operation.source,
    ...Array.from(fragmentsReferenced).map(fragmentName => {
      const fragment = fragments[fragmentName];
      if (!fragment) {
        throw new Error(`Cannot find fragment "${fragmentName}"`);
      }
      return fragment.source;
    })
  ].join("\n");
```
https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-codegen-core/src/compiler/visitors/generateOperationId.ts

However in `apollo-codegen-swift`, the missing `\n` causing operationId mismatch with server.

```
if (fragmentsReferenced.size > 0) {
  this.printNewlineIfNeeded();
  this.printOnNewline(
    "public var queryDocument: String { return operationDefinition"
  );
  fragmentsReferenced.forEach(fragmentName => {
    this.print(
      `.appending(${this.helpers.structNameForFragmentName(
        fragmentName
      )}.fragmentDefinition)`
    );
  });
  this.print(" }");
}
```

https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-codegen-swift/src/codeGeneration.ts

designatednerd:
@Codebear98 Is this any better with iOS 0.13.0 (calling through to tooling 2.16) that's adding the trailing newline now?

Codebear98:
checked apollo-ios 0.13.0 don't have such issues anymore, operationID is now valid! @designatednerd thanks for the upgraded!

PR #1363, opened 2019-06-25
Fix issue with operation id doesn't match query document for swift (#1362) 

Codebear98:
Added back `n` to each fragment in codeGeneration.ts 

#1362


zionts:
Hi @Codebear98 — thanks for putting this together, and sorry for the delay on getting back to you! Since so much has changed in the repo in the meantime, I'm going to close this out for now, but feel free to reopen or to file an issue if this is still a problem for you!

PR #1378, opened 2019-07-05
Support --json and --markdown for federated schema:check

jgzuke:
Support `--markdown` and `--json` flags for `apollo schema:check` command. Reintroduce tests and update snapshots.

Also refactor `check.ts` slightly, including renaming `serviceName` to `graphName` in markdown output now that there is a more pronounced difference between the terms with federation.

TODO:

### DIFF
--- a/packages/apollo/src/commands/service/check.ts
+++ b/packages/apollo/src/commands/service/check.ts
@@ -66,26 +66,31 @@ export function formatTimePeriod(hours: number): string {
   return pluralize(Math.floor(hours / 24), "day");
 }
 
+type CompositionErrors = Array<{
+  service?: string;
+  field?: string;
+  message: string;
+}>;
+
 interface TasksOutput {
   config: ApolloConfig;
   checkSchemaResult: CheckSchema_service_checkSchema;
   shouldOutputJson: boolean;
   shouldOutputMarkdown: boolean;
   federationSchemaHash?: string;
-  compositionErrors?: Array<{
-    service?: string;
-    field?: string;
-    message: string;
-  }>;
+  serviceName: string | undefined;
+  compositionErrors?: CompositionErrors;
 }
 
 export function formatMarkdown({
   checkSchemaResult,
+  graphName,
   serviceName,
   tag
 }: {
   checkSchemaResult: CheckSchema_service_checkSchema;
-  serviceName: string;
+  graphName: string;
+  serviceName?: string | undefined;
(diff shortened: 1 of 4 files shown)


packages/apollo/src/commands/service/check.ts
```diff
@@ -147,6 +154,32 @@ ${
...
+  return `
+### Apollo Service Check
+🔄 Validated graph composition on schema tag \`${tag}\` for service \`${serviceName}\` on graph \`${graphName}\`.
+❌ Found **${compositionErrors.length} composition errors**
+
+| Service   | Field     | Message   |
+| --------- | --------- | --------- |
+${compositionErrors
+  .map(
+    ({ service, field, message }) => `| ${service} | ${field} | ${message} |`
```
mayakoneval: Is there a way to make these line up with the header? I think in the human readable sections we use the heroku table.

packages/apollo/src/commands/service/check.ts
```diff
@@ -602,32 +597,77 @@ export default class ServiceCheck extends ProjectCommand {
...
         })
       );
     }
 
-    this.log(formatHumanReadable({ checkSchemaResult }));
+    if (compositionErrors) {
+      // Add a cosmetic line break
+      console.log("");
+
+      cli.table(compositionErrors, {
```
mayakoneval: Just curious, does the heroku table and the cli table display tables differently?

packages/apollo/src/commands/service/check.ts
```diff
@@ -602,32 +597,77 @@ export default class ServiceCheck extends ProjectCommand {
...
+
+      const { service } = config;
+      if (!service) {
+        throw new Error(
+          "Service mising from config. This should have been validated elsewhere"
+        );
+      }
 
-      if (!serviceName) {
+      const graphName = config.service && config.service.name;
```
mayakoneval: This might be a nice place for a comment explaining the naming conventions as a result of federation / the mismatch between this and the gateway!

packages/apollo/src/commands/service/check.ts
jgzuke: Not sure! I havent seen the output of heroku table. This was moved from https://github.com/apollographql/apollo-tooling/pull/1378/files#diff-c61805bdfe85787383d4ae9927bd0774L567 to keep the non json/markdown behavior the same

PR #1385, opened 2019-07-10
Dart support for vscode-apollo and language-server

### LINKED ISSUE #1365: Add support for dart in vscode-dart
We are working on improving graphql support for dart. vscode-apollo is the best plugin to provide syntax highlighting in vscode and we wouldn't want to create a separate plugin just to allow syntax highlighting in dart.

This would just be a matter of adding a new template syntax to vscode-apollo. We can use this as a reference implementation: https://github.com/apollographql/apollo-tooling/blob/master/packages/vscode-apollo/syntaxes/graphql.py.json

If the Apollo team is OK with this, I or someone on my team can submit a pull request with the additional syntax.

The syntax we would support will look like this:
```dart
var query = gql("""
                query GetCard(\$id:CardId!) {
                  card(id:\$id) {
                    id
                    name
                    completed
                }
                """);
```

venkatd:
This resolves #1365.

Similar to how python handles this, adding support for graphql syntax highlighting and code completion using a basic function.

```dart
String gql(String x) => x;

var query = gql(r"""
query getPeople {
  allPeople {
    people {
      id
      birthYear
      created
    }
  }
}
""");
```

Screenshot of it in action:

image

I need a bit of guidance on building and testing the extension locally (including the language server). What's the process developers are typically using on `vscode-apollo`?

In order to test the functionality out, I patched the extension and language server directly in `/Users/venkat/.vscode/extensions/apollographql.vscode-apollo-1.7.4`. Then I applied those same changes to this repo. However I'd like to double check it all works.

### DIFF
--- a/packages/vscode-apollo/package.json
+++ b/packages/vscode-apollo/package.json
@@ -110,6 +110,16 @@
         "embeddedLanguages": {
           "meta.embedded.block.graphql": "graphql"
         }
+      },
+      {
+        "injectTo": [
+          "source.dart"
+        ],
+        "scopeName": "inline.graphql.dart",
+        "path": "./syntaxes/graphql.dart.json",
+        "embeddedLanguages": {
+          "meta.embedded.block.graphql": "graphql"
+        }
       }
     ],
     "commands": [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,11 +19,11 @@
 - `apollo-graphql`
   - <First `apollo-graphql` related entry goes here>
 - `apollo-language-server`
-  - <First `apollo-language-server` related entry goes here>
+  - Add Dart operation extraction [#1385](https://github.com/apollographql/apollo-tooling/pull/1385)
 - `apollo-tools`
   - <First `apollo-tools` related entry goes here>
 - `vscode-apollo`
-  - <First `vscode-apollo` related entry goes here>
+  - Add Dart support for vscode [#1385](https://github.com/apollographql/apollo-tooling/pull/1385)
 
 ## `apollo@2.16.1`, `apollo-language-server@1.13.1`, `vscode-apollo@1.8.1`
 
--- a/packages/apollo-language-server/src/document.ts
+++ b/packages/apollo-language-server/src/document.ts
@@ -70,6 +70,8 @@ export function extractGraphQLDocuments(
       return extractGraphQLDocumentsFromPythonStrings(document, tagName);
     case "ruby":
       return extractGraphQLDocumentsFromRubyStrings(document, tagName);
+    case "dart":
+      return extractGraphQLDocumentsFromDartStrings(document, tagName);
     default:
       return null;
   }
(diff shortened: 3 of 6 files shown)


packages/vscode-apollo/package.json
```diff
@@ -100,6 +100,16 @@
         "embeddedLanguages": {
           "meta.embedded.block.graphql": "graphql"
         }
+      },
+      {
+        "injectTo": [
+          "source.dart"
+        ],
+        "scopeName": "inline.dart.python",
```
JakeDawkins: It looks like this is wrong. You'd need it to be `inline.graphql.dart`. That fixed the highlighting issue for me :)

venkatd: 🤦‍♂️thanks for finding that! I ran the extension in debug mode with this change and everything worked.

venkatd:
@JakeDawkins looks like it all works when running in debug mode now :).

When I build the `.vsix` extension and install it like this, I get the following error:

```
[Error - 10:45:32 PM] Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: Cannot use GraphQLList "[Person]" from another module or realm.

Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate "graphql" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.
  Code: -32603 
```

Could it be a conflict with running `npm i` and how lerna handles it? Will look into it further

JakeDawkins:
@venkatd I ran the same thing and it worked for me! You may need to uninstall any existing version of the extension and delete the extension cache for the install to work properly. At least if the VSIX file that's being generated is the same as a version you've previously installed.

`rm -rf  ~/.vscode/extensions/apollographql.vscode-apollo-1.7.5` with whatever version you're trying to install

venkatd:
@JakeDawkins thanks, I got it working! Cleared my cache, re-cloned the repo, and basically wiped the slate clean. I see now there are some merge conflicts. What's the preferred way to address them? Should I merge master into my branch or rebase?

JakeDawkins:
@venkatd If you could rebase this, that'd be great!

venkatd:
@JakeDawkins just rebased and tested it out with a built .vsix extension again. Let me know if there's anything else needed before merge!

JakeDawkins:
Thanks so much for the work (and the patience) @venkatd! Just released with v1.9.0 of the extension!

venkatd:
@JakeDawkins thanks for the guidance along the way and glad it's in the new release! Just tried the official upgrade all looks to be working.

AlexKenbo:
> @JakeDawkins thanks for the guidance along the way and glad it's in the new release! Just tried the official upgrade all looks to be working.

Tell me how to set it up?
I only have a schematic - will this work?

venkatd:
@AlexKenbo the dart-specific setup just requires you to define a function like this:
```dart
String gql(String query) => query;
```

Because it will interpret anything like so as GraphQL:
```dart
gql("""
query getData {
...
}
""")
```

PR #1396, opened 2019-07-14
[Federation] service:push, user experience improvements

zionts:
* Do not require users to specify `--federated` as long as `--serviceName` is provided
* Change error messages for lack of full specification to point to how to remedy the problem
* Provide step-by-step output logging information regarding which aspect of service:push succeeded or failed

TODO:

### DIFF
--- a/packages/apollo/src/commands/service/push.ts
+++ b/packages/apollo/src/commands/service/push.ts
@@ -200,9 +200,9 @@ export default class ServicePush extends ProjectCommand {
 
     if (result.didUpdateGateway) {
       this.log(
-        `The gateway for the ${graphString} graph was updated with a new schema, composed from the updated ${
+        `The gateway for the ${graphString} graph was updated with a new schema, composed from the updated '${
           result.implementingServiceName
-        } service\n`
+        }' service\n`
       );
     } else if (isFederated) {
       this.log(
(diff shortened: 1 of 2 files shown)


packages/apollo/src/commands/service/push.ts
```diff
@@ -175,30 +203,26 @@ export default class ServicePush extends ProjectCommand {
 
     if (result.didUpdateGateway) {
       this.log(
-        `The gateway for the ${
-          result.graphName
-        } graph was updated with a new schema\n`
+        `The gateway for the ${graphString} graph was updated with a new schema, composed from the updated ${
+          result.service
```
trevor-scheer: Now I'm a bit confused, based on my previous comments I slightly misunderstood `service` vs. `graphName`. Can we document the distinction if there's supposed to be one? In one case, they're the same, but in another, they're not. If you know the answer to my misunderstanding, some comments would be greatly appreciated. I don't expect you to have that necessarily, though, unless both codepaths are related to the work you're doing.

zionts: Yeah, I think this could do with some clean-up. How about I take another pass at this and change around some names. I'll run the commands locally to make sure everything still prints out correctly.

trevor-scheer:
LGTM!

PR #1398, opened 2019-07-15
[WIP] variant vs. tag vs. config.tag clean-up

zionts:
This PR started out trying to do something about `service:delete` still requiring `--federated`, but then got super distracted and tried to clean up all of the `flags.variant` vs. `flags.tag` nonsense, but in a probably not super-productive way. I'm not sure how useful those commits are, so we can certainly revert them if we have another solution for this, but consistency is key, and it's something we're sorely lacking here. Leaving this open for now as a point of discussion rather than as an intended solution. I'm hoping someone who understands the nuances better can pick this up :)

TODO:


zionts:
Closing this out due to bitrot 🤖

PR #1399, opened 2019-07-15
Do not require `--federated` for service:delete

zionts:
Use --serviceName like everywhere else. Also deprecate the --federated
flag.

TODO:

### DIFF
--- a/packages/apollo/src/commands/service/delete.ts
+++ b/packages/apollo/src/commands/service/delete.ts
@@ -15,7 +15,7 @@ export default class ServiceDelete extends ProjectCommand {
       char: "f",
       default: false,
       description:
-        "Indicates that the schema is a partial schema from a federated service"
+        "[Deprecated: use --serviceName to indicate federation] Indicates that the schema is a partial schema from a federated service"
     }),
     serviceName: flags.string({
       required: true,
@@ -34,9 +34,9 @@ export default class ServiceDelete extends ProjectCommand {
             throw new Error("No service found to link to Engine");
           }
 
-          if (!flags.federated) {
+          if (!flags.federated && !flags.serviceName) {
             this.error(
-              "Deleting a service is only supported for federated services. Use the --federated flag if this is a federated service."
+              "Deleting a service is only supported for federated services. Use the --serviceName flag if this is a federated service."
             );
           }
 
--- a/packages/apollo/src/commands/service/delete.ts
+++ b/packages/apollo/src/commands/service/delete.ts
@@ -14,6 +14,7 @@ export default class ServiceDelete extends ProjectCommand {
     federated: flags.boolean({
       char: "f",
       default: false,
+      hidden: true,
       description:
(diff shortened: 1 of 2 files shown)


packages/apollo/src/commands/service/delete.ts
```diff
@@ -34,9 +34,9 @@ export default class ServiceDelete extends ProjectCommand {
             throw new Error("No service found to link to Engine");
           }
 
-          if (!flags.federated) {
+          if (!flags.federated && !flags.serviceName) {
```
trevor-scheer: Can we remove the logical dependency on `flags.federated` altogether here?

```suggestion
          if (!flags.serviceName) {
```

packages/apollo/src/commands/service/delete.ts
```diff
@@ -34,9 +34,9 @@ export default class ServiceDelete extends ProjectCommand {
             throw new Error("No service found to link to Engine");
           }
```
trevor-scheer: Can we log a deprecation notice for the federated flag here?

```js
if (flags.federated) {
  this.log("The --federated flag is no longer required when running federated commands. Use of the flag will not be supported in future versions of the CLI.");
}
```

packages/apollo/src/commands/service/delete.ts
```diff
@@ -34,9 +34,9 @@ export default class ServiceDelete extends ProjectCommand {
...
 
-          if (!flags.federated) {
+          if (!flags.federated && !flags.serviceName) {
             this.error(
-              "Deleting a service is only supported for federated services. Use the --federated flag if this is a federated service."
+              "Deleting a service is only supported for federated services. Use the --serviceName flag if this is a federated service."
```
trevor-scheer: This entire condition shouldn't be possible given that it's a `required` flag. The CLI framework will throw if the flag isn't provided. If we need it for the sake of making TS happy, we can leave it in - but a TODO comment to remove it altogether would be great (should just need to get the types correct).

packages/apollo/src/commands/service/delete.ts
```diff
@@ -15,7 +15,7 @@ export default class ServiceDelete extends ProjectCommand {
       char: "f",
       default: false,
       description:
```
trevor-scheer: Can we add the `hidden` flag here as well?

packages/apollo/src/commands/service/push.ts
```diff
@@ -24,8 +24,9 @@ export default class ServicePush extends ProjectCommand {
     federated: flags.boolean({
       char: "f",
       default: false,
+      hidden: true,
       description:
-        "Indicates that the schema is a partial schema from a federated service"
+        "[Deprecated: use --serviceName to indicate federation] Indicates that the schema is a partial schema from a federated service"
```
trevor-scheer: As above, can we log a deprecation notice for the federated flag at the beginning of this command?

```js
if (flags.federated) {
  this.log("The --federated flag is no longer required when running federated commands. Use of the flag will not be supported in future versions of the CLI.");
}
```

PR #1412, opened 2019-07-18
[BE-362] Add overrideComposedSchema option to service:push

zionts:
By default, it is forbidden to push a schema directly to a federated
graph. If users try to push a schema directly, the API will return
errors unless 'overrideComposedSchema' is specified in the arguments.
This change adds the boolean flag that corresponds to this "force push"
to the CLI.

It also runs lint-fix, unfortunately in the same commit.

See https://github.com/mdg-private/monorepo/pull/2533 for corresponding PR

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - `apollo`
   - Shorten `client:check` and `service:check` output in CI [#1404](https://github.com/apollographql/apollo-tooling/pull/1404)
   - service:check add null check for validation config [#1471](https://github.com/apollographql/apollo-tooling/pull/1471)
+  - `service:push` add `overrideComposedSchema` option, which allows schemas to be force-pushed directly to federated graphs
 - `apollo-codegen-core`
   - <First `apollo-codegen-core` related entry goes here>
 - `apollo-codegen-flow`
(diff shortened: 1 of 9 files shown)


zionts:
@trevor-scheer I'm opening this as a draft PR until https://github.com/mdg-private/monorepo/pull/2533 is merged and deployed, but it's ready for a review.

zionts:
@JakeDawkins not sure how to fix this up, getting 
```
error TS2688: Cannot find type definition file for 'apollo-env'.
```
when trying to build

zionts:
We can also close this out for now

trevor-scheer:
@zionts what's the reason for enabling this behavior via flag at all? Why would I want to override the composed schema? Seems like a footgun, allowing users to get into an invalid state w.r.t. the state of the individual services.

zionts:
I'm completely fine closing this out until a customer needs it and we have a valid use case! I _do_ think that a valid use case will come up, and the reason for the push of urgency was that a user wrote in asking, but it was a misinterpretation. I think we can still actually close this, but we'll need to update the error message cloud-side if so :)

zionts:
Curious what your thoughts are @JakeDawkins — I've got no real horse in this race; it just felt like the natural configuration option to go in parallel with the restriction we have, especially if someone wanted to revert to a monolithic infrastructure from a federated one.

zionts:
If we ever decide to reopen this, we should change the error message back after https://github.com/mdg-private/monorepo/pull/2872

PR #1413, opened 2019-07-18
Add helpful message for invalid API key

JakeDawkins:
Right now, there's no error being printed when schema lookup fails. 

This PR aims to fix that, and add at least a safeguard for people using incorrect API keys. This can be partially checked on the client by checking the `config.service.name` against the identifier in the API key.

See the comments in the file changes for more explanation.

For the future, we really should standardize our error strategy to make sure they're being printed prettily in the editor. Right now, we're getting some pretty ugly stack traces.

TODO:

### DIFF
--- a/packages/apollo-language-server/src/engine/index.ts
+++ b/packages/apollo-language-server/src/engine/index.ts
@@ -274,8 +274,12 @@ export class ApolloEngineClient extends GraphQLDataSource {
       }
     });
 
-    if (!(data && data.service)) {
-      throw new Error();
+    if (!(data && data.service) || errors) {
+      throw new Error(
+        errors
+          ? errors.map(error => error.message).join("\n")
+          : "No service returned. Make sure your service name and API key match"
+      );
     }
 
     const schemaTags: string[] = data.service.schemaTags.map(
--- a/packages/apollo-language-server/src/project/client.ts
+++ b/packages/apollo-language-server/src/project/client.ts
@@ -169,7 +169,8 @@ export class GraphQLClientProject extends GraphQLProject {
         total: totalTypes
       },
       tag: this.config.tag,
-      loaded: this.serviceID ? true : false,
+      loaded:
+        this.serviceID && (this.schema || this.serviceSchema) ? true : false,
       lastFetch: this.lastLoadDate
     };
   }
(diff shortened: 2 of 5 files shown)


packages/apollo-language-server/src/project/client.ts
```diff
@@ -169,7 +169,8 @@ export class GraphQLClientProject extends GraphQLProject {
         total: totalTypes
       },
       tag: this.config.tag,
-      loaded: this.serviceID ? true : false,
+      loaded:
+        this.serviceID && (this.schema || this.serviceSchema) ? true : false,
```
JakeDawkins: This checks to make sure the project actually _has_ a schema. Previously, this would pass if there are 0 types on both the client and service, and it would show users in the console that everything was fine. This is _not_ true :)

packages/apollo-language-server/src/project/client.ts
```diff
@@ -271,15 +272,19 @@ export class GraphQLClientProject extends GraphQLProject {
...
+            schemaTags,
+            fieldStats
+          } = await engineClient.loadSchemaTagsAndFieldStats(serviceID);
+          this._onSchemaTags && this._onSchemaTags([serviceID, schemaTags]);
+          this.fieldStats = fieldStats;
+          this.lastLoadDate = +new Date();
+
+          this.generateDecorations();
+        } catch (e) {
+          console.error(e);
```
JakeDawkins: Just wrapped this in a try/catch so we're not letting any unhandled rejections through

packages/apollo-language-server/src/engine/index.ts
```diff
@@ -274,8 +274,12 @@ export class ApolloEngineClient extends GraphQLDataSource {
       }
     });
 
-    if (!(data && data.service)) {
-      throw new Error();
+    if (!(data && data.service) || errors) {
+      throw new Error(
```
JakeDawkins: if we can't find a service, we should at least give a recommendation for why

packages/apollo-language-server/src/project/client.ts
trevor-scheer: Nit: `Boolean(this.serviceID && (this.schema || this.serviceSchema))`

packages/apollo-language-server/src/project/client.ts
trevor-scheer: Just making a note that I believe that errors here are meant to bubble up to the LoadingHandler. They're try/catch in the handler so it can communicate them to the extension. This may be a better UX or we may need to improve the messages we send across the connection.

packages/apollo-language-server/src/providers/schema/engine.ts
```diff
@@ -44,6 +45,15 @@ export class EngineSchemaProvider implements GraphQLSchemaProvider {
     }
 
     const [id, tag = "current"] = parseServiceSpecifier(client.service);
+
+    // make sure the API key is valid for the service we're requesting a schema of.
+    const keyServiceName = getServiceFromKey(engine.apiKey);
+    if (id !== keyServiceName) {
```
trevor-scheer: Yesssss

packages/apollo-language-server/src/project/client.ts
JakeDawkins: This has to be here or else I get an unhandled rejection error 😬 I'd love to know how to properly fix this though

PR #1420, opened 2019-07-20
[language server/vscode]: Fix find all references

### LINKED ISSUE #1418: vscode-apollo: Find All References doesn't work in most cases
**Intended outcome:**

I have a workspace similar to this:

```
packages/graphql-server
packages/client1
packages/client2
```

client1 and client2 both have their own `apollo.config.js` similar to this:
```
module.exports = {
  client: {
    addTypename: true,
    service: {
      name: "graphql-server",
      localSchemaFile: "schema.json"
    }
  }
};
```

schema.json is generated using apollo:schema:download.

If I right-click and select `Find All References` on a field definition in graphql-server, it should find references to that field in queries/mutations defined in client1/client2.

**Actual outcome:**

`Find All References` always returns `No Results`.

right-click and select `Find All References` on a field definition

**Versions**
Running latest master as of Jul 18, 2019 https://github.com/apollographql/apollo-tooling/commit/e700ffad860b01492f885bd6176c6ac8fccd6ecb

So I was debugging the `Find All References` functionality myself to try and understand how it works. https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-language-server/src/languageProvider.ts#L387 is the relevant piece of code. It looks to only work if the file you're running the `Find All References` from is itself a client project and the type of the node you selected in either a `FieldDefinition` or `FragmentDefinition`. And then it looks in that same project for references.

That doesn't make sense to me. Wouldn't field definitions always be wherever the graphql server is implemented, outside of a client project?

Note that the `FragmentDefinition` block works as expected because fragment definitions are defined in the client.

TLadd:
When find all references is called, iterate over all projects in
the workspace to look for references instead of confining the search
to the project that triggered the `Find all references`

Fixes #1418

TODO:


packages/apollo-language-server/src/languageProvider.ts
```diff
@@ -351,26 +351,31 @@ ${argumentNode.description ? argumentNode.description : ""}
...
-    const document = project.documentAt(uri, position);
+    const projectFindReferencesRunIn = this.workspace.projectForFile(uri);
+    if (!projectFindReferencesRunIn) return null;
+    const document = projectFindReferencesRunIn.documentAt(uri, position);
     if (!(document && document.ast)) return null;
 
-    if (!project.schema) return null;
+    function getRefsInSingleProject(project: GraphQLProject) {
+      if (!project.schema) return null;
+      if (!(document && document.ast)) return null;
```
wtgtybhertgeghgtwtg: I don't think you have to repeat this every iteration, since it's already checked above and'll be the same `document`.

TLadd: The duplicate check is required by Typescript because `document.ast` could have been mutated in between the time that the first check occurs and when this function gets called. I did hoist out `positionInDocument`.

zionts:
@TLadd , this looks like a great bug-fix, and it also looks to actually be remarkably merge-able, given nearly half a year has passed here. If you would like to clean up the changelog here, I'm happy to take a pass at review! Let me know what works for you :)

TLadd:
Thanks @zionts. Fixed the changelog.

Issue #1427, opened 2019-07-21
codegen: output warning if query type generation is suppressed because of naming conflict


thefliik:
As mentioned in #670, `apollo codegen` requires query names to be unique for a project when generating types. If two queries have the same name, type generation for the second query will be silently skipped.

This behavior is unexpected. It would be very helpful is apollo codegen output a console warning, including a filepath, whenever this occurs.

JakeDawkins:
Hey @thefliik, I'd like to see if I can add a warning to this. Can you provide me with a simple reproduction repo that's affected by this to save me some time?

thefliik:
@JakeDawkins here you go: https://github.com/thefliik/apollo-tooling-issue

I left some simple reproduction steps in the repo's readme

Issue #670, opened 2018-11-08
Why does codegen:generate require query names to be unique?


everdimension:
It makes no sense.

I have a component with a query like this (simplified example):

```
const postsQuery = gql`
  query Posts {
    posts {
      id
      title
    }
  }
`;
```

And another componentn with a query like this:

```
const postsQuery = gql`
  query Posts {
    posts {
      id
      author
    }
  }
`;
```

When I run `codegen:generate` to generate typescript type, I get an error:

> *There can be only one operation named "Posts"*

This forces me to think of unique query names for each query. But why? I don't need a unique query name here. The query is unique based on its content. The name is only helpful for debugging purposes.

And everything works fine. Apollo and graphql don't care about two queries having the same name. But I cannot generate types.

This seems to me to be horribly inconvenient. I want to colocate queries with my components. With this rule you could just as well disallow having two functions with the same name across the project.

Can this check be disabled somehow? Or do you think that my approach is wrong? If so, can you please try to explain the best course of action?

warmbowski:
I just ran into this too. Yeah, if I'm colocating the generated types near the component, I would like to opt out of the de-duplication of query names.

JakeDawkins:
Thanks for the thoughts, everyone! I've definitely thought about this a good bit.

As mentioned above, currently typenames are dependent on the operation name. While unique operation naming may not be a requirement in any way for GraphQL, I'd argue that it's also a best practice. 

Naming operations has many more purposes than just a display name in code. Naming them uniquely makes them easier to find in a codebase, in debugging tools, etc. Not to mention the possibility of importing the wrong types when they're not unique. In addition to that, a lot of our current/future features rely on operation name for things like operation registries, metrics tracking, etc. 

Right now, for those reasons, we don't have any plans to build in support for non-unique operation names, especially since the workaround is so simple.

I'm open to the idea of additional warnings and errors though, if that's what it takes to make these things easier. If anyone has any ideas, you can open an additional issue suggesting those. Hopefully that answers this 😄

warmbowski:
That sounds reasonable, and I would think that 95% of the time the operation names would be unique anyway. I came across this in making a demo in which I was creating the same react component three different ways. 

But it would be great to have a warning logged to the console like "Detected duplicate operation names, only one will have types generated", or something to that effect.

ekfn:
[...]

PR #1428, opened 2019-07-21
docs: clarify apollo codegen usage

thefliik:
This PR updates the `packages/apollo/README.md` file to clarify the usage/limitations of apollo codegen. It's intended as a partial fix / quick improvement for #1427.

cc @JakeDawkins

### DIFF
--- a/packages/apollo/README.md
+++ b/packages/apollo/README.md
@@ -93,6 +93,11 @@ _See code: [src/commands/client/check.ts](https://github.com/apollographql/apoll
 
 Generate static types for GraphQL queries. Can use the published schema in Apollo Engine or a downloaded schema.
 
+Important usage notes:
+
+- Codegen requires GQL query/mutation/subscription names for a project to be unique. At the moment, if two seperate GQL queries are defined with the same name, type generation for the second query will be silently skipped.
+- Codegen does not formally support a monorepo structure. If used in a monorepo, codegen should be set up for each project separately (including seperate `apollo.config.js` files) and the `apollo client:codegen` command should be run for each project individually.
+
 ```
 USAGE
   $ apollo client:codegen [OUTPUT]


JakeDawkins:
@thefliik Thanks for this, and sorry for this, but I don't think this is the best approach for handling this for a couple reasons.

First, the section of the docs where you added your comments is autogenerated, so it would get overwritten. We don't have a way to add more than a command's description inline with the command's usage docs right now.

Secondly, the fact that codegen can be only run on a single project, not at a monorepo's root isn't unique to codegen. The whole CLI and all commands are designed currently to only operate on a single project. We do hope to change this with the next major version of the CLI though.

I think it could be useful to add an explicit note on the [configuration docs](https://www.apollographql.com/docs/references/apollo-config/), but I'm unsure.

As much as it's imperfect right now, I think the bigger win would be to adjust the warning that's produced with duplicate operation names to mention that types were not generated, rather than add these docs in here.

thefliik:
> the section of the docs where you added your comments is autogenerated

Ah. That does explain the strange formatting as well.

> As much as it's imperfect right now, I think the bigger win would be to adjust the warning that's produced with duplicate operation names to mention that types were not generated, rather than add these docs in here.

That's definitely the bigger win, but as you said over in #670.

> We're also pretty heavily backlogged on codegen issues at the moment, so I'm not sure how quickly something like this could be added/reviewed.

An imperfect docs update is better than the current situation (i.e. nothing). This being said, I'm not certain I've ever seen those configuration docs you linked to (though it seems likely I looked at them when I configured the Apollo VS Code extension). In general, I imagine someone looking for information on proper Apollo CLI usage will come here.

> but I don't think this is the best approach for handling this for a couple reasons

Correct me if I'm wrong, but I'm getting the impression that your preferred approach is "do nothing and wait for a PR addressing #1427 to be merged in" ? That's not a good solution as I imagine addressing #1427 could take quite a while.

PR #1429, opened 2019-07-21
Standardize language server errors

JakeDawkins:
This PR is the result of investigating how other implementations of the language server protocol handle errors. This PR proposes a new method of handling errors across the language server and its clients (CLI and VS Code). The intention is to produce a more friendly, consistent means by which to log and handle errors, as well as set up the basis for what could become usage metrics and error reporting, and verbose/debugging modes.

### The state of things now

- No control over the granularity of errors (full stack traces, all the time)
- No proper debug logging mechanism to help trace errors.
- No standard way to log normal messages/warnings from the language server to all clients
- Most language server implementations don't fail out. If an operation is unsuccessful, they just return an empty response, and the clients are responsible for messaging.

### Proposed Solution

At a high level, this method would eliminate the throwing of errors, and rely on errors being passed as notifications from the server to a client. For clients that don't consume the language server from a JSON-RPC transport (like the CLI which calls it directly) the server's error mechanism would be to fallback to console logs and errors. 

**The key with this proposal is to ONLY have graceful exits** from the language server. If something goes wrong, we should explain the error and do nothing. Failing out completely isn't great. Instead, clients should see empty responses from failed operations.

- `Debug` classes on the client (vscode) and server.
  - The client `Debug` class is responsible for printing errors, warnings, and info messages to the console. In CLI projects (or where `connection` is unavailable), the util will `log`/`warn`/`error` to the console
  - The server `Debug` class is responsible to notifying the client of errors via the `serverDebugMessage` notification.
  - There is a `serverDebugMessage` notification handler on the client which calls the `Debug` class methods to display errors to the user.
- How do we support stack traces for these errors?
  - By default, for Errors only (not warnings/info logs) we print the top 4 frames of the stack trace. We do this by creating a stack inside the `Debug` module and removing the unhelpful frames.
  - The server sends these frames as a string to the client, which can print it.
  - On the client, if a stack trace hasn't been provided (like if the client itself was erroring), we create a stack trace in the `Debug` module of the client.
  - I think we can support a `-v` or `--verbose` mode where errors and warnings have traces. We could also make sure those stacks aren't trimmed.
  - The `[INFO]` messages could be used for debug logs (maybe)

**Notes:**

- Inspiration from the [Crane Server](https://github.com/HvyIndustries/crane/blob/master/server/src/server.ts) implementation.
  - One of the few LSP implementations that explicitly handle errors _along with_ returning empty responses.
- This will also give us consistent hooks for printing debug info when in a `debug` mode (not implemented) and telemetry (not implemented).

**Future work:**

- [ ] the language server's `Debug` class _could_ allow for a separate error logger to be passed to it to support arbitrary clients that can't support console logs or notifications from the LSP, but since we don't have any clients for that, we shouldn't worry about that right now.

TODO:

### DIFF
--- a/packages/apollo-language-server/src/utilities/debug.ts
+++ b/packages/apollo-language-server/src/utilities/debug.ts
@@ -1,0 +1,37 @@
+import { IConnection } from "vscode-languageserver";
+
+export class Debug {
+  public static connection: IConnection;
+
+  public static SetConnection(conn: IConnection) {
+    Debug.connection = conn;
+  }
+
+  public static info(message: string) {
+    Debug.connection.sendNotification("serverDebugMessage", {
+      type: "info",
+      message: message
+    });
+  }
+
+  public static error(message: string) {
+    Debug.connection.sendNotification("serverDebugMessage", {
+      type: "error",
+      message: message
+    });
+  }
+
+  public static warning(message: string) {
+    Debug.connection.sendNotification("serverDebugMessage", {
+      type: "warning",
+      message: message
+    });
+  }
+
+  public static sendErrorTelemetry(message: string) {
+    Debug.connection.sendNotification("serverDebugMessage", {
+      type: "errorTelemetry",
+      message: message
+    });
+  }
+}
--- a/packages/apollo-language-server/src/utilities/debug.ts
+++ b/packages/apollo-language-server/src/utilities/debug.ts
@@ -1,5 +1,19 @@
 import { IConnection } from "vscode-languageserver";
 
+// for errors (and other logs in debug mode) we want to print a stack trace showing where they were thrown.
+// This uses an Error's stack trace, removes the three frames regarding this file (since they're useless) and
+// returns the rest of the trace.
+const createAndTrimStackTrace = () => {
+  let stack: string | undefined = new Error().stack;
+  // remove the lines in the stack from _this_ function and the caller (in this file) and shorten the trace
+  return stack && stack.split("\n").length > 2
+    ? stack
+        .split("\n")
+        .slice(3, 7)
+        .join("\n")
+    : stack;
+};
+
 export class Debug {
   public static connection?: IConnection;
 
@@ -17,12 +31,14 @@ export class Debug {
   }
 
   public static error(message: string) {
+    const stack = createAndTrimStackTrace();
     Debug.connection
       ? Debug.connection.sendNotification("serverDebugMessage", {
           type: "error",
-          message: message
+          message: message,
+          stack
         })
-      : console.error("[ERROR] " + message);
+      : console.error(`[ERROR] ${message}\n${stack}`);
   }
 
   public static warning(message: string) {
--- a/packages/apollo-language-server/src/utilities/debug.ts
+++ b/packages/apollo-language-server/src/utilities/debug.ts
@@ -1,37 +1,44 @@
 import { IConnection } from "vscode-languageserver";
 
 export class Debug {
-  public static connection: IConnection;
+  public static connection?: IConnection;
 
   public static SetConnection(conn: IConnection) {
     Debug.connection = conn;
   }
 
   public static info(message: string) {
-    Debug.connection.sendNotification("serverDebugMessage", {
-      type: "info",
-      message: message
-    });
+    Debug.connection
+      ? Debug.connection.sendNotification("serverDebugMessage", {
+          type: "info",
+          message: message
+        })
+      : console.log("[INFO] " + message);
   }
 
   public static error(message: string) {
-    Debug.connection.sendNotification("serverDebugMessage", {
-      type: "error",
-      message: message
-    });
+    Debug.connection
+      ? Debug.connection.sendNotification("serverDebugMessage", {
+          type: "error",
+          message: message
+        })
+      : console.error("[ERROR] " + message);
   }
 
   public static warning(message: string) {
-    Debug.connection.sendNotification("serverDebugMessage", {
-      type: "warning",
-      message: message
-    });
+    Debug.connection
+      ? Debug.connection.sendNotification("serverDebugMessage", {
+          type: "warning",
+          message: message
+        })
+      : console.warn("[WARNING] " + message);
   }
 
   public static sendErrorTelemetry(message: string) {
-    Debug.connection.sendNotification("serverDebugMessage", {
-      type: "errorTelemetry",
-      message: message
-    });
+    Debug.connection &&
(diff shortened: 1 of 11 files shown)


packages/apollo-language-server/src/utilities/debug.ts
```diff
@@ -0,0 +1,60 @@
...
+        .split("\n")
+        .slice(3, 7)
+        .join("\n")
+    : stack;
+};
+
+export class Debug {
+  public static connection?: IConnection;
+
+  public static SetConnection(conn: IConnection) {
```
JakeDawkins: using a static `connection` in this class, so we don't need to instantiate the class and pass it around anywhere. The user can just import the class itself from this file and call it without worrying like in the `workspace` file below

packages/apollo-language-server/src/workspace.ts
```diff
@@ -123,20 +124,26 @@ export class GraphQLWorkspace {
...
-          );
-
-          const existingProjects =
-            this.projectsByFolderUri.get(folder.uri) || [];
-
-          this.projectsByFolderUri.set(folder.uri, [
-            ...existingProjects,
-            ...projectsForConfig
-          ]);
+          if (config) {
```
JakeDawkins: Had to modify this file to be able to accept null configs. Since the new pattern would be to return empty responses and log errors, we'll need to do this a few places.

packages/apollo/src/Command.ts
```diff
@@ -103,6 +103,13 @@ export abstract class ProjectCommand extends Command {
     this.ctx = { flags, args } as any;
 
     const config = await this.createConfig(flags);
+
+    if (!config) {
```
JakeDawkins: Configs can now be null :)

JakeDawkins: I use `this.error` and `this.exit` here because there aren't many errors that happen in the CLI itself. Most happen in the language server. 

Errors in the CLI already have a handler (this.error) and should return with an exit code of 1. We _could_ wrap this up into another function but I don't see the need yet.

packages/vscode-apollo/src/debug.ts
```diff
@@ -0,0 +1,71 @@
+import { window, workspace, OutputChannel } from "vscode";
+
+// for errors (and other logs in debug mode) we want to print a stack trace showing where they were thrown.
+// This uses an Error's stack trace, removes the three frames regarding this file (since they're useless) and
+// returns the rest of the trace.
+const createAndTrimStackTrace = () => {
```
JakeDawkins: I think it's okay that this logic is duplicated in multiple places (language server and here) since it's only one thing, we don't have a great place to put it, and we could change the formatting based on where it is later.

packages/vscode-apollo/src/debug.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.outputConsole && this.outputConsole.appendLine(`[INFO] ${message}`);
+  }
+
+  /**
+   * Displays and error message prefixed with [ERROR]
+   * Creates and shows a truncated stack trace
+   */
+  public static error(message: string, stack?: string) {
+    const stackTrace = stack || createAndTrimStackTrace();
+    Debug.showConsole();
```
JakeDawkins: since errors are urgent, show the output window

packages/vscode-apollo/src/debug.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  public static warning(message: string, _stack?: string) {
+    this.outputConsole && this.outputConsole.appendLine(`[WARN] ${message}`);
+  }
+
+  /**
+   * TODO: enable error reporting and telemetry
+   * Displays and warning message prefixed with [WARN]
+   */
+  // public static sendErrorTelemetry(message: string) {
```
JakeDawkins: We could use this to track errors that are thrown in sentry/google analytics/something else.

packages/apollo-language-server/src/utilities/debug.ts
```diff
@@ -0,0 +1,63 @@
...
+  return stack && stack.split("\n").length > 2
+    ? stack
+        .split("\n")
+        .slice(3, 7)
+        .join("\n")
+    : stack;
+};
+
+export class Debug {
+  public static connection?: IConnection;
```
trevor-scheer: Can this be private?

packages/apollo/src/Command.ts
trevor-scheer: You've already handled the null config case within `this.createConfig()`, I think this is redundant.

packages/vscode-apollo/src/debug.ts
```diff
@@ -0,0 +1,74 @@
+import { window, workspace, OutputChannel } from "vscode";
+
+/**
+ * for errors (and other logs in debug mode) we want to print
+ * a stack trace showing where they were thrown. This uses an
+ * Error's stack trace, removes the three frames regarding
+ * this file (since they're useless) and returns the rest of the trace.
+ */
+const createAndTrimStackTrace = () => {
+  let stack: string | undefined = new Error().stack;
```
trevor-scheer: Can `stack` be undefined if you're initializing it inline like this?

packages/vscode-apollo/src/debug.ts
```diff
@@ -0,0 +1,74 @@
...
+  // remove the lines in the stack from _this_ function and the caller (in this file) and shorten the trace
+  return stack && stack.split("\n").length > 2
+    ? stack
+        .split("\n")
+        .slice(3, 7)
+        .join("\n")
+    : stack;
+};
+
+export class Debug {
```
trevor-scheer: Can we just require this class be instantiated with an `OutputChannel` so we don't have to check the existence of `this.outputConsole` everywhere?

packages/apollo/src/Command.ts
JakeDawkins: We handled the case of not having a config, but the command itself still needs to exit. It can't call `this.createService` with an empty config.

packages/vscode-apollo/src/debug.ts
JakeDawkins: Unfortunately. The type of `new Error().stack` can be undefined 😞

packages/vscode-apollo/src/debug.ts
JakeDawkins: Since we're never explicitly initializing the class, just using the static functions on it, I don't think so. I'll refactor to exit early instead

PR #1431, opened 2019-07-22
swift: Fix --operationIdsPath option

### LINKED ISSUE #1430: swift codegen - fragments not included in queries in operation IDs JSON
**Intended outcome:**

When generating a manifest of operation IDs / queries via the `--operationIdsPath` option, the JSON file should include fragment definitions.

**Actual outcome:**

Fragment definitions are missing from the queries.

See repo here (codegen command used is in `package.json`): https://github.com/darrenclark/apollo-operation-ids-issue

Diff with a fixed copy of the JSON file: https://github.com/darrenclark/apollo-operation-ids-issue/compare/fixed

**Versions**

2.16

darrenclark:
Generated file includes fragments now (fixes #1430)

TODO:

### DIFF
--- a/packages/apollo-codegen-core/src/compiler/index.ts
+++ b/packages/apollo-codegen-core/src/compiler/index.ts
@@ -66,6 +66,7 @@ export interface Operation {
   }[];
   filePath: string;
   source: string;
+  sourceWithFragments?: string;
   rootType: GraphQLObjectType;
   selectionSet: SelectionSet;
 }
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -223,12 +223,13 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
         );
 
         if (this.context.options.generateOperationIds) {
-          const { operationId } = generateOperationId(
+          const { operationId, sourceWithFragments } = generateOperationId(
             operation,
             fragments,
             fragmentsReferenced
           );
           operation.operationId = operationId;
+          operation.sourceWithFragments = sourceWithFragments;
           this.printNewlineIfNeeded();
           this.printOnNewline(
             swift`public let operationIdentifier: String? = ${SwiftSource.string(
--- a/packages/apollo/src/generate.ts
+++ b/packages/apollo/src/generate.ts
@@ -245,7 +245,7 @@ function writeOperationIdsMap(context: CompilerContext) {
     .forEach(operation => {
       operationIdsMap[operation.operationId!] = {
         name: operation.operationName,
-        source: operation.source
+        source: operation.sourceWithFragments!
       };
     });
   fs.writeFileSync(


designatednerd:
Good catch! Is there a way to update the tests so that we can make sure we don't accidentally drop this in the future? You may need to ping @trevor-scheer, I'm a bit shit at typescript.

designatednerd:
@trevor-scheer Any help you could give making heads or tails of the CI failures would be lovely as well

packages/apollo-codegen-core/src/compiler/index.ts
```diff
@@ -63,6 +63,7 @@ export interface Operation {
   }[];
   filePath: string;
   source: string;
+  sourceWithFragments?: string;
```
trevor-scheer: Is there a reason for this to be a `?`. It seems to always be a string when returned from `generateOperationId()`. If we can remove the `?` here, then we can also clean up the `!` below.

trevor-scheer:
One comment that I'd like to have addressed, and +1 for adding tests please!

CI issues are not going to be addressed in the short term, but if everything but Azure Node 10 is passing then we're good. The others sometimes just need a rerun 😞

packages/apollo-codegen-core/src/compiler/index.ts
darrenclark: If I remember correctly, I made it optional because the `sourceWithFragments` only gets added when `--operationIdsPath` is specified.

Similar to the `operationId` property in this interface:

https://github.com/apollographql/apollo-tooling/blob/0735a1f840be41ea207fcf394b150270e5143ff1/packages/apollo-codegen-core/src/compiler/index.ts#L55-L59

darrenclark:
@designatednerd @trevor-scheer For sure! Will add tests.

I'm not too familiar with the code base, any recommendations on where to add tests?  Was thinking of adding another test case here: https://github.com/apollographql/apollo-tooling/blob/0735a1f840be41ea207fcf394b150270e5143ff1/packages/apollo/src/commands/client/__tests__/generate.test.ts#L112-L136

trevor-scheer:
@darrenclark that looks like a fine place to add a test. A couple comments:
* If this change fixes some behavior w.r.t this flag, should we have expected a change in the snapshot belonging to the test you pointed to (that uses that flag)?
* IIRC, these codegen tests can get flaky in CI (specifically, adding new tests to this file seems to be troublesome 😞). If your new test(s) pass locally but you see some weirdness in CI, just let me know.

Thanks!

zionts:
@darrenclark @trevor-scheer @designatednerd looks like some great conversation was had here! Is there a plan to take this on to the finish line or should we close it out for now?

manicmaniac:
@darrenclark  @designatednerd @trevor-scheer @zionts 

I want this feature to use persisted query in our product.
Can I take over the rest tasks? Or should I open another PR?

designatednerd:
I'm good with a takeover if @darrenclark is

darrenclark:
Ohh yea, apologies, I had forgot about this PR 😅

@manicmaniac Go for it, thanks! 👍

manicmaniac:
I'm trying to add tests in https://github.com/apollographql/apollo-tooling/blob/a3b261ab08ea82781b0f2e307c84456c299fda25/packages/apollo/src/commands/client/__tests__/generate.test.ts#L112-L136 as mentioned before.

But I ran into another issue where generate.test.ts doesn't pass because of https://github.com/apollographql/apollo-tooling/issues/1853.
(Although the author of the issue says that `--runInBand` suppresses this error, it doesn't work in my environment.)

So it seems I have to resolve https://github.com/apollographql/apollo-tooling/issues/1853 before addressing this error.

manicmaniac:
I just created another PR and cherry-picked the commit in this PR.

https://github.com/apollographql/apollo-tooling/pull/2017

darrenclark:
@manicmaniac 👍   I'll close this PR in favour of yours.  Thanks for taking this over!

PR #1432, opened 2019-07-22
Initial Implementation of apollo service:codegen

JacksonKearl:
Adding `apollo service:codegen`!

This gives the user well-typed `context`, `arguments`, and `parent` in resolver functions, and provides support for all the federation directives:

`@key`:
- Generate types for a  __resolveReference 
- Split the object into its representation and its other components. Representation gets provided as `parent` to resolvers.

`@external`:
- Disallow providing resolvers for @external fields, except for...

`@provides`:
- Allow `@provide`’d fields to be on resolvers

`@requires`:
- Add a field's `@require`’d keys to the `parent` passed to that field's resolver.
 
For a full project using Federation and codegen, check out my [typed federation-demo](https://github.com/JacksonKearl/federation-demo/tree/types)
For a full project using Context and codegen, check out my [typed fullstack-tutorial](https://github.com/apollographql/fullstack-tutorial/tree/types)


JacksonKearl:
Pretty sure those errors are unrelated?

JacksonKearl:
Currently TS is generated by just string concatenation. We could instead do it with some sort of fancy builder library, but I thought this was simple enough to not justify bringing in a dependency for that. Thoughts?

packages/apollo-server-codegen/src/IR/Types.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  public translate(translator: Translator) {
+    return translator.translateCompoundType(this);
+  }
+}
+
+export const makeType = (sdlVal: TypeNode): TypeDefinition => {
+  switch (sdlVal.kind) {
+    case "NamedType":
```
trevor-scheer: Can we use the `Kind` object for these cases as well?

JacksonKearl:
Things to add:
- [ ] Input Types
- [ ] interfaces and `implements`
- [ ] Unions

packages/apollo-server-codegen/src/IR/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  ScalarDefinition
+};
+
+export function sdlToIR(
+  sdl: DocumentNode
+): [ObjectDefinition[], EnumDefinition[], ScalarDefinition[]] {
+  const objectDefinitions: TypelessObjectDefinition[] = [];
+  const enumDefinitions = [];
+  const scalarDefinitions = [];
+  for (const definition of sdl.definitions) {
```
trevor-scheer: Personal preference, but this seems like a nice use case for `graphql`'s `visit()` fn. Just recommending it in case you're unfamiliar.

Here's one example, but it's littered throughout federation / AS code. It's great for traversing SDL (and modifying / returning new SDL if you choose!).

https://github.com/apollographql/apollo-server/blob/42ab6b86b7fb3adcdb5f3fd1036cfa2b7a524803/packages/apollo-federation/src/composition/normalize.ts#L132

packages/apollo-server-codegen/src/IR/index.ts
```diff
@@ -0,0 +1,75 @@
...
+      default:
+        console.log(
+          `Ignoring value of type ${definition.kind}: ${JSON.stringify(
+            definition
+          )}`
+        );
+    }
+  }
+
+  const flatMap = <T, E>(
```
trevor-scheer: You may consider importing `apollo-env` (which polyfills `Array.flatMap()`)

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
```diff
@@ -0,0 +1,210 @@
...
+    ].join("\n");
+  }
+
+  private generateHeader() {
+    // Generate some utility higher order types that we'll reference later on.
+    return `// This is a machine generated file.
+// Use "apollo service:codegen" to regenerate.
+type PromiseOrValue<T> = Promise<T> | T
+type Nullable<T> = T | null | undefined
+type Index<Map extends Record<string, any>, Key extends string, IfMissing> = Map[Key] extends object ? Map[Key] : IfMissing
```
trevor-scheer: This is neat!

packages/apollo-server-codegen/src/Translators/index.ts
```diff
@@ -0,0 +1,108 @@
...
+        .map(v => v.translate(t))
+        .filter(t => t !== undefined)
+        .join("\n")
+    : v.translate(t) || ""
+  )
+    .split("\n")
+    .join("\n" + "  ");
+
+import { TypeScriptTranslator } from "./TypeScriptTranslator";
+export { TypeScriptTranslator };
```
trevor-scheer: Nit: I think we can just
`export { TypeScriptTranslator } from "./TypeScriptTranslator";`

packages/apollo-server-codegen/src/Translators/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  public abstract translateScalarDefinition(t: IR.ScalarDefinition): string;
+}
+
+// All IR components are Translatable
+export interface Translatable {
+  translate(t: Translator): string | undefined;
+}
+
+// Helper for indenting while translating and one or many Translatable's
+export const translateAndIndent = (
```
trevor-scheer: Spoke offline about offloading whitespace / indentation work to `prettier` 🎉
PS: thanks for `recast` @benjamn 😄

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,138 @@
...
+  inStock: Boolean
+  shippingEstimate: Int @requires(fields: "price weight")
+}
+```
+
+```ts
+Product: {
+  // parent is {upc: string}
+  inStock(parent) {...}
+  // parent is {upc: string, price: number, weight: number}
```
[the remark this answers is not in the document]
trevor-scheer: Nice, glad we were on the same page with `Nullable<T>`

packages/apollo/src/commands/service/codegen.ts
```diff
@@ -0,0 +1,129 @@
...
+        if (err) throw new Error(JSON.stringify(err));
+        resolve(data.toString());
+      })
+    );
+    const inputFileType = extname(path);
+    const getSDL = () => {
+      const type = fileTypesMapping[inputFileType];
+      switch (type) {
+        case "javascript":
+          const getInnerSDL = /(?:(?:gql|graphql)`)((?:[^`\\]|\\`|\\n|\\r|\\\\)*)`/;
```
trevor-scheer: I believe this is the approach our codegen libraries use now, but it would be great to start using ASTs for this. If you're not opposed to making this change, I'd be happy to help - though I haven't done it myself (but have wanted to!)

@benjamn will know the most. AST via `recast` and `visit()` from `ast-types`?

trevor-scheer:
This is incredible, thanks for your hard work on this (and the wonderful PR)! Can't wait for the blog post 😉

packages/apollo/src/commands/service/codegen.ts
benjamn: The following should work, though I strongly recommend typing it in yourself, so you can see how much help TypeScript is providing:
```js
import { parse } from "recast/parsers/babel";
import { visit, types } from "recast";

const { Identifier } = types.namedTypes;
const inputText = "gql`query { whatever }`";
const operations: string[] = [];

visit(parse(inputText), {
  visitTaggedTemplateExpression(path) {
    this.traverse(path);
    const expr = path.node;
    if (
      Identifier.check(expr.tag) &&
      expr.tag.name === "gql"
    ) {
      expr.quasi.quasis.forEach(q => {
        operations.push(q.value.raw);
      });
    }
  },
});
```
I don't know how anyone writes AST transformations without auto-generated AST type information.

packages/apollo-server-codegen/README.md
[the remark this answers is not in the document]
JacksonKearl: Updated the examples to use nonnullable types and added a section on nullability.

packages/apollo-server-codegen/src/IR/Objects.ts
```diff
@@ -0,0 +1,113 @@
...
+export class TypelessObjectDefinition {
+  public resolvers: TypelessResolverDefinition[];
+  public fields: FieldDefinition[];
+  public name: string;
+
+  constructor(
+    private definition: ObjectTypeDefinitionNode | ObjectTypeExtensionNode
+  ) {
+    this.name = definition.name.value;
+    const isQueryOrMutation = this.name === "Query" || this.name === "Mutation";
```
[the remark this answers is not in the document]
JacksonKearl: Forgot about those lol

packages/apollo-server-codegen/src/IR/Objects.ts
```diff
@@ -0,0 +1,113 @@
...
+  public isTypeExtension: boolean;
+
+  constructor(
+    definition: ObjectTypeDefinitionNode | ObjectTypeExtensionNode,
+    typeless: TypelessObjectDefinition,
+    types: TypelessObjectDefinition[],
+    provided: string[]
+  ) {
+    this.name = definition.name.value;
+    this.isTypeExtension = definition.kind === "ObjectTypeExtension";
```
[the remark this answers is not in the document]
JacksonKearl: `isTypeExtensionNode` works 👍

packages/apollo-server-codegen/src/IR/Resolvers.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  constructor(
+    typeless: TypelessResolverDefinition,
+    types: TypelessObjectDefinition[],
+    public provided: boolean
+  ) {
+    this.arguments = (typeless.fieldDefinition.arguments || []).map(
+      argumentDefinition => new ArgumentDefinition(argumentDefinition)
+    );
+    this.external = !!(typeless.fieldDefinition.directives || []).find(
```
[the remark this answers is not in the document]
JacksonKearl: Moved to `Array#some` and renamed the bools 👍

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
JacksonKearl: 🤙

packages/apollo-server-codegen/src/Translators/index.ts
JacksonKearl: Weird, I thought I tried that and it gave me a syntax error but seems to work now.

packages/apollo-server-codegen/src/IR/Types.ts
JacksonKearl: Done!

packages/apollo-server-codegen/src/IR/index.ts
JacksonKearl: Saw that, but decided to not import the whole thing just for a polyfill if that's good by you.

trevor-scheer: Yep, fine by me 👍

packages/apollo/src/commands/service/codegen.ts
JacksonKearl: Got it, thanks!

packages/apollo-server-codegen/src/IR/index.ts
JacksonKearl: Going to keep it as is for now, as to me this more explicitly shows that we're only interested in top-level definitions (and that we're treating certian classes of definitions (type definitions, type extensions, interface definition, etc.) exactly the same.

packages/apollo-server-codegen/src/IR/index.ts
JacksonKearl: Actually went back and decided to just go for the import lol

packages/apollo-server-codegen/src/IR/Resolvers.ts
```diff
@@ -0,0 +1,127 @@
...
+      (this.fieldDefinition.directives || [])
+        .filter(
+          directive =>
+            directive.name.value === "provides" &&
+            directive.arguments &&
+            directive.arguments.length
+        )
+        .map(key => (key.arguments![0].value as StringValueNode).value),
+      providedFields =>
+        providedFields.split(" ").map(field => ({
```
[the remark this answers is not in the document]
JacksonKearl: Yep had to change it a bit, but works now!

packages/apollo-server-codegen/src/IR/Objects.ts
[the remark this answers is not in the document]
JacksonKearl: Done!

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,152 @@
...
+extend type User @key(fields: ID) {
+  id: ID @external
+  name: String @external
+}
+```
+
+```ts
+{
+  User: {
+    id() {...}  // Error
```
jbaxleyiii: I don't think this is the correct behavior here. Give a query like this:

```graphql
{
  me {
    reviews {
      id
    }
  }
}
```

Where Review.id is an external field owned by the reviews service, the accounts service needs to be able to provide the `id` field so that the reviews service can make the join. This may mean mapping from `Review.internalID` value in the users service using a custom resolver

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,152 @@
...
+  shippingEstimate: Int @requires(fields: "price weight")
+}
+```
+
+```ts
+Product: {
+  // parent is {upc: string}
+  inStock(parent) {...}
+  // parent is {upc: string, price: number, weight: number}
+  shippingEstimate(parent) {...}
```
jbaxleyiii: 👏 👏 👏

packages/apollo-server-codegen/src/IR/Descriptions.ts
```diff
@@ -0,0 +1,13 @@
+import { Translatable, Translator } from "../Translators";
+
+export class Description implements Translatable {
```
jbaxleyiii: As a general comment for this PR, I think it would be incredibly valuable to add some documentation to these files signalling their intention. Take for example the interaction between `Translator` and `Description`, I'm not sure on initial read what that relationship gains. For future improvements and care of this addition I think we need more explanation as to the structure and relationship between the types.

packages/apollo-server-codegen/src/IR/Descriptions.ts
```diff
@@ -0,0 +1,13 @@
...
+  public description: string;
+  constructor(node: { description?: { value: string } }) {
```
jbaxleyiii: Is this an ASTNode? Can we use actual ASTNode typings here?

packages/apollo-server-codegen/src/IR/Enums.ts
```diff
@@ -0,0 +1,18 @@
+import { EnumTypeDefinitionNode } from "graphql";
+import { Translatable, Translator } from "../Translators";
+import { Description } from "./Descriptions";
+
+export class EnumDefinition implements Translatable {
```
jbaxleyiii: Should `Translatable` take an ASTNode type generic for the types it supports? In this case `Translatable<EnumTypeDefinitionNode>`

packages/apollo-server-codegen/src/IR/Objects.ts
```diff
@@ -0,0 +1,159 @@
...
+  InterfaceTypeExtensionNode,
+  Kind
+} from "graphql";
+
+import { Translatable, Translator } from "../Translators";
+import { Description } from "./Descriptions";
+import { ResolverDefinition, TypelessResolverDefinition } from "./Resolvers";
+import { CompoundType, makeType, TypeDefinition } from "./Types";
+
+export const SELECTION_OFFSET = 7; // number of chars added in the below parse(`query { ${source} }`) call
```
jbaxleyiii: Can you expand on this some more in docs? I'm not sure I follow the need of it.

packages/apollo-server-codegen/src/IR/Objects.ts
```diff
@@ -0,0 +1,159 @@
...
+const parseSelections = (source: string) =>
+  (parse(`query { ${source} }`).definitions[0] as OperationDefinitionNode)
```
jbaxleyiii: This should probably become a shared utility for parsing a selectionset directly https://github.com/apollographql/apollo-server/blob/master/packages/apollo-gateway/src/utilities/graphql.ts#L100

packages/apollo-server-codegen/src/IR/Resolvers.ts
```diff
@@ -0,0 +1,168 @@
...
+import { TypelessObjectDefinition } from "./Objects";
+import {
+  CompoundType,
+  findRootType,
+  makeType,
+  TypeDefinition,
+  NonNullType
+} from "./Types";
+
+const parseSelections = (source: string) =>
```
jbaxleyiii: Let's not copy and paste this around, but instead make it a common utility in the `apollo-graphql` package https://github.com/apollographql/apollo-server/blob/master/packages/apollo-gateway/src/utilities/graphql.ts#L100

packages/apollo-server-codegen/src/IR/Resolvers.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  public getProvides(types: TypelessObjectDefinition[], errors: string[]) {
+    const provideDirectives = (this.fieldDefinition.directives || []).filter(
+      directive =>
+        directive.name.value === "provides" &&
+        directive.arguments &&
+        directive.arguments.length
+    );
+
+    const providedSelections = provideDirectives
```
jbaxleyiii: This seems like we should tighten up the type definitions to avoid the casting below this line

packages/apollo-server-codegen/src/IR/Resolvers.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    if (providedSelections.length === 0) return [];
+
+    const fieldType = findRootType(makeType(this.fieldDefinition.type));
+    const providedObject = types.find(type => type.name === fieldType);
+
+    if (!providedObject) {
+      const loc = this.fieldDefinition.type.loc;
+      errors.push(
+        `(${loc!.start},${loc!.end}) Could not find type ${fieldType}.`
```
jbaxleyiii: The error format is for the inline vscode errors correct? Can we turn this into a `makeVSCodeError` or something to ensure this doesn't get changed later?

packages/apollo-server-codegen/src/IR/Types.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export class ListType implements TypeDefinition {
+  constructor(public base: TypeDefinition) {}
+
+  public translate(translator: Translator, nullable: boolean = true) {
+    return translator.translateListType(this, nullable);
+  }
+}
+
+export const findRootType = (t: TypeDefinition): string => {
```
jbaxleyiii: Hmm, I thought this was already a utility in either our code base or in graphql-js. I can't seem to find it off hand but maybe @JakeDawkins or @martijnwalraven remembers where it is?

One thing to note, we try to not use `instanceof` because it can break across duplicate module boundaries even if the package is of the same version. We are going to be changing this in graphql-js at some point, but for the time being, can we use either duck typing or predicate functions for these utilities to minimize the spreading of `instanceof` in the codebase?

packages/apollo-server-codegen/src/IR/Types.ts
```diff
@@ -0,0 +1,123 @@
...
+      const field = baseType.fields.find(
+        field => field.name === node.name.value
+      );
+
+      const startLoc = node.loc!.start - SELECTION_OFFSET + errorLocation[0];
+      const endLoc = node.loc!.end - SELECTION_OFFSET + errorLocation[0];
+      const adjustedErrorLocation = [startLoc, endLoc];
+      if (!field) {
+        errors.push(
+          `(${adjustedErrorLocation}) Could not find field "${node.name.value}" on type "${baseType.name}".`
```
jbaxleyiii: Same comment about a `makeVSCodeError` call to ensure this format is preserved

jbaxleyiii: cc @JakeDawkins we could be using this in more places

packages/apollo-server-codegen/src/IR/Types.ts
```diff
@@ -0,0 +1,123 @@
...
+    });
+  }
+
+  public translate(translator: Translator) {
+    return translator.translateCompoundType(this);
+  }
+}
+
+export const makeType = (sdlVal: TypeNode): TypeDefinition => {
+  switch (sdlVal.kind) {
```
jbaxleyiii: This isn't exhaustive, what happens in the default case here?

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
```diff
@@ -0,0 +1,364 @@
...
+      ...[...objects, ...enums, ...scalars, ...unions, ...inputObjects].map(
+        def => def.translate(this)
+      )
+    ].join("\n");
+  }
+
+  private generateHeader() {
+    // Generate some utility higher order types that we'll reference later on.
+    return `// This is a machine generated file.
+// Use "apollo service:codegen" to regenerate.
```
jbaxleyiii: can we avoid printing these in every file? Will this pollute other types?

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  public translateDescription(t: IR.Description): string {
+    return "/**\n * " + t.description.split("\n").join("\n * ") + "\n */\n";
+  }
+
+  public translateNamedType(t: IR.NamedType, nullable: boolean): string {
+    const getName = () => {
+      switch (t.name) {
+        case "Int":
```
jbaxleyiii: Let's add `Float` in here as well

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
```diff
@@ -0,0 +1,364 @@
...
+      t.name,
+      // Bit of hack here to use optional types instead of Nullable, for better compatibility with
+      // destructuring default assignment: ({arg = default}) => ...
+      t.type instanceof IR.NonNullType ? `: ` : `?: `,
+      new IR.NonNullType(t.type).translate(this)
+    ].join("");
+  }
+
+  public translateResolverDefinition(t: IR.ResolverDefinition): string {
+    if (t.isNotProvidedAndExternal) {
```
jbaxleyiii: This is mentioned above, but I don't think this is the correct behavior

.vscode/settings.json
```diff
@@ -13,5 +13,8 @@
     "lib": true
   },
   "typescript.tsdk": "node_modules/typescript/lib",
-  "debug.node.autoAttach": "on"
+  "debug.node.autoAttach": "on",
+  "cSpell.words": [
+    "codegen"
```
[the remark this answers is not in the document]
JacksonKearl: cSpell! It's a linter for spelling mistakes that knows about camelCase. Do you this it not tracked?

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
JacksonKearl: There is only one generated file, and no, type definitions in a module won't impact other modules without first explicitly importing them in that module.

packages/apollo-server-codegen/README.md
JacksonKearl: I thought I allowed fields used in `key`s to be provided always, but I might have missed it. 

Would that be okay, or should I just make it so everything can be provided?

.vscode/settings.json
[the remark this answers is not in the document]
jbaxleyiii: oh no, I think it is awesome! Just didn't know what it was :)

packages/apollo-server-codegen/src/IR/Descriptions.ts
JacksonKearl: It's not an ASTNode, as not all nodes can have descriptions (extension definition nodes, name nodes, etc)

packages/apollo-server-codegen/src/IR/Enums.ts
JacksonKearl: We could add a type parameter like that, but what would it actually do?

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
JacksonKearl: Done!

packages/apollo-server-codegen/src/IR/Types.ts
JacksonKearl: It is exhaustive, `TypeNode` is 
```ts
export type TypeNode = NamedTypeNode | ListTypeNode | NonNullTypeNode;
```

packages/apollo-server-codegen/src/IR/Objects.ts
JacksonKearl: Added more docs

packages/apollo-server-codegen/src/IR/Objects.ts
JacksonKearl: I made it into a shared utility in this package, but not in `apollo-graphql` as per your other comment. Resolving this comment for now.

packages/apollo-server-codegen/src/IR/Resolvers.ts
JacksonKearl: Would that mean adding it to apollo-graphql and concurrently updating the dependency in apollo-server and using it from there too? Or just add it to apollo-graphql for now and worry about apollo-server later?

packages/apollo-server-codegen/src/IR/Resolvers.ts
JacksonKearl: Got this for free when moving over to the shared util implementation

packages/apollo-server-codegen/src/IR/Resolvers.ts
JacksonKearl: Done!

packages/apollo-server-codegen/src/IR/Types.ts
JacksonKearl: Got rid of the instance of checks. This is specifically finding the root type for my IR types, so I don't think you'd have it anywhere else.

packages/apollo-server-codegen/README.md
JacksonKearl: Using keys and provides to determine what cant be resolved now.

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
JacksonKearl: Using keys and provides now

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,152 @@
...
+type User @key(fields: "id") {
+  id: ID!
+  name: String!
+}
+```
+
+```ts
+{
+  Review: {
+    // `representation` is type `{timestamp: number, author: {id: number}} | {id: string}`
```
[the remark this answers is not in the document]
JacksonKearl: would it be `{__typename: 'Review'}`? That doesn't seem all that helpful on serverside...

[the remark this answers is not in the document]
JacksonKearl: And trying to compose a graph that uses interfaces as entities fails currently. I don't know what the desired behavior is there.

packages/apollo-server-codegen/src/IR/Descriptions.ts
JacksonKearl: I added some docs and a architecture.md document, lmk what you think!

JacksonKearl:
@jbaxleyiii added architecture.md and left some replied to your comments

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,152 @@
...
+const resolvers: Resolvers<TContext, any> {
+  User: {
+    name({ foo, bar }) => ... // this is allowed. `foo` and `bar` will be type `any`.
+  }
+}
+```
+
+### Scalars
+
+Scalars are simply converted to type `any`, though if there is sufficient community demand, these could be specified in a manner similar to the `TInternalReps` strategy above.
```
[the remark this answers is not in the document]
JacksonKearl: added custom scalar and internal enum value support.

packages/apollo-server-codegen/src/IR/Enums.ts
trevor-scheer: I'm not sure the generic makes sense here, as there a some non-ASTNode things that also implement `Translatable`.

packages/apollo-server-codegen/src/IR/Resolvers.ts
trevor-scheer: The latter is fine, concurrent isn't important - let's just add a note somewhere via a WIP PR or a JIRA ticket to make the switch. This would also make for a nice, small PR of its own that we could land immediately if you'd like.

packages/apollo-server-codegen/README.md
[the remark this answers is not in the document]
trevor-scheer: I think these may provide the context that @jbaxleyiii is getting at:
https://github.com/dotansimha/graphql-code-generator/issues/1633
https://www.apollographql.com/docs/apollo-server/features/unions-interfaces/

Unions and interfaces use a `__resolveType()` resolver, which defaults to looking at `__typename`. I'm not sure if @jbaxleyiii meant `__resolveReference` or `__resolveType` here.

Sidenote: that's unfortunate, I think that's something that should be working. Is reproducing as straightforward as it sounds / if you already have one could you send it to me please 😄 (if not I'll just try to repro in a test case)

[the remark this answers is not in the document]
JacksonKearl: Yeah if you put an `@key` on an interface one of the validators throws I believe.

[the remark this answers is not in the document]
JacksonKearl: And I do implement `__resolveType` now, but I didn't at time of initial review I beleive

packages/apollo-server-codegen/ARCHITECTURE.md
```diff
@@ -0,0 +1,67 @@
+# Architecture
+
+## Overview
+
+The code generator is a set of modules designed to accept a GraphQL schema, and from it generate type definitions in TypeScript format (other languages may be supported in the future).
+
+Similar to projects like [graphql code generator](https://graphql-code-generator.com), the generated type definitions describe both the type of the GraphQL objects represented in the schema, and the type of the `resolvers` function that an Apollo Server serving would require to serve the schema.
```
trevor-scheer:
```suggestion
Similar to projects like [graphql code generator](https://graphql-code-generator.com), the generated type definitions describe both the type of the GraphQL objects represented in the schema, and the type of the `resolvers` function that an Apollo Server would require to serve the schema.
```

packages/apollo-server-codegen/ARCHITECTURE.md
```diff
@@ -0,0 +1,67 @@
...
+
+However, this project expands upon existing efforts by providing accurate type definitions for federated schemas, built using the recently announced Apollo Federation platform.
```
trevor-scheer: Things like "recently announced" don't age well in docs, but I do think it's right to nod to federation here.

packages/apollo-server-codegen/ARCHITECTURE.md
```diff
@@ -0,0 +1,67 @@
...
+
+This package serves only as an API, to interface with it we suggest either the `Apollo CLI`, where it is available as `apollo server:codegen`, or the `[.....]` plugin we have built for `graphql-code-generator`.
```
trevor-scheer: > `[.....]`
TBD? We should probably leave this out unless this thing is built / being built?

packages/apollo-server-codegen/ARCHITECTURE.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Design
+
+The generator is built using a two-layered approach, where the frontend is able to convert an SDL-formatted input string to an internal representation (`IR`), and one or backends are able to convert that `IR` to files of type declaration source code in a particular programming language. Currently only a backend for TypeScript is provided.
```
trevor-scheer: I'd like to suggest a different nomenclature here, since frontend/backend have pretty well-defined meanings already. I'd remap backend -> consumer, but not sure what to remap frontend to. Something related to normalization, translation, or parsing?

packages/apollo-server-codegen/ARCHITECTURE.md
```diff
@@ -0,0 +1,67 @@
...
+  public translateDescription(t: IR.Description): string {
+    return [
+      "/**",
+      ...t.description.split("\n").map(line => `* ${line}`),
+      " */"
+    ].join("\n");
+  }
+```
+
+Additionally, every node in the `IR` implements [`Translateable`](Translators/index.ts):
```
trevor-scheer:
```suggestion
Additionally, every node in the `IR` implements [`Translatable`](Translators/index.ts):
```

packages/apollo-server-codegen/ARCHITECTURE.md
```diff
@@ -0,0 +1,67 @@
...
+
+```ts
+export interface Translatable {
+  translate(t: Translator): string | undefined;
+}
+```
+
+This allows a `Translator`'s translation to recursively descend into the `IR` tree by simply calling `.translate(this)` on each IR node, and the node will handle delegation back to the `Translator`.
```
trevor-scheer: Can we rephrase? I don't follow what this means:
> and the node will handle delegation back to the `Translator`

packages/apollo-server-codegen/ARCHITECTURE.md
JacksonKearl: This is consistent with the nomenclature compilers use, so I think its reasonable to use it here in ARCHITECTURE.md, but maybe not in something like README.md.

packages/apollo-server-codegen/ARCHITECTURE.md
JacksonKearl: removed it, turns out the plugin is something like 4 lines to wire up, so I dont see much of a point in us releasing it.

packages/apollo-server-codegen/ARCHITECTURE.md
JacksonKearl: removed "recently announced"

packages/apollo-server-codegen/ARCHITECTURE.md
JacksonKearl: Yeah I think its fine to just remove that part.

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
```diff
@@ -0,0 +1,296 @@
...
+    return [
+      `export interface Resolvers<TOptions extends OptionTypes = {}> {`,
+      ...types.map(type =>
+        this.rootTypes.includes(type)
+          ? `${type}: ${type}Resolver<TOptions>`
+          : `${type}?: ${type}Resolver<TOptions>`
+      ),
+      ...scalars.map(scalar => `${scalar}: any`),
+      ...enums.map(
+        _enum =>
```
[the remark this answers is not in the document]
JacksonKearl: enum is a reserved word.

packages/apollo-server-codegen/src/Translators/TypeScriptTranslator.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  public translateDescription(t: IR.Description): string {
+    return "/**\n * " + t.description.split("\n").join("\n * ") + "\n */\n";
+  }
+
+  public translateNamedType(t: IR.NamedType, nullable: boolean): string {
+    let name = (() => {
+      switch (t.name) {
+        case "Int":
+          return "number";
```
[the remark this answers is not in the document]
JacksonKearl: Done

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,222 @@
...
+```gql
+scalar JWT
+
+type Query {
+  me(token: JWT): User
+}
+```
+
+```ts
+const resolvers: Resolvers<{Scalars: {JWT: {header: {alg: string, type: string}}}> = {
```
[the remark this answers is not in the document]
JacksonKearl: One would probably declare `type MyResolvers = Resolvers<...>` if they're going to be using the `Resolvers` type in multiple places.

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,222 @@
...
+  BLUE
+}
+
+type Query {
+  favoriteColor: AllowedColor
+}
+```
+
+```ts
+const resolvers: Resolvers<{Enums: {AllowedColor: '#f00' | '#0f0' | '#00f'}> = {
```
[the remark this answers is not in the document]
JacksonKearl: I don't know what you mean by "part of configuration". To me, the type parameters of `Resolvers` are the configuration. If a person wanted to reuse the same type in multiple places, I imagine they'd create an alias for the configured type (`type MyResolvers = Resolvers<....>`), much the same as they'd do for any other thing they're reusing in multiple places.

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+```ts
+const resolvers: Resolvers<{InternalReps: { User: {internalID: number} }}> {
+  User: {
+    name({ internalID }) => ... // id will be of type `number`
+  }
+}
+```
+
+Alternatively, if the user would like more type freedom, they can pass `any` to the second type parameter of the emitted `Resolvers` definition, and the resolvers will use `any` as the type of their `parent` argument:
```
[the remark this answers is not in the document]
JacksonKearl: Fixed

packages/apollo-server-codegen/README.md
```diff
@@ -0,0 +1,222 @@
...
+}
+```
+
+> You will still need to create your own implementation of the actual [`GraphQlScalarType`](https://graphql.org/graphql-js/type/#graphqlscalartype) object.
+
+If no declaration for a scalar type is provided in the options, it will default to `unknown`.
+
+### Enums
+
+Enums are by default converted to a basic `or` of all their type options. If you are using [Internal Enum Values](), you can specify the possible internal values in the `Enums` field of the options object:
```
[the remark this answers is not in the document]
JacksonKearl: fixed

packages/apollo-server-codegen/src/IR/Resolvers.ts
JacksonKearl: Learna is absolutely refusing to make "apollo-graphql" available within "apollo-server-codegen". I've tried everything, including manually writing the package-lock, and it just won't install it. Any tips are appreciated. @abernix in particular.

packages/apollo-server-codegen/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.base",
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./lib"
+  },
+  "include": ["src/**/*"],
+  "exclude": ["**/__tests__/*"]
```
trevor-scheer: I believe you'll need a `references` key here for the `apollo-graphql` dependency.
`"references": [{ "path": "../apollo-graphql" }]`

packages/apollo/package.json
```diff
@@ -57,15 +58,19 @@
...
     "lodash.pickby": "4.6.0",
     "moment": "2.24.0",
+    "prettier": "1.18.2",
+    "recast": "0.18.1",
     "strip-ansi": "5.2.0",
     "tty": "1.0.1",
     "vscode-uri": "1.0.6"
   },
   "devDependencies": {
+    "@types/prettier": "^1.18.0",
```
trevor-scheer: IIRC, `typescript` is the exception to the rule (though I don't remember why). All other `devDependencies` should be declared in the monorepo's root `package.json`.

JacksonKearl: moved it

packages/apollo-server-codegen/tsconfig.json
JacksonKearl: added!

packages/apollo-server-codegen/README.md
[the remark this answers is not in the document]
JacksonKearl: resolved offline and added docs

packages/apollo-server-codegen/ARCHITECTURE.md
```diff
@@ -0,0 +1,67 @@
...
+
+```ts
+type ReviewRepresentation = { creationDate: string; author: { id: string } };
+```
+
+to be created requires knowledge of the `Author` type, in particular that it's `id` field is type `string`.
+
+To solve this problem, the `IR` generator, [`getSDLFromIR`](/IR/index.md), works in two phases: it first compiles the SDL into a "typeless" representation, where the only basic type knowledge needed for the later construction of federation-aware types is created. Then, it injects that basic global type knowledge back into itself, to form the final, federation-aware `IR`.
+
+> Typically, `buildSchema` from `graphql-js` could be used to get the general type information, however in the case of Federation the SDL may be ill formed for `buildSchema` (undefined directives, extended classes that aren't otherwise defined, etc.), so that is not possible in this case.
```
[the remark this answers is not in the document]
JacksonKearl: I feel good about the current implementation, but wouldn't be neccesarily opposed to a possible rewrite to "`buildSchema` + hacks" in the future.

packages/apollo-server-codegen/src/IR/Resolvers.ts
JacksonKearl: Resolved.

zionts:
@trevor-scheer @JakeDawkins , looks like a lot of great work was put into this, but I also know that we're doing a large refactor of the CLI right now. I'm going to close this out for now with our intention being to use this PR for inspiration down the line when we're ready to take this on, but feel free to reopen if I'm misinterpreting our intentions here!

PR #1433, opened 2019-07-23
Add client schema support to vscode extension

### LINKED ISSUE #847: Add @client directive support in VS Code
I am trying to set up IDE to ease GraphQL development for our junior team members. And not gonna lie, to help myself too.

I have a query that operates on client state:

```
import gql from 'graphql-tag';

export default gql`
mutation LoadProfile($profile: ProfileInput!) {
  loadProfile(profile: $profile) @client {
    accessToken
    avatar
    name
  }
}
`
```

Though, removing the `@client` does not invalidate the query.

image

Is this expected behaviour or I have misconfigured something or it's a bug?

---

Otherwise, everything seem to be working - autocompletion for both client queries and remote queries, go-to definition, validation, for instance:

image

Just that dropping `@client` from client queries doesn't throw anything.

jasonpaulos:
Specifically,
* Augment the client schema with metadata about which fields/objects
are client-only.
* Show information about client-only fields on hover.
* Add auto complete support for injecting client directives when
appropriate.
* Add a validation rule to detect missing client directives.
* Add a code action to suggest the addition of a clienet directive
if it is missing. Also add the ability in general for validation
rules to do this.
* Add go to definition support for custom client-only directives.
* Add information on hover for directives.
* Add auto complete support for directives.
* Forbid unknown directives.

---

Since this update forbids unknown directives from being used, libraries like `apollo-link-rest` that provide their own client directives used for links will have to make the extension aware of their directives. This can be done like so (assuming the package comes with a `schema.graphql` file [which none do currently]):
```javascript
var apolloRestSchema = require.resolve('apollo-link-rest/schema.graphql');

module.exports = {
  client: {
    name: 'Space Explorer [web]',
    service: 'apollographql-8341',
    includes: [
      'src/**/*.{ts,tsx,js,jsx,graphql,gql}',
      apolloRestSchema
    ]
  },
};
```
To make it easier to use the vscode extension with `apollo-client`, the extension has a baked-in copy of its directive definitions which are included if it detects that none of the `apollo-client` directives are already defined (either from Relay's `@connection` or someone including the `apollo-client` schema like above). This means that no config changes should be needed for developers who are only using `apollo-client`'s directives.

---

Fixes #847.

---

Progress:
- [x] Initial features 🎉 
- [x] Decide if including the `KnownDirectivesRule` validation rule causes more good than harm.
  - I definitely want to include this. Everyone should be taking advantage of "build time" validation.
- [x] Think about how plugins like `apollo-link-rest` can add custom client directives (should we bake them into the extension or should they live in their own plugin?)
  - ~~🙅‍♂ No to baked in directives!~~ Client schema extensions need to be configurable in the `apollo.config.js` config file.
- [x] ~~Figure out if I can use permalinks to our docs instead of the current linking in the baked in descriptions of client directives.~~
  - Permalinks are not available at the moment
- [x] Get tests passing
- [x] Add to changelog


packages/apollo-language-server/src/project/client.ts
```diff
@@ -211,10 +219,109 @@ export class GraphQLClientProject extends GraphQLProject {
...
+        The variable name to export this field as.
+        """
+        as: String!
+      ) on FIELD
+
+      """
+      Specify a custom store key for this result. See
+      https://www.apollographql.com/docs/react/advanced/caching/#the-connection-directive
+      """
+      directive @connection(
```
alloy: Currently this seems to equal Relay’s version of the directive, but I don’t know if that’s a guarantee and so perhaps this could yield wrong information to a Relay user?

alloy: Actually I was wrong, this doesn’t match Relay’s directive, which has a `filters` parameter (plural) https://relay.dev/docs/en/pagination-container#connection

alloy:
Just did a bit more testing to see if I could at least define all of Relay’s directives by adding a `.graphql` file to my source tree. This works for those that are not already defined in your PR. E.g. if I add the definition for the `relay` directive it is removed from the problems pane:

Screenshot 2019-07-26 at 13 03 48

But then if I add the definition of Relay’s `@connection` directive it seems to also undo the definition of the `@relay` directive:

Screenshot 2019-07-26 at 13 03 51

packages/apollo-language-server/src/utilities/source.ts
```diff
@@ -181,3 +182,46 @@ export function getASTNodeAndTypeInfoAtPosition(
...
+): boolean | null {
+  const { loc } = fieldNode;
+  if (!loc) return null;
+
+  let underClientDirective: boolean = false;
+
+  const typeInfo = new TypeInfo(schema);
+  visit(
+    root,
+    visitWithTypeInfo(typeInfo, {
```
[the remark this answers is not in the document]
jasonpaulos: Good catch, fixing this

jasonpaulos:
Great feedback @alloy! It's clear to me now that hardcoded definitions of Apollo directives may not be the best choice.

I've made some changes so that the Apollo directives are no longer included in the base extension; rather, the extension now includes as a default search path `node_modules/*/schema.graphql`, so that individual modules that expose custom client-side directives (like `apollo-client` and `apollo-link-rest`) can define them in a `schema.graphql` file that will automatically be used for validation, autocomplete, hover information, etc. by the extension.

The advantages of this approach are:
- The directive definitions are controlled in the same codebase as their implementation, so the vscode extension will never contain out of date definitions.
- Even though a `schema.graphql` will have to be added to frontend libraries like `apollo-client`, it should get ignored when bundling and won't add to the bundle size.
- Since the directives are defined in actual files, VSCode will let you cmd+click (or equivalent) to go to their definitions.
- It allows us to show an error for unknown directives while still exposing an easy way for 3rd party libraries/frameworks to included custom directives.

And of course, if you override the `includes` option and/or provide your own local schema like in your screenshot @alloy, you can add Relay definitions easily without them conflicting with Apollo definitions.

I'm interested to hear thoughts about this approach and any potential downsides!

jasonpaulos:
After discussing with @jbaxleyiii, we concluded that having the extension automatically check for `node_modules/*/schema.graphql` files and include them in the client schema has some faults. For projects with many dependencies, it can be very expensive to search through all of node_modules, which would impact the extension's startup time. Additionally, it's possible that some dependencies of dependencies might contain schema declarations that are unwanted by developers.

Since it's still very important to allow developers and package authors to configure, override, and include custom client schema like directives, I am currently figuring out a way we can support these processes in the `apollo.config.js` file. I believe the config file is a more natural place for this to happen, as long as the options and common use cases are documented. I will include commits/details here after I have more concrete ideas for what this could look like.

alloy:
Sorry for the delay in replying, but moving it to the config file echoes my thoughts entirely 👍

packages/apollo-language-server/src/project/client.ts
jasonpaulos: This is a great example of why the Apollo directives should be overridable.

packages/apollo-language-server/src/diagnostics.ts
```diff
@@ -72,11 +72,25 @@ export function diagnosticsFromError(
       source: `GraphQL: ${type}`,
       message: error.message,
       severity,
-      range: rangeForASTNode(highlightNodeForNode(node) || node)
```
trevor-scheer: This is the earliest line I can comment on, but should L65 (the return type definition `Diagnostic[]` be updated to `GraphQLDiagnostic[]`?

packages/apollo-language-server/src/utilities/source.ts
```diff
@@ -181,3 +182,39 @@ export function getASTNodeAndTypeInfoAtPosition(
...
+
+export function isFieldResolvedLocally(
+  fieldNode: FieldNode,
+  root: ASTNode
+): boolean | null {
+  const { loc } = fieldNode;
+  if (!loc) return null;
+
+  let underClientDirective: boolean = false;
+  visit(root, {
```
trevor-scheer: Let's include this when we pair, I think we can simplify quite a bit!

packages/vscode-apollo/syntaxes/graphql.json
```diff
@@ -52,10 +53,10 @@
     },
     "graphql-type-interface": {
       "name": "meta.type.interface.graphql",
-      "begin": "\\s*\\b(?:(extends)?\\b\\s*\\b(type)|(interface)|(input))\\b\\s*([_A-Za-z][_0-9A-Za-z]*)?",
+      "begin": "\\s*\\b(?:(extend)?\\b\\s*\\b(type)|(interface)|(input))\\b\\s*([_A-Za-z][_0-9A-Za-z]*)?",
```
trevor-scheer: Great catch!

packages/apollo-language-server/src/diagnostics.ts
```diff
@@ -72,11 +72,25 @@ export function diagnosticsFromError(
...
+      range: rangeForASTNode(highlightNodeForNode(node) || node),
+      error
     };
   });
 }
 
+export interface GraphQLDiagnostic extends Diagnostic {
```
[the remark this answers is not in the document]
jasonpaulos: This seems to be a vscode pattern that `Diagnostic` and other types follow.

packages/apollo-language-server/src/diagnostics.ts
jasonpaulos: Yes!

packages/apollo-language-server/src/project/client.ts
```diff
@@ -211,10 +219,69 @@ export class GraphQLClientProject extends GraphQLProject {
...
+  get missingApolloClientDirectives(): readonly DefinitionNode[] {
+    const { serviceSchema } = this;
+
+    const serviceDirectives = serviceSchema
+      ? serviceSchema.getDirectives().map(directive => directive.name)
+      : [];
+
+    const clientDirectives = this.typeSystemDefinitionsAndExtensions
+      .filter(def => def.kind === Kind.DIRECTIVE_DEFINITION)
+      .map(def => (def as DirectiveDefinitionNode).name.value);
```
trevor-scheer: Nothing wrong with this approach, but predicates are a nice way to tell TS that we know the type of something, and it looks like you'd be able to use it in a couple places.

```ts
function isDirectiveDefinitionNode(node: ASTNode): node is DirectiveDefinitionNode {
  return def.kind === Kind.DIRECTIVE_DEFINITION;
}
```

Then `isDirectiveDefinitionNode` becomes your filter function. From there, TS will understand you've got `DirectiveDefinitionNode[]` as a result!

packages/apollo-language-server/src/project/defaultClientSchema.ts
```diff
@@ -0,0 +1,45 @@
+import { GraphQLDocument } from "../document";
+import { Source } from "graphql";
+
+export const apolloClientSchema = `#graphql
```
trevor-scheer: What's the #graphql do? Is that triggering highlighting?

packages/apollo-language-server/src/utilities/source.ts
```diff
@@ -181,3 +185,39 @@ export function getASTNodeAndTypeInfoAtPosition(
...
+export function isFieldResolvedLocally(
+  fieldNode: FieldNode,
+  root: ASTNode
+): boolean | null {
+  const { loc } = fieldNode;
+  if (!loc) return null;
+
+  let underClientDirective: boolean = false;
+
+  function visitor(
```
trevor-scheer: Some comments explaining the logic of this function would be very helpful. The way we're leveraging location may be confusing. Explaining the problem you ran into when solving this as well as the unintended perf benefit of not recursing are worth mentioning.

packages/apollo-language-server/src/utilities/source.ts
```diff
@@ -181,3 +185,39 @@ export function getASTNodeAndTypeInfoAtPosition(
...
+    node: FieldNode | InlineFragmentNode | FragmentDefinitionNode
+  ) {
+    if (
+      loc &&
+      node.loc &&
+      node.loc.start <= loc.start &&
+      loc.end <= node.loc.end &&
+      node.directives &&
+      node.directives.find(directive => directive.name.value === "client") !=
```
trevor-scheer: When reducing arrays to true or false based on the contents:
i.e. do I have **at least one directive in here** that's named client?
i.e. is **every item in this array** an x?

We have the lesser-known, more semantic `Array.some()` and `Array.every()` functions that I'm a big fan of.

packages/apollo-language-server/src/project/defaultClientSchema.ts
jasonpaulos: Yes it is!

alloy:
(@jasonpaulos Let me know when I should give it another go.)

packages/apollo-language-server/src/utilities/source.ts
jasonpaulos: Great point, I think `Array.some()` makes a lot of sense here

packages/apollo-language-server/src/project/client.ts
jasonpaulos: This is a great suggestion, but unfortunately it doesn't remove the need to cast the definition to a `DirectiveDefinitionNode `. It seems that TypeScript is unable to modify the return type of the array with the `filter` method:
```javascript
filter(callbackfn: (value: T, index: number, array: ReadonlyArray<T>) => unknown, thisArg?: any): T[];
```
Since the return type is always `T[]`, I'm not sure if there's any way we can avoid casting like this.

jasonpaulos: Nevermind, it turns out you have to do this,
```javascript
const apolloDirectives = apolloAst.definitions
  .filter(isDirectiveDefinitionNode)
  .map(def => def.name.value);
```
NOT this,
```javascript
const apolloDirectives = apolloAst.definitions
  .filter(def => isDirectiveDefinitionNode(def))
  .map(def => def.name.value);
```

jasonpaulos:
@alloy Feel free to review this again! Now the extension will add in Apollo directives like `@client`, `@export`, etc. by default, unless a directive by the same name already exists in an included file. So since Relay defines `@connection` already, you shouldn't see any Apollo directives being suggested in typeahead. Let me know if this works for you.

packages/apollo-language-server/src/project/client.ts
trevor-scheer: TIL! That makes sense since the second way is "spoofing" the fn signature in a sense.

JakeDawkins:
I'm confused why this isn't being built by azure 🤔

JakeDawkins:
@jasonpaulos checked this on windows, and everything seems to be working properly :)

JakeDawkins:
@alloy any update here? Have you gotten a chance to give it another pass?

alloy:
I have not. I’ll try to do so later today, but as I’m on holiday I may not get to it and you should just merge.

Reading @jasonpaulos’ description it seems like Apollo specific directives may still be auto-completed/recognized in schemas that don’t support those, unless they happen to provide their own copy, is that correct?

jasonpaulos:
@alloy thanks for taking the time to review! You are correct that none of the Apollo directives will be defined if there's any name collisions with already defined directives. We want to make the extension as useful as possible for Apollo users, so it made sense for us to always try to include those directives. Perhaps we could introduce a config option to explicitly disable them, like `includeApolloClientDirectives: false` if there is a significant use case for it.

Also, by default the extension does now include the graphql `KnownDirectivesRule`, so it's strange that you were not seeing validation errors with unknown directives. Perhaps you were overriding the validation rules in the config? Or maybe the missing `apollo-env` build problem caused weird behavior--if you run `npm run clean && npm install` this should fix that problem.

Lastly, the only way to add in extra definitions is by including them in your config like below, which I think is pretty reasonable.
```javascript
var apolloRestSchema = require.resolve('apollo-link-rest/schema.graphql'); // or some other way to get to a file containing SDL

module.exports = {
  client: {
    name: 'Space Explorer [web]',
    service: 'apollographql-8341',
    includes: [
      'src/**/*.{ts,tsx,js,jsx,graphql,gql}',
      apolloRestSchema
    ]
  },
};
```

alloy:
> Also, by default the extension does now include the graphql `KnownDirectivesRule`, so it's strange that you were not seeing validation errors with unknown directives. […
 Or maybe the missing `apollo-env` build problem caused weird behavior--if you run `npm run clean && npm install` this should fix that problem.

Had to reclone in order to get a working build again, but after that indeed the `KnownDirectivesRule` showed up and I was getting problems reported 👌 

> Lastly, the only way to add in extra definitions is by including them in your config like below, which I think is pretty reasonable.

It’s definitely not unreasonable 😄 I’m unsure what other frameworks might do, but Relay doesn’t define all of its directives in graphql files, but rather as strings inside JS modules. For example, [here’s](https://github.com/facebook/relay/blob/79d23f9224b794b75fc740c47883820a6a7dd17e/packages/relay-compiler/transforms/RelayRelayDirectiveTransform.js#L22-L41) the `@relay` directive. So other than collecting them all, concatenating them into a temp file, and feeding that file to the config, as per your example, I don’t see a way to refer to those.

I’m fine going with that as a first version and see how it pans out, though 👍 

----

I just now realised another tougher issue in combination with Relay, which is its [`@argumentDefinitions`](https://relay.dev/docs/en/classic/graphql-in-relay#argumentdefinitions) and [`@arguments`](https://relay.dev/docs/en/classic/graphql-in-relay#arguments) directives, which define (pseudo) local arguments to fragments.

The problem with these directives is that their parameters are by definition unknown upfront, which is probably also why `relay-compiler` handles these directly in its parser. Seeing as there is no way to have an argument signature that says “accept anything”, usage of these directives leads to problems being reported:

Screenshot 2019-08-16 at 23 29 42

I don’t see a solution other than disabling the `KnownArgumentNames` rule entirely, which is a shame. If you see an opportunity for how this could be improved upon, then please do let me know 🙏 

Having said all that, this all is in no way a blocker for this PR imo, which is looking like a great next step forward for this vscode extension :clap:

alloy:
(I think the `@argumentDefinitions` and `@arguments` directives can only really be handled well by having actual code that understands them, which would also allow for discovering and validating the correct arguments are passed to the right fragments.)

alloy:
Following up on my earlier comments, I was able to make it all work. E.g. here’s a `KnownArgumentNames` validation rule that knows how to handle Relay’s fragment arguments: https://github.com/relay-tools/vscode-apollo-relay/blob/master/src/RelayKnownArgumentNames.ts

cltnschlosser:
Hey guys, we've found this commit to be the cause of a huge performance regression in codegen. See https://github.com/apollographql/apollo-tooling/issues/1554#issuecomment-537655205 for more details, hopefully we can improve this approach.

pirtlj:
Any thought about putting this back in at some point?

Issue #1182, opened 2019-04-12
Feature request: Compress insignificant whitespace in queries when generating code


lilyball:
When generating code, the resulting query objects include a string literal with the complete GraphQL query definition, including all of the whitespace. This makes it look rather ugly in the source, and also just seems completely unnecessary to send to the server.

Specifically I'm generating Swift code, though I'd assume this holds true for other languages as well.

It would be nice if `apollo client:codegen` would just replace all sequences of insignificant whitespace (i.e. whitespace not in strings) with a single space character. It looks like the codegen already strips out comments, so we don't have to worry about those, and without comments newlines have no significance either, so those can be included in the whitespace compression.

For example, given the following query:

```graphql
query Foo($name: String!) {
  user(login: $name) {
    id
    displayName
  }
}
```

`apollo client:codegen` produces the following definition:

```swift
public final class FooQuery: GraphQLQuery {
  public let operationDefinition =
    "query Foo($name: String!) {\n  user(login: $name) {\n    __typename\n    id\n    displayName\n  }\n}"
  …
}
```

What I'm asking for is to instead produce:

```swift
public final class FooQuery: GraphQLQuery {
  public let operationDefinition =
    "query Foo($name: String!) { user(login: $name) { __typename id displayName } }"
```

There's obviously a couple of spaces that could still be trimmed (e.g. between punctuation), but I'm not looking for the 100% optimal compression, just something that's readable and omits the large runs of obviously-irrelevant whitespace.

JakeDawkins:
Yeah, I think this would still be a good thing to do. Either this or a multiline string

david-coyle-sjc:
I'm seeing an issue here, where my iOS swift sources, pass through the literal `\n` in the query, it's not converted to an actual new line.  This causes the queries to fail, obviously.
I've even removed all the newlines from my .graphql query specs, and yet the codegen is adding them back in.

designatednerd:
@david-coyle-sjc Can you take a look at PR #1441 - that adds support for a multiline string, which should stop that literal from being passed through.

designatednerd:
@lilyball - are you still interested in an option for this? I moved to the multiline-string for readability, but from a standpoint of "Do we really need to send whitespace?", I *think* the answer is no. I could take a look at some kind of option to not send the whitespace, although that may be a bit.

lilyball:
@designatednerd I'm still interested in this.

designatednerd:
[...]

PR #1441, opened 2019-07-26
Swift Codegen Upgrades

designatednerd:
In this PR: 
- Added `CaseIterable` conformance to `enum` types so all known cases can be easily iterated.
- Added comments with the original fragment or query definition to `operationDefinition` and `fragmentDefinition`
- Trimmed out excess whitespace from the final strings sent to the server for `operationDefinition` and `fragmentDefinition`
- Removed force-unwrap when the thing being unwrapped is a double optional

Note to tooling team: When this gets merged it should be released as minor-version alpha so I can make sure everything works well against our API and I haven't missed anything. 

Also I am teh suck at typescript so would love any suggestions on how to do stuff better.

- [X] Update CHANGELOG.md with your change (include reference to issue & this PR)

### DIFF
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -879,6 +879,10 @@ exports[`Swift code generation #initializerDeclarationForProperties() should gen
 
 exports[`Swift code generation #structDeclarationForFragment() should generate a struct declaration for a fragment that includes a fragment spread 1`] = `
 "public struct HeroDetails: GraphQLFragment {
+  /// fragment HeroDetails on Character {
+  ///   name
+  ///   ...MoreHeroDetails
+  /// }
   public static let fragmentDefinition =
     \\"fragment HeroDetails on Character {\\\\n  name\\\\n  ...MoreHeroDetails\\\\n}\\"
 
@@ -953,6 +957,10 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
 
 exports[`Swift code generation #structDeclarationForFragment() should generate a struct declaration for a fragment with a concrete type condition 1`] = `
 "public struct DroidDetails: GraphQLFragment {
+  /// fragment DroidDetails on Droid {
+  ///   name
+  ///   primaryFunction
+  /// }
   public static let fragmentDefinition =
     \\"fragment DroidDetails on Droid {\\\\n  name\\\\n  primaryFunction\\\\n}\\"
 
@@ -997,6 +1005,12 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
 
 exports[`Swift code generation #structDeclarationForFragment() should generate a struct declaration for a fragment with a subselection 1`] = `
 "public struct HeroDetails: GraphQLFragment {
+  /// fragment HeroDetails on Character {
+  ///   name
+  ///   friends {
+  ///     name
+  ///   }
+  /// }
   public static let fragmentDefinition =
     \\"fragment HeroDetails on Character {\\\\n  name\\\\n  friends {\\\\n    name\\\\n  }\\\\n}\\"
 
@@ -1077,6 +1091,10 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
 
 exports[`Swift code generation #structDeclarationForFragment() should generate a struct declaration for a fragment with an abstract type condition 1`] = `
 "public struct HeroDetails: GraphQLFragment {
+  /// fragment HeroDetails on Character {
+  ///   name
+  ///   appearsIn
+  /// }
   public static let fragmentDefinition =
     \\"fragment HeroDetails on Character {\\\\n  name\\\\n  appearsIn\\\\n}\\"
 
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -1712,7 +1712,7 @@ public struct ReviewInput: GraphQLMapConvertible {
   /// Comment about the movie, optional
   public var commentary: Swift.Optional<String?> {
     get {
-      return graphQLMap[\\"commentary\\"] as! Swift.Optional<String?>
+      return graphQLMap[\\"commentary\\"] as? Swift.Optional<String?>
     }
     set {
       graphQLMap.updateValue(newValue, forKey: \\"commentary\\")
@@ -1722,7 +1722,7 @@ public struct ReviewInput: GraphQLMapConvertible {
   /// Favorite color, optional
   public var favoriteColor: Swift.Optional<ColorInput?> {
     get {
-      return graphQLMap[\\"favorite_color\\"] as! Swift.Optional<ColorInput?>
+      return graphQLMap[\\"favorite_color\\"] as? Swift.Optional<ColorInput?>
     }
     set {
       graphQLMap.updateValue(newValue, forKey: \\"favorite_color\\")
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -173,14 +173,7 @@ exports[`Swift code generation #classDeclarationForOperation() should correctly 
 exports[`Swift code generation #classDeclarationForOperation() should generate a class declaration for a mutation with variables 1`] = `
 "public final class CreateReviewMutation: GraphQLMutation {
   public let operationDefinition =
-\\"\\"\\"
-mutation CreateReview($episode: Episode) {
-  createReview(episode: $episode, review: {stars: 5, commentary: \\"Wow!\\"}) {
-    stars
-    commentary
-  }
-}
-\\"\\"\\"
+    \\"mutation CreateReview($episode: Episode) {\\\\n  createReview(episode: $episode, review: {stars: 5, commentary: \\\\\\"Wow!\\\\\\"}) {\\\\n    stars\\\\n    commentary\\\\n  }\\\\n}\\"
 
   public let operationName = \\"CreateReview\\"
 
@@ -265,15 +258,7 @@ mutation CreateReview($episode: Episode) {
 exports[`Swift code generation #classDeclarationForOperation() should generate a class declaration for a query with a fragment spread nested in an inline fragment 1`] = `
 "public final class HeroQuery: GraphQLQuery {
   public let operationDefinition =
-\\"\\"\\"
-query Hero {
-  hero {
-    ... on Droid {
-      ...HeroDetails
-    }
-  }
-}
-\\"\\"\\"
+    \\"query Hero {\\\\n  hero {\\\\n    ... on Droid {\\\\n      ...HeroDetails\\\\n    }\\\\n  }\\\\n}\\"
 
   public let operationName = \\"Hero\\"
 
@@ -405,13 +390,7 @@ query Hero {
 exports[`Swift code generation #classDeclarationForOperation() should generate a class declaration for a query with conditional fragment spreads 1`] = `
 "public final class HeroQuery: GraphQLQuery {
   public let operationDefinition =
-\\"\\"\\"
-query Hero {
-  hero {
-    ...DroidDetails
-  }
-}
-\\"\\"\\"
+    \\"query Hero {\\\\n  hero {\\\\n    ...DroidDetails\\\\n  }\\\\n}\\"
 
   public let operationName = \\"Hero\\"
 
@@ -571,13 +550,7 @@ query Hero {
 exports[`Swift code generation #classDeclarationForOperation() should generate a class declaration for a query with fragment spreads 1`] = `
 "public final class HeroQuery: GraphQLQuery {
   public let operationDefinition =
-\\"\\"\\"
-query Hero {
-  hero {
-    ...HeroDetails
-  }
-}
-\\"\\"\\"
+    \\"query Hero {\\\\n  hero {\\\\n    ...HeroDetails\\\\n  }\\\\n}\\"
 
   public let operationName = \\"Hero\\"
 
@@ -676,13 +649,7 @@ query Hero {
 exports[`Swift code generation #classDeclarationForOperation() should generate a class declaration for a query with variables 1`] = `
 "public final class HeroNameQuery: GraphQLQuery {
   public let operationDefinition =
-\\"\\"\\"
-query HeroName($episode: Episode) {
-  hero(episode: $episode) {
-    name
-  }
-}
-\\"\\"\\"
+    \\"query HeroName($episode: Episode) {\\\\n  hero(episode: $episode) {\\\\n    name\\\\n  }\\\\n}\\"
 
   public let operationName = \\"HeroName\\"
 
@@ -760,13 +727,7 @@ query HeroName($episode: Episode) {
 exports[`Swift code generation #classDeclarationForOperation() should generate a class declaration with an operationIdentifier property when generateOperationIds is specified 1`] = `
 "public final class HeroQuery: GraphQLQuery {
   public let operationDefinition =
-\\"\\"\\"
-query Hero {
-  hero {
-    ...HeroDetails
-  }
-}
-\\"\\"\\"
+    \\"query Hero {\\\\n  hero {\\\\n    ...HeroDetails\\\\n  }\\\\n}\\"
 
   public let operationName = \\"Hero\\"
 
@@ -886,12 +847,7 @@ exports[`Swift code generation #initializerDeclarationForProperties() should gen
 exports[`Swift code generation #structDeclarationForFragment() should generate a struct declaration for a fragment that includes a fragment spread 1`] = `
 "public struct HeroDetails: GraphQLFragment {
   public static let fragmentDefinition =
-\\"\\"\\"
-fragment HeroDetails on Character {
-  name
-  ...MoreHeroDetails
-}
-\\"\\"\\"
+    \\"fragment HeroDetails on Character {\\\\n  name\\\\n  ...MoreHeroDetails\\\\n}\\"
 
   public static let possibleTypes = [\\"Human\\", \\"Droid\\"]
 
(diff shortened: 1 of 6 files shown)


lilyball:
Multiline strings would make the raw queries easier to read, but I'd prefer to strip irrelevant whitespace (see #1182) instead, because the raw queries aren't what I'm looking at the source for.

designatednerd:
@lilyball I actually don't think those are incompatible goals - I'm planning to take a look at stripping that whitespace at the point where it's about to go out to the network next week.

lilyball:
Once you strip all irrelevant whitespace, that should strip all newlines too, at which point the multiline string only serves to turn one line into 3. The only exception I can think of to stripping newlines is if you put a multiline string literal in your GraphQL directly that includes a newline (looking at the spec right now it uses the exact same `"""` token to start a [multiline string literal](https://graphql.github.io/graphql-spec/June2018/#StringValue)).

Speaking of, I hope this PR handles multiline string literals in the GraphQL source. But glancing at the diff right now I'm guessing it doesn't.

lilyball:
Note that if you simply want to avoid having to escape quotes and backticks, you can use a raw string literal instead of a multiline string literal (you just have to detect if the raw string literal terminator occurs anywhere in the source text, and if so, increase the number of hash marks used for the delimiters).

lilyball:
Well, I suppose if the source string then includes a newline inside of a multiline string literal you'll need to escape that, using the special form of escapes that raw strings use.

designatednerd:
@lilyball It's not about avoiding escape quotes and backticks - it's about seeing if the provided GraphQL operation is actually being properly represented properly by the generated code more easily.

> Speaking of, I hope this PR handles multiline string literals in the GraphQL source. But glancing at the diff right now I'm guessing it doesn't.

Can you clarify what you mean by this a bit?

designatednerd:
OK - I'll take a fresh look monday if the multi-line string delimiter in GQL is `"""` - you're correct that would need special case handling. I can add some tests around that.

designatednerd:
Personally (and I don't believe I'm alone on this given [iOS 179](https://github.com/apollographql/apollo-ios/issues/179)), I find it considerably easier to read everything with a multi-line string than with everything glommed together without whitespace. I agree that getting rid of the whitespace when it hits the network is helpful though, that way you're not sending data you don't need to, and I'll deal with that in a separate PR (partly because I think I may have to do this on the client end rather than in codegen)

lilyball:
A multiline string is definitely better than what we have today, because what we have today is just really hard to read, but I'm not generally reading the operation definition in the generated Swift source anyway so we don't need to optimize for the prettiest output, just something reasonably legible.

As far as stripping whitespace, there would still be one space between tokens, and I think it's quite reasonable to keep some spacing around `{`, `}`, `:`, and `,` for legibility's sake, but everything else can go. I'd much rather see

```swift
public final class ChatUserInfoByIdQuery: GraphQLQuery {
  public let operationDefinition =
    "query ChatUserInfoByID($userID: ID!, $channelID: ID!) { user(id: $userID) { __typename ...IdentityWithRolesFragment isModerator(channelID: $channelID) } chatRoomBanStatus(userID: $userID, channelID: $channelID) { __typename isPermanent } }"
```

then see

```swift
public final class ChatUserInfoByIdQuery: GraphQLQuery {
  public let operationDefinition = """
query ChatUserInfoByID($userID: ID!, $channelID: ID!) {
  user(id: $userID) {
    __typename
    ...IdentityWithRolesFragment
    isModerator(channelID: $channelID)
  }
  chatRoomBanStatus(userID: $userID, channelID: $channelID) {
    __typename
    isPermanent
  }
}
"""
```

And the reason is because if I'm actually looking at the generated Swift file, it's because I want to see the Swift API, not because I want to see the original query definition. If I wanted to see that I could just open the `.graphql` file.

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -62,7 +62,15 @@ export class SwiftGenerator<Context> extends CodeGenerator<
...
-    this.printOnNewline(`"${escapedString(string)}"`);
+    this.printNewline();
+    this.print(`"""`);
+
+    string.split("\n").forEach(line => {
+      this.printNewline();
+      this.print(line);
+    });
+    this.printNewline();
+    this.print(`"""`);
```
lilyball: Can we indent the multiline string by the appropriate amount? The indentation of the final `"""` is trimmed from all lines (which must have at least the same indentation or it's a syntax error), so indenting this won't affect the runtime string value.

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -62,7 +62,15 @@ export class SwiftGenerator<Context> extends CodeGenerator<
   }
 
   multilineString(string: string) {
-    this.printOnNewline(`"${escapedString(string)}"`);
+    this.printNewline();
+    this.print(`"""`);
+
+    string.split("\n").forEach(line => {
+      this.printNewline();
+      this.print(line);
```
lilyball: As mentioned at the top-level discussion, this isn't handling any GraphQL multiline strings (which use the same `"""` token). Also looking at this right now, it's not going to handle any backslash-escapes in the GraphQL source either.

packages/apollo-codegen-swift/src/language.ts
designatednerd: Yeah fair point - I'll take a better look at this today or tomorrow

packages/apollo-codegen-swift/src/language.ts
designatednerd: Yeah...blargh. I might take another pass at this as a comment for readability, but leave the actual `operationDefinition` as a one-liner for the reasons you outlined.

designatednerd:
@lilyball I think I've addressed your comments, and I've also tried to address #1182. 

Basically since the only thing I want the multiline string literal for is to see what's happening, a comment serves this purpose without breaking the crap out of everything else. Then trimming the excess whitespace is also considerably easier. 

Let me know if you have questions or concerns - it sounds like you've spent quite a bit more time with the GraphQL spec than I have so far, so if there's anything that I'm not thinking of, with this approach, please let me know!

packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
```diff
@@ -2,8 +2,14 @@
...
 "public final class CreateReviewMutation: GraphQLMutation {
+  /// mutation CreateReview($episode: Episode) {
+  ///   createReview(episode: $episode, review: {stars: 5, commentary: \\"Wow!\\"}) {
+  ///     stars
+  ///     commentary
+  ///   }
+  /// }
   public let operationDefinition =
-    \\"mutation CreateReview($episode: Episode) {\\\\n  createReview(episode: $episode, review: {stars: 5, commentary: \\\\\\"Wow!\\\\\\"}) {\\\\n    stars\\\\n    commentary\\\\n  }\\\\n}\\"
+    \\"mutation CreateReview($episode: Episode) {\\\\n createReview(episode: $episode, review: {stars: 5, commentary: \\\\\\"Wow!\\\\\\"}) {\\\\n stars\\\\n commentary\\\\n }\\\\n }\\"
```
lilyball: It looks like you're trimming adjacent spaces but doing nothing about the embedded newlines.

packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
```diff
@@ -2,8 +2,14 @@
 
 exports[`Swift code generation #classDeclarationForOperation() should generate a class declaration for a mutation with variables 1`] = `
 "public final class CreateReviewMutation: GraphQLMutation {
+  /// mutation CreateReview($episode: Episode) {
+  ///   createReview(episode: $episode, review: {stars: 5, commentary: \\"Wow!\\"}) {
+  ///     stars
+  ///     commentary
+  ///   }
+  /// }
```
lilyball: To be honest, I very explicitly *do not want this*. This comment just bloats the file. I notice all your tests are relatively short snippets, but for comparison, I have a single query I'm looking at right now that spans 56 lines, and I'd really not like to have a 56-line comment in the generated source obscuring the actual datatype.

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -29,7 +29,16 @@ export interface Property {
 }
 
 export function escapedString(string: string) {
-  return string.replace(/"/g, '\\"').replace(/\n/g, "\\n");
+  var lines = string
+    .split(/\n/g)
+    .map(line => {
+      return line.trim();
```
lilyball: This is going to trim meaningful whitespace from multiline strings. As an example from the spec:

```graphql
mutation {
  sendEmail(message: """
    Hello,
      World!

    Yours,
      GraphQL.
  """)
}
```

is equivalent to

```graphql
mutation {
  sendEmail(message: "Hello,\n  World!\n\nYours,\n  GraphQL.")
}
```

but with this PR it would instead turn into

```graphql
mutation {
  sendEmail(message: "Hello,\nWorld!\n\nYours,\nGraphQL.")
}
```

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -29,7 +29,16 @@ export interface Property {
 }
 
 export function escapedString(string: string) {
-  return string.replace(/"/g, '\\"').replace(/\n/g, "\\n");
```
lilyball: FWIW, this escaping won't work with multiline strings either, because `\` is not special in a multilline string except as part of the token `\"""`.

packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
designatednerd: Ah I think I missed something on the request there. This should now be updated.

packages/apollo-codegen-swift/src/language.ts
designatednerd: I've opened #1450 on handling of multiline strings in the Swift codegen.

To be honest, I have a longer-term plan to move the generation of codegen to Swift so it's easier for Swift developers to contribute. Since failure to handle multiline strings isn't a regression, I'd prefer to wait to address this issue until I'm working on the updated codegen, since I have a *way* better handle on Swift's string handling than Typescript's.

packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
designatednerd: There is at least one other person who has [explicitly requested multiline string support](https://github.com/apollographql/apollo-ios/issues/179). 

In your opinion, what is better when we're talking about things of similar length: A multiline string that has to be compiled or a multiline comment that doesn't?

lilyball: The comment approach is better. Can we make it an option as to whether to have it though?

packages/apollo-codegen-swift/src/language.ts
lilyball: Actually, come to think of it, the old code probably worked fine after all, because `string` presumably contains the properly-interpreted contents of the multiline string. In that case, this PR will explicitly break multiline strings.

I would strongly encourage you to add a test case that has a multiline string with backslash-escapes in it (since backslash-escapes, except for `\"""`, aren't meaningful in multiline strings), verify its behavior *without this PR*, and then see what this PR does to it.

designatednerd: I snuck a couple additional tests in on #1452 to validate this. I'll try to get this being less cranky in the morning.

designatednerd:
@lilyball Would be more than happy to take a look at a PR adding the ability to opt out of this feature, as at the moment I need to focus a bit elsewhere. 

Otherwise, any further questions here?

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -29,7 +29,23 @@ export interface Property {
...
+    return string.replace(/"/g, '\\"').replace(/\n/g, "\\n");
+  } else {
+    // Strip unnecessary whitespace.
+    var lines = string
+      .split(/\n/g)
+      .map(line => {
+        return line.trim();
+      })
+      .map(line => {
+        return line.replace(/"/g, '\\"');
```
JacksonKearl: tiny style note: 
```ts
a => { return b }
```
can be rewritten to 
```ts
a => b
```

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -29,7 +29,23 @@ export interface Property {
...
 
 export function escapedString(string: string) {
-  return string.replace(/"/g, '\\"').replace(/\n/g, "\\n");
+  if (string.includes('"""')) {
+    // This includes a multi-line string literal, and we may strip out meaningful
+    // whitespace if we try to strip whitespace. Don't try.
+    return string.replace(/"/g, '\\"').replace(/\n/g, "\\n");
+  } else {
+    // Strip unnecessary whitespace.
+    var lines = string
```
JacksonKearl: `var` is so 2000-late, the hip thing nowadays is `const` or `let` (which is precisely opposite swift's `let`). ;)

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -29,7 +29,23 @@ export interface Property {
 }
 
 export function escapedString(string: string) {
-  return string.replace(/"/g, '\\"').replace(/\n/g, "\\n");
+  if (string.includes('"""')) {
+    // This includes a multi-line string literal, and we may strip out meaningful
+    // whitespace if we try to strip whitespace. Don't try.
+    return string.replace(/"/g, '\\"').replace(/\n/g, "\\n");
+  } else {
+    // Strip unnecessary whitespace.
```
JacksonKearl: This block could be rewritten: 
```ts
return string
  .split(/\n/g)
  .map(line => line.trim())
  .map(line => line.replace(/" /g, '\\"')
  .join(" ");
```

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -1045,9 +1063,15 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
...
+              if (isOptional) {
+                this.printOnNewline(
+                  `return graphQLMap["${name}"] as? ${typeName}`
+                );
+              } else {
+                this.printOnNewline(
+                  `return graphQLMap["${name}"] as! ${typeName}`
+                );
+              }
             });
```
JacksonKearl: Can be rewritten:
```ts
this.withinBlock(() => 
   this.printOnNewline( isOptional 
       ? `return graphQLMap["${name}"] as? ${typeName}`
       : `return graphQLMap["${name}"] as! ${typeName}`))
```

or even
```ts
this.withinBlock(() => 
   this.printOnNewline(`return graphQLMap["${name}"] as${isOptional ? '?' : '!'} ${typeName}`))
```

IMO this makes it a bit more immediately clear what the effect of `isOptional` is, but this is a style thing that could realistically go either way.

packages/apollo-codegen-swift/src/language.ts
designatednerd: 👍Yeah Swift (for now) you still always have to write a return statement so that's why I did

packages/apollo-codegen-swift/src/language.ts
designatednerd: > which is precisely opposite swift's `let`). ;)

This is why I drink.

packages/apollo-codegen-swift/src/codeGeneration.ts
designatednerd: This is more in line with how this is done throughout the rest of the file - I like the first option but I'm slightly reluctant to touch the other stuff at the moment if I don't have to.

packages/apollo-codegen-swift/src/language.ts
designatednerd: FWIW I had to leave the `return` in for `trim()` because otherwise it returns the result of the `trim()` command (which is `void`) rather than the actual trimmed string.

JacksonKearl: Oh darn. Always a bummer when things break chaining for no reason.

designatednerd: I mean there is _a_ reason, just a bit of an annoying one

designatednerd: Womp, [looks like it's needed for both](https://circleci.com/gh/apollographql/apollo-tooling/27832)

JacksonKearl: Oh no, I'm sad at the `trim` spec, not your code. 🙃

JacksonKearl: Wait a sec, are we thinking of the same `trim`? Says here that it returns the trimmed string and doesn't modify the original. Confused. https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/string/trim

JacksonKearl: Are you using VSCode? It should be able to tell you about return values of stuff right in the editor. Also we have some scripts to let VSCode run the tests for you locally, so you don't have to wait for Circle to get around to your build.

designatednerd: Yeah if I don't use a return, VSCode tells me I can't do a `replace` on `void`: 
Screen Shot 2019-08-01 at 10 37 44 PM

designatednerd:
🎉OK - let's leave this open until after #1451 gets merged, I'd love to get that one out in a patch release since it's a bugfix, and this one should really go in a minor release and get alpha'd before we release the hounds on it.

packages/apollo-codegen-swift/src/language.ts
JacksonKearl: Oh sorry, you need to remove the `{}` as well. Just `line => line.trim()`, not `line => { line.trim() }`. Its a convenience syntax for when the function consists of returning a single expression.

designatednerd: AHA. this worked 😃

designatednerd:
OK thanks to @abernix for some help testing this against the iOS project - added a nil-coalesce to `.none` for double-optionals if there's no value to fix a build issue. This is now ready to go out as a minor release!

lilyball:
I've finally upgraded Apollo tooling and I'm still not happy with the comment. For example I just opened one of my queries and my *entire screen* is filled with the comment. It's even worse than I thought because the comment includes all of the `__typename` fields that aren't present in the original query. It's a 67-line comment and it just makes this file harder to work with.

I'm also surprised to see it's a doc comment, except it doesn't use any markdown formatting, which means it looks truly awful when viewed in Xcode's Quick Help. The only reason to use a doc comment is for Quick Help to show it, so I don't understand why it was done like this.

designatednerd:
@lilyball The reason it includes all the `typename` is to make sure the query in the comment is *exacty* what's being sent to the server - and unfortunately that's a requirement to make sure we're staying type-safe end to end. 

Good call on using markdown in the doc comment - I'm a little swamped at the moment after coming back from moving, if you have some time I'd be more than happy to take a look at a PR if you have some time.

PR #1447, opened 2019-07-28
Clear projectForFileCache when schema is reloaded

### LINKED ISSUE #1446: Reload schema command doesn't reflect added fields
I have a apollo.config.js like this:
```javascript
module.exports = {
  client: {
    includes: ['src/**/*.js', 'src/*.js'],
    service: {
      url: 'http://localhost:4000'
    }
  },
};
```

If I add a field to`http://localhost:4000` and run the `Apollo: Reload schema` command, this field is not picked up unless I completely restart VSCode.

Hoping to get this working for a demo on Apollo I'm giving next week :). Will make the workflow a lot smoother if I can modify the backend and then show the results immediately.

A minimal reproduction is below:
[reload-test.zip](https://github.com/apollographql/apollo-tooling/files/3438621/reload-test.zip)

Steps to reproduce:
- `npm install`
- `npm server.js`
- Open the folder in vscode and in `src/client.js` you'll be able to see that autocomplete is working properly
- stop the server, uncomment the year property, and restart the server
- In vscode, run the `Apollo: Reload schema` command
- Go back to `src/client.js` and try to type something. You'll see that the year field is not being picked up
- Restart vscode and go back to `src/client.js` you'll see the year field is now available

However, if I do the same but I comment out a field such as author and call the reload schema command, this gets picked up properly without needing a restart with a message like `Cannot query field "author" on type "Book".`

Any guidance on where to begin investigating would be appreciated!

venkatd:
Resolves #1446 

When I stepped through the debugger I noticed that the cache wasn't getting properly cleared which resulted in outdated autocomplete results. Adding cache clearing to`project.reloadSchema` appears to resolve the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 - `apollo-graphql`
   - <First `apollo-graphql` related entry goes here>
 - `apollo-language-server`
-  - <First `apollo-language-server` related entry goes here>
+  - Fix cache invalidation bug for reload schema which caused outdated results in autocomplete [#1446](https://github.com/apollographql/apollo-tooling/pull/1446)
 - `apollo-tools`
   - <First `apollo-tools` related entry goes here>
 - `vscode-apollo`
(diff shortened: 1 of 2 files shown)


venkatd:
Can confirm that if I build the extension `npm run package-extension` and install from vsix, it resolves my issue.

JakeDawkins:
Great catch! Thank you!

Issue #1449, opened 2019-07-30
[Swift] Access modifiers are omitted when generating a single file output with a namespace


gsabran:
**Intended outcome:**

I am generating Swift code using a namespace. The output looks like 

```swift
public enum MyNameSpace {
    struct Hero: GraphQLFragment {
        public var name: String
        public var age: Int
    }
}
```

while it should be

```swift
public enum MyNameSpace {
    public struct Hero: GraphQLFragment {
        public var name: String
        public var age: Int
    }
}
```

As far as I can tell this is a regression introduced by https://github.com/apollographql/apollo-tooling/pull/1241. This change should only apply to multi file outputs that use public extensions of the namespace.

This issue prevents the usage of the generated objects outside of the exact module where they are defined.

If this is indeed the issue, the required package version in [apollo-ios](https://github.com/apollographql/apollo-ios/blob/master/scripts/check-and-run-apollo-cli.sh)  should be bumped once it is fixed.

Let me know if I can provide more information to help diagnose!

designatednerd:
Argh! Thanks for the heads up, I'll get on this.

designatednerd:
This is fixed in `2.17`, and will be part of iOS `0.14.0`

PR #1451, opened 2019-07-30
Swift: Fix regression for individual files. 

designatednerd:
#1449 Pointed out that #1241 introduced a regression when files are being output individually. This PR updates our logic to make sure that information is shared and then fix the issue.

### DIFF
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -175,7 +175,7 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
       options: { namespace },
       fragments
     } = this.context;
-    const isRedundant = !!namespace;
+    const isRedundant = !!namespace && !outputIndividualFiles;
     const modifiers = isRedundant ? ["final"] : ["public", "final"];
 
     this.classDeclaration(
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -47,6 +47,15 @@ export interface Options {
   customScalarsPrefix?: string;
 }
 
+/**
+ * The main method to call from outside this package to generate Swift code.
+ *
+ * @param context The `CompilerContext` to use to generate code.
+ * @param outputIndividualFiles Generates individual files per query/fragment if true,
+ *                              otherwise shoves everything into one giant file.
+ * @param only [optional] The path to a file which is the only file which should be regenerated.
+ *             If absent, all files will be regenerated.
+ */
 export function generateSource(
   context: CompilerContext,
   outputIndividualFiles: boolean,
@@ -137,6 +146,13 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
     this.printOnNewline("import Apollo");
   }
 
+  /**
+   * Generates the class declaration for an operation.
+   *
+   * @param operation The operaton to generate the class declaration for.
+   * @param outputIndividualFiles If this operation is being output as individual files, to help prevent
+   *                              redundant usages of the `public` modifier in enum extensions.
+   */
   classDeclarationForOperation(
     operation: Operation,
     outputIndividualFiles: boolean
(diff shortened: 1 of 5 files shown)


gsabran:
Thanks for making this PR so promptly!
Do you also need to change `isRedundant` here https://github.com/apollographql/apollo-tooling/blob/2843a4363c00cf85436b0950dbcca85a97630549/packages/apollo-codegen-swift/src/codeGeneration.ts#L175 to `const isRedundant = outputIndividualFiles && !!namespace;`?

designatednerd:
@gsabran Guh, good catch - there were a couple things that seem to have not made the bus on my last commit.

designatednerd:
I'm gonna take another look at this in the morning before merging - @JacksonKearl let me know if any of the TS looks fishy to you

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -172,7 +175,7 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
       options: { namespace },
       fragments
     } = this.context;
-    const isRedundant = !!namespace;
+    const isRedundant = !!namespace && !outputIndividualFiles;
```
gsabran: Isn't the codegen written with extensions when `outputIndividualFiles` is true? In that case this should be
```swift
const isRedundant = !!namespace && outputIndividualFiles;
```
right?

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -144,13 +144,16 @@ export class SwiftGenerator<Context> extends CodeGenerator<
...
     closure: Function
   ) {
     this.printNewlineIfNeeded();
     this.comment(description);
 
     const isRedundant =
-      adoptedProtocols.includes("GraphQLFragment") && !!namespace;
+      adoptedProtocols.includes("GraphQLFragment") &&
+      !!namespace &&
+      !outputIndividualFiles;
```
gsabran: Same question regarding whether `!outputIndividualFiles` should be `outputIndividualFiles`

packages/apollo-codegen-swift/src/codeGeneration.ts
designatednerd: You're correct that it's written with extensions when `outputIndividualFiles` is true, but `isRedundant` should only ever be true if we're **not** outputting individual files.

If `isRedundant` is true, codegen omits the `public`, which is the behavior filed as a bug.

packages/apollo-codegen-swift/src/language.ts
designatednerd: same answer

designatednerd:
@gsabran I'll wait on your  👍to merge this one - let me know if you have more questions!

gsabran:
`public` is redundant for extensions so the desired outputs are

```swift
public enum MyNameSpace {}

public extension MyNameSpace {
   struct Hero: GraphQLFragment { // <- public would be redundant here
        public var name: String
        public var age: Int
    }
}
```

but not for enums as 
```swift
public enum MyNameSpace {
    public struct Hero: GraphQLFragment { // <- public is required here
        public var name: String
        public var age: Int
    }
}
```

so I think we should only look for redundancy when the codegen generates extensions, ie when it outputs multiple files. Maybe I'm misunderstanding what `outputIndividualFiles` stands for.

designatednerd:
AHA. I think I see where I missed something. Good catch!

packages/apollo-codegen-swift/src/codeGeneration.ts
designatednerd: After figuring out the thing I fucked up, yep, you're right!

designatednerd:
@gsabran better?

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -137,7 +137,10 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
     this.printOnNewline("import Apollo");
   }
 
-  classDeclarationForOperation(operation: Operation) {
+  classDeclarationForOperation(
+    operation: Operation,
+    outputIndividualFiles: boolean
```
JacksonKearl: I'd add a docstring to describe what exactly this parameter is doing (maybe with examples of the difference in output?), because it's not very clear to me just looking at it.

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -350,152 +361,165 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
       variant: Variant;
       typeCase?: TypeCase;
     },
+    outputIndividualFiles: boolean,
```
JacksonKearl: same here re docstring

designatednerd:
@gsabran @JacksonKearl Would love 👍s from you guys if you think this is ready to go - would love to get this merged soon so we can put it out with a patch release

gsabran:
thanks for making this change!

PR #1452, opened 2019-07-30
[Swift] Fix escaping of parameter and class names

designatednerd:
This fix addresses [iOS 193](https://github.com/apollographql/apollo-ios/issues/193), where the name of the parameter was being escaped in a property, but the name of the type was not. 

Also made sure the names of classes are being properly escaped, and the names of parameters on protocol property declarations.

Note: I've left the type unescaped on protocol properties purposely, because I'm not sure if we're using `Self` as a bound anywhere, and I'd like to fix the other stuff without breaking the crap out of that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - `apollo-codegen-scala`
   - <First `apollo-codegen-scala` related entry goes here>
 - `apollo-codegen-swift`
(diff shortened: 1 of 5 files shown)


designatednerd:
Also threw in a couple tests so we can better check regressions on #1441. Merging.

PR #1454, opened 2019-07-31
AP-682 Strip `FragmentDefinition`s in `client:check` when only included in `@client` fields

justinanastos:
We're getting `client:check` failures when using a `Fragment `in a `@client` field because while the fields with `@client` are stripped, `Fragment`s that are nested below them are not. We now check that `Fragment`s can be removed if they are only used in `@client` fields.

I added tests to the existing functionality on one commit, added a failing test for how I expected the method to behave, and then fixed the test. Please look at the last new test to see a reproduction.

---

TODO:

### DIFF
--- a/packages/apollo-language-server/src/project/client.ts
+++ b/packages/apollo-language-server/src/project/client.ts
@@ -390,7 +390,7 @@ export class GraphQLClientProject extends GraphQLProject {
     for (const operationName in current) {
       const document = current[operationName];
 
-      let serviceOnly: DocumentNode = removeDirectiveAnnotatedFields(
+      let serviceOnly = removeDirectiveAnnotatedFields(
         removeDirectives(document, clientOnlyDirectives as string[]),
         clientSchemaDirectives as string[]
       );
--- a/packages/apollo-language-server/src/utilities/__tests__/graphql.test.ts
+++ b/packages/apollo-language-server/src/utilities/__tests__/graphql.test.ts
@@ -78,10 +78,6 @@ describe("withTypenameFieldAddedWhereNeeded", () => {
 });
 
 describe("removeDirectiveAnnotatedFields", () => {
-  it("should be a function", () => {
-    expect(typeof removeDirectiveAnnotatedFields).toBe("function");
-  });
-
   it("should remove fields with matching directives", () => {
     expect(
       print(
@@ -98,7 +94,7 @@ describe("removeDirectiveAnnotatedFields", () => {
         `);
   });
 
-  it("should remove object fields with matching directives", () => {
+  it("trim selections sets that are client only", () => {
     expect(
       print(
         removeDirectiveAnnotatedFields(
@@ -116,6 +112,83 @@ describe("removeDirectiveAnnotatedFields", () => {
     ).toMatchInlineSnapshot(`
       "query Query {
         fieldToKeep
+      }
+      "
+    `);
+  });
+
+  it("should remove fragments when a directive is used on a fragment spread", () => {
+    expect(
+      print(
+        removeDirectiveAnnotatedFields(
+          parse(`
+            {
+              me { name }
+              ...ClientFields @client
+            }
+            fragment ClientFields on Query {
+              hello
+            }
+          `),
+          ["client"]
+        )
+      )
+    ).toMatchInlineSnapshot(`
+        "{
+          me {
+            name
+          }
+        }
+        "
+    `);
+  });
+
+  it("should remove fragments when client directive is used inline", () => {
+    expect(
+      print(
+        removeDirectiveAnnotatedFields(
+          parse(`
+            {
+              me { name }
+              ... on Query @client {
+                hello
+              }
+            }
+          `),
+          ["client"]
+        )
+      )
+    ).toMatchInlineSnapshot(`
+            "{
+              me {
+                name
+              }
+            }
+            "
+        `);
+  });
+
+  it("should remove fragments when the client directive is on the definition", () => {
+    expect(
+      print(
+        removeDirectiveAnnotatedFields(
+          parse(`
+            fragment ClientObject on Query @client {
+              hello
+            }
+            {
+              me { name }
+              ... ClientObject
+            }
+          `),
+          ["client"]
+        )
+      )
+    ).toMatchInlineSnapshot(`
+      "{
+        me {
+          name
+        }
       }
       "
     `);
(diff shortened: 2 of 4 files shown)


packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -103,6 +113,18 @@ export function removeDirectiveAnnotatedFields(
         )
       )
         return null;
+
+      // We're keeping this field. Save the names of all fragment spreads that are in this field's selection
+      // set. This is intentionally _not_ recursive because descendents of this field may a directive that we
+      // want to remove; so we only know that this field's selection set is safe to keep.
+      if (node.selectionSet) {
+        node.selectionSet.selections.forEach(selection => {
```
jbaxleyiii: Instead of looking through selection sets (which would end up happening recursively due to the visitor btw as all `Fields` would be visited), we should visit on locations where fragments can be present to determine if they are impacted by the `@client` directive. The three locations can be found [here](https://github.com/graphql/graphql-js/blob/35c68d403cf1020e1d30ea29f0546352d03dd49e/src/language/kinds.js#L19-L21).

The other part of the story (which you are doing below!) is to trim the fragments that are no longer used) after collecting the impacted ones. Instead of checking for selection > @client > fragment spread, we should do a general check at the end for fragments orphaned by the trimming of selection sets done through field splitting.

Note that this will require doing a recursive lookup for fragments that could be impacted by the client directive (prior art https://github.com/graphql/graphql-js/blob/5c9946b0d0e2e082772db4ccd50013caa0b1ed79/src/validation/ValidationContext.js#L110)

```graphql
fragment One on Node {
  ...Two
  user {
    friends {
       name
      ...Two @client 
    }
  }
}
fragment Two on Node {
  id
}

query {
  me {
    ...One
  }
}
```

In the above example we *don't* want to trim `Two` as it is still used even though we *do* want to trim the usage of `Two` on `friends`.

In a similar issue, if a fragment is annotated as `@client` and contains fragments, we need to recurse and trim those fragments *if* they aren't used elsewhere  in the operation (i.e.):

```graphql
fragment One on Node @client {
  ... Two
}
fragment Two on Node {
  id
}
query {
  me { name ...One }
}
```

In the above query, all that should be left in the document is `query { me { name } }` as both `One` and `Two` should be trimmed by the removal

packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -88,12 +88,22 @@ export function removeDirectives(ast: ASTNode, directiveNames: string[]) {
 }
 
 // remove fields where a given directive is found
-export function removeDirectiveAnnotatedFields(
-  ast: ASTNode,
+export function removeDirectiveAnnotatedFields<AST extends ASTNode>(
```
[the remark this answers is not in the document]
justinanastos: `visit` returns `any`. This specifies that we're going to return the same thing as we're passed in and that thing must extend `ASTNode`.

I got errors from TypeScript for returning an `ASTNode` and trying to assign it to the more specific `DocumentNode`.

Deleting the change now I see that it's not _necessary_, but it also removes some `any`s. I had originally been working with modifying `removeDirectivess` as well, which would also benefit from this improved typing. Thoughts?

justinanastos:
@jbaxleyiii Great suggestions for new test cases and awesome explanations of what my first pass was missing. 

I've added the suggested test cases and another one to make sure we don't remove fragments that we didn't remove because of `@client` directives. I made the removal recursive as well.

packages/apollo-language-server/src/project/client.ts
```diff
@@ -385,12 +385,13 @@ export class GraphQLClientProject extends GraphQLProject {
...
     )
       return current;
+    console.warn({ clientOnlyDirectives, clientSchemaDirectives });
 
     const filtered = Object.create(null);
     for (const operationName in current) {
       const document = current[operationName];
 
-      let serviceOnly: DocumentNode = removeDirectiveAnnotatedFields(
+      let serviceOnly = removeDirectiveAnnotatedFields(
```
justinanastos: @jbaxleyiii I'm happy to walk back the <AST extends ASTNode>` types for `removeDirectiveAnnotatedFields`. With that change, we don't need to define the type here because it's inferred correctly.

packages/apollo-language-server/src/utilities/__tests__/graphql.test.ts
```diff
@@ -91,14 +87,14 @@ describe("removeDirectiveAnnotatedFields", () => {
         )
       )
     ).toMatchInlineSnapshot(`
-            "query Query {
-              fieldToKeep
-            }
-            "
-        `);
+                  "query Query {
```
justinanastos: Dear jest, stop changing my whitespace.

packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -87,31 +93,180 @@ export function removeDirectives(ast: ASTNode, directiveNames: string[]) {
...
+    FragmentSpread(node) {
+      fragmentSpreadNodeNames.add(node.name.value);
+    }
+  });
+
+  // Strip unused fragment definitions. Flag if we've removed any so we know if we need to continue
+  // recursively checking.
+  ast = visit(ast, {
+    FragmentDefinition: {
+      enter(node) {
```
jbaxleyiii: Why are you using `enter` here and `leave` below?

packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -87,31 +93,180 @@ export function removeDirectives(ast: ASTNode, directiveNames: string[]) {
...
+      ) {
+        anyNodesRemoved = true;
+        return null;
+      }
+
+      return undefined;
+    }
+  });
+
+  if (anyNodesRemoved) {
```
jbaxleyiii: This recursion shouldn't be needed since the visitor will recurse for you going top down. As soon as a node should be trimmed, it *can't* have any children to need to be trimmed

packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -87,31 +93,180 @@ export function removeDirectives(ast: ASTNode, directiveNames: string[]) {
...
+
+        return undefined;
+      }
+    }
+  });
+
+  if (anyFragmentsRemoved) {
+    // We've removed fragments and might have orphaned more fragments, so recursively try to remove more
+    // orphaned fragments.
+    return removeOrphanedFragmentDefinitions(
```
jbaxleyiii: So this is handling the case of:

```graphql
fragment Two on Node {
  id
}
fragment One on Query {
  hero {
    ...Two @client 
  }
}
{ ...One }
```
Where the visitor goes past the `Two` fragment first since it isn't know that it needs to be trimmed. After we run through and see that `One` has the only usage of `Two` and it is client only, it has now become orphaned.

If I understand this correctly, I think we should explain the recursion in a comment with a reference query like above

packages/apollo-language-server/src/utilities/graphql.ts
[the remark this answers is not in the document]
jbaxleyiii: I'm surprised to see that `visit` returns `any` but it sure does! https://github.com/graphql/graphql-js/blob/master/src/language/visitor.js#L226

I'm good with this as long as we add a comment to the function explaining the type casting

packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -87,31 +93,180 @@ export function removeDirectives(ast: ASTNode, directiveNames: string[]) {
...
-      )
+      ) {
+        // If we're removing a fragment definition then save the name so we can remove anywhere this fragment
+        // was spread
+        if (node.kind === Kind.FRAGMENT_DEFINITION) {
+          removedFragmentDefinitionNames.add(node.name.value);
+        }
+
+        // All nested fragment spreads inside of this definition are now eligible to be removed
+        visit(ast, {
```
jbaxleyiii: I don't think we should nest visitors here. This implementation doesn't do what I think you want it to anyway. By revisiting the root ast, instead of the specific node that you are in currently, you aren't "filtering" fragment spreads inside of this definition.

Since the visitor works depth first, you should be able to visit on `FragmentDefinition` to create your list, then visit the FragmentSpread next as long as you sort the document prior to doing the work so that fragment definitions always come first

justinanastos: Absolutely right.

packages/apollo-language-server/src/utilities/graphql.ts
justinanastos: I intended to remove all `FragmentDefinition` that are flagged for removal and then remove all associated spreads.

My understanding is that all the nodes would be visited with `enter` before _any_ would be visited by `leave`. Is that correct?

jbaxleyiii: Ah, not that isn't quite how it works.

```
{
  me {
    name
  }
  bar
}
```

Would be `enter(me), enter(name), leave(me), enter(bar)`

justinanastos: Ah! So this should be split into operations to accomplish what I was after. 👍

justinanastos: This makes less sense than I'd originally thought; there's no need to look at `FragmentSpread`s here because we're removing anything that isn't in the list generated above, which includes all `FragmentSpread`s.

packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -87,31 +99,178 @@ export function removeDirectives(ast: ASTNode, directiveNames: string[]) {
...
+    );
+  }
+
+  return ast;
+}
+
+/**
+ * Remove nodes that have zero-length selection sets
+ *
+ * The `ast` param must extend ASTNode. We use a genetic to indicate that this function returns the same type
```
jbaxleyiii: Generic*

packages/apollo-language-server/src/utilities/graphql.ts
jbaxleyiii: This appears to still be nested? Why was this resolved?

justinanastos: I'm sorry about that; I have no idea. I'll figure this last piece out on Monday! Thanks again @jbaxleyiii !

justinanastos: My original logic was to visit all the descendents of `node` to find all fragment spreads being removed because this field was being removed as well. I wrote this to visit `ast` by accident; great catch @jbaxleyiii .

I think that I do, however, want to use a nested `visitor` here to flag all the fragment spreads under this node as eligible for deletion. I've added a comment to hopefully explain this. Please let me know if this is the correct way to handle this problem.

PR #1462, opened 2019-08-06
[Swift] Fix build failures for optionals

designatednerd:
This PR fixes issues introduced by #656, which I prematurely had released. 🤦‍♀️

The issues are most obvious in [this failed run of the main iOS SDK](https://travis-ci.org/apollographql/apollo-ios/jobs/568034798), but the short version is that we weren't fully checking for optionals with that code, and the compiler was pretty angry about it.

### DIFF
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -1069,7 +1069,7 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
   public var friends: [Friend?]? {
     get {
       return (resultMap[\\"friends\\"] as? [ResultMap?]).flatMap { (value: [ResultMap?]) -> [Friend?] in value.map { (value: ResultMap?) -> Friend? in value.flatMap { (value: ResultMap) -> Friend in Friend(unsafeResultMap: value) } } }?.filter({
-        guard let type = $0 else { return false }; return Friend.possibleTypes.contains($type.__typename)
+        guard let type = $0 else { return false }; return Friend.possibleTypes.contains(type.__typename)
       })
     }
     set {
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -1069,7 +1069,7 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
   public var friends: [Friend?]? {
     get {
       return (resultMap[\\"friends\\"] as? [ResultMap?]).flatMap { (value: [ResultMap?]) -> [Friend?] in value.map { (value: ResultMap?) -> Friend? in value.flatMap { (value: ResultMap) -> Friend in Friend(unsafeResultMap: value) } } }?.filter({
-        guard let type = $0 else { return false}; return Friend.possibleTypes.contains($type.__typename)
+        guard let type = $0 else { return false }; return Friend.possibleTypes.contains($type.__typename)
       })
     }
     set {
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -1068,8 +1068,8 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
   /// The friends of the character, or an empty list if they have none
   public var friends: [Friend?]? {
     get {
-      return (resultMap[\\"friends\\"] as? [ResultMap?]).flatMap { (value: [ResultMap?]) -> [Friend?] in value.map { (value: ResultMap?) -> Friend? in value.flatMap { (value: ResultMap) -> Friend in Friend(unsafeResultMap: value) } } }.filter({
-        Friend.possibleTypes.contains($0.__typename)
+      return (resultMap[\\"friends\\"] as? [ResultMap?]).flatMap { (value: [ResultMap?]) -> [Friend?] in value.map { (value: ResultMap?) -> Friend? in value.flatMap { (value: ResultMap) -> Friend in Friend(unsafeResultMap: value) } } }?.filter({
+        guard let type = $0 else { return false}; return Friend.possibleTypes.contains($type.__typename)
       })
     }
     set {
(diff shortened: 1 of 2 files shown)


packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
```diff
@@ -1069,7 +1069,7 @@ exports[`Swift code generation #structDeclarationForFragment() should generate a
   public var friends: [Friend?]? {
     get {
       return (resultMap[\\"friends\\"] as? [ResultMap?]).flatMap { (value: [ResultMap?]) -> [Friend?] in value.map { (value: ResultMap?) -> Friend? in value.flatMap { (value: ResultMap) -> Friend in Friend(unsafeResultMap: value) } } }?.filter({
-        guard let type = $0 else { return false}; return Friend.possibleTypes.contains($type.__typename)
+        guard let type = $0 else { return false }; return Friend.possibleTypes.contains($type.__typename)
```
koenpunt: Seeing this I think you're not aware of how these snapshots are usually updated. 

If you run `npm test -- -u`, it will automatically write the new snapshots (after which you should of course check if they are correct), so you never have to write them by hand.

designatednerd: Unfortunately for reasons I don't totally understand, `npm test` is totally failing locally. Otherwise yes, that's what I should be doing

designatednerd:
Gonna get a look from @martijnwalraven later today, I know he had a couple concerns about it clobbering new conformances to a type on the server

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -727,12 +727,12 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
...
                 "ResultMap",
                 structName
-              )}.filter`
+              )}?.filter`
             );
             this.withinBlock(
               () =>
                 this.printOnNewline(
-                  `${structName}.possibleTypes.contains($0.__typename)`
+                  `guard let type = $0 else { return false }; return ${structName}.possibleTypes.contains(type.__typename)`
```
koenpunt: For readability it might be nice to have this on 2 lines instead.

```
guard let type = $0 else { return false };
return ${structName}.possibleTypes.contains(type.__typename)
```

koenpunt: Also, I'm trying to see how `$0` can be `nil` after being `flatMap`ped.. That said, we might want to replace `flatMap` with `compactMap` since `flatMap` is deprecated.

designatednerd: For some reason this was being real dumb about letting me use more than one line, and I really didn't feel like fighting with typescript at that moment. 

And I suspect it's *because* we're using `flatMap` where we should be using `compactMap`, but I'd like to hear some thoughts from @martijnwalraven before I rip this apart any further.

designatednerd: (OK i at least got this onto a separate line)

designatednerd:
@koenpunt @martijnwalraven I took a stab at getting the swift code to look a bit simpler just through turning off codegen and refactoring: 

Screen Shot 2019-08-06 at 1 54 23 PM

I think this should prevent crashes since `Friend` just gets the unsafe result map with the unknown type and you don't need to worry about the type in that instance. 

The problem is that the existing knot of `flatMap`ping is all done in [mapExpressionForType](https://github.com/apollographql/apollo-tooling/blob/1907dc399f689ee6b942e5d16dc8193f4f1bd6e4/packages/apollo-codegen-swift/src/helpers.ts#L265), which from the looks of the callers is doing a WHOLE bunch of other stuff I don't totally understand. 

Any suggestions on that would be most appreciated.

koenpunt:
> @koenpunt @martijnwalraven I took a stab at getting the swift code to look a bit simpler just through turning off codegen and refactoring:

That looks _way_ more readable!  But indeed `mapExpressionForType` covers a lot of things. Maybe if @martijnwalraven can shed some light on what all these methods (`typeNameFromGraphQLType`, `mapExpressionForType`, etc) do, that could really help. But given the fact it's about 2 years that code was written, it even might be hard for him to recall how everything is working.

koenpunt:
> I think this should prevent crashes since `Friend` just gets the unsafe result map with the unknown type and you don't need to worry about the type in that instance.

This doesn't crash when the value in the array is optional (gql; `friends: [Friend]`), but for required types (gql; `friends: [Friend!]!`) a `compactMap` would still be required I think.

martijnwalraven:
@koenpunt Can you help me understand what issue you’re trying to solve? Where does the crash when adding a new implementing type for an interface come from?

There is probably a bug in the current generated code. But I don't think filtering out objects of unknown types is the answer. Introducing a new object type should be backwards compatible if you’re accessing it through an interface. Let’s say you have:
```
interface ContentItem {
  title: String
}

type BlogPost implements ContentItem {
  title: String
}

type Query {
  favoriteItems: [ContentItem]
}
```
```
query {
  favoriteItems {
    title
  }
}
```
If the server then introduces:
```
type Poem implements ContentItem {
  title: String
}
```
and adds a new item of type `Poem` to the list of `favoriteItems`, it should still be returned from the query and be accessible as a `ContentItem` (so you could get its `title` like you would for any other implementation of that interface).

As you mentioned, it's been a while since I worked on this codebase, so it's very likely I made a mistake. But if you agree the above makes sense as the behavior we want, we can figure out what changes we need to make to the generated code.

martijnwalraven:
> > @koenpunt @martijnwalraven I took a stab at getting the swift code to look a bit simpler just through turning off codegen and refactoring:
> 
> That looks _way_ more readable! But indeed `mapExpressionForType` covers a lot of things. Maybe if @martijnwalraven can shed some light on what all these methods (`typeNameFromGraphQLType`, `mapExpressionForType`, etc) do, that could really help. But given the fact it's about 2 years that code was written, it might even be hard for him to recall how everything is working.

One of the reasons `mapExpressionForType` is complicated is because it has to deal with nested types, like a non-null list of nullable lists of non-null elements (e.g. `[[BlogPost!]]!`). We may be able to simplify the generated code, but having it output expressions instead of intermediate variables makes it a lot easier to deal with nesting because we can just call `mapExpressionForType` recursively.

koenpunt:
> Where does the crash when adding a new implementing type for an interface come from?

The crash is not with interfaces (or at least not what I experienced), but with union types.

So when a schema like this:

```
union ContentItem = BlogPost

type BlogPost {
  title: String
}

type Query {
  favoriteItems: [ContentItem]
}
```

Gets updated to;

```
union ContentItem = BlogPost | Image

type Image {
  url: String!
}

type BlogPost {
  title: String
}

type Query {
  favoriteItems: [ContentItem]
}
```

martijnwalraven:
@koenpunt There isn't really a difference between interfaces and unions in terms of the generated code. And they are in fact very similar in GraphQL, the only difference is where the implementing types are declared. So it seems what I wrote above should apply just as well to unions.

I'm wondering what specifically is causing the crash. Which line in the generated code is responsible?

designatednerd:
@koenpunt Would really appreciate it if you could let us know - I may wind up needing to temporarily revert #656 on Master if we can't get this figured out in the next day or two as it's blocking a release on the iOS SDK.

koenpunt:
> I'm wondering what specifically is causing the crash. Which line in the generated code is responsible?

Can't recall, it's been 9 months :) and the schema changed in the meantime, so the problem currently doesn't exist in our project. I will create a reproduction project, but in the meantime you could revert for now, since no one else reported this issue.

designatednerd:
OK @koenpunt I'm going to close this one out - can you please open a new issue and tag me when you're able to repro? Thank you!

koenpunt:
@designatednerd I'm not able to reproduce it atm, so it might be that it has been fixed (accidentally) with some recent refactor. Or I'm not recreating the same scenario I had in the past, but that is something I can't know for sure.

PR #1464, opened 2019-08-07
Revert PR #656

designatednerd:
This PR reverts changes from #656, they were causing build issues on iOS and the originator is going to come up with a new reproduction before we take another stab at this. 

This will need to go out as a patch release for `swift` and `tooling` since I ~~'m a moron~~ got a little ahead of myself and didn't fully test these locally before getting them released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@
   - <First `apollo-codegen-flow` related entry goes here>
 - `apollo-codegen-scala`
   - <First `apollo-codegen-scala` related entry goes here>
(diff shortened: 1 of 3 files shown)


designatednerd:
Note that it's `+5 / -17` rather than `+3 / -15` to match the other PR due to the update to the changelog.

abernix:
Released in the following packages.

```
Successfully published:                                                                                                    
 - apollo-codegen-swift@0.35.2                                                                                             
 - apollo@2.17.2 
```

PR #1466, opened 2019-08-08
Error with duplicate operation names

### LINKED ISSUE #1427: codegen: output warning if query type generation is suppressed because of naming conflict

JakeDawkins:
Currently, the `apollo-language-server` requires operation names in a client project to be unique in order to properly generate types, but this isn't enforced by an error. This PR adds an error so the user is aware of a name collision. 

When generating a type using codegen, if there are duplicate type names, only one of the operations will get types generated. this error will enforce the already designed intention of the CLI.

The following screenshot shows the error as well as showing that **the error doesn't interfere with `--watch` mode**.

Screen Shot 2019-08-08 at 7 47 37 PM

resolves #1427

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
   - <First `apollo-graphql` related entry goes here>
 - `apollo-language-server`
   - Add debugging util classes for better error/warning handling [#1429](https://github.com/apollographql/apollo-tooling/pull/1429)
-  - Add warning for duplicate client operation names [#1466](https://github.com/apollographql/apollo-tooling/pull/1466)
+  - Add error for duplicate client operation names [#1466](https://github.com/apollographql/apollo-tooling/pull/1466)
 - `apollo-tools`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -19,7 +19,7 @@
 - `apollo-graphql`
   - <First `apollo-graphql` related entry goes here>
 - `apollo-language-server`
-  - Fix issue where fragment definitions only included in `@client` fields would not be stripped ((AP-682)(https://golinks.io/AP-682), [#1454](https://github.com/apollographql/apollo-tooling/pull/1454))
```
justinanastos: I think this happened by merging instead of rebasing? This will delete my changelog entry!

PR #1482, opened 2019-08-23
Fix for swift 5.1 compiler warning when type is an enum with a none case

cltnschlosser:
Resolves https://github.com/apollographql/apollo-ios/issues/724

cc @designatednerd


designatednerd:
This looks good in theory but I want to test this against our existing codebase before I merge. I'm out moving until 3 September so it'll probably be around then, just FYI

cltnschlosser:
Yep, no problem. I saw that you were out, just thought I’d throw up the pull request anyway :)

cltnschlosser:
@designatednerd Have you had a chance to look at this?

cltnschlosser:
@designatednerd Updated it, but any advice on those CI failures. Seem to be completely unrelated.

designatednerd:
Those tests are super tempermental, I kicked both builds

designatednerd:
@cltnschlosser Mind pushing a dummy commit? Looks like our CI may have temporarily lost its mind and needs a kick at the push level.

cltnschlosser:
I pushed an empty file and that build passed, then I dropped that commit and the build failed :(

PR #1501, opened 2019-09-05
Automate publishing via Git tag triggered CircleCI workflow.

abernix:
In the same spirit as the following PRs from Apollo Server:

* https://github.com/apollographql/apollo-server/pull/3246
* https://github.com/apollographql/apollo-server/pull/3245
* https://github.com/apollographql/apollo-server/pull/3254

Check out those PRs for more details!

But, in a nutshell:

1. `npm run release:version-bump`
   
   This should behave the same as `npm run release` did previously, except rather than doing the actual publish, it will only bump the versions and push the tags to CircleCI.

2. `npm run release:start-ci-publish`

   
   This will create and push a `publish/{timestamp}` tag to CircleCI which will kick off the actual publish pipeline.

...and the rest should happen in CircleCI!  A notification will be sent to Slack when the _Confirmation_ step is pending approval, and also when the publish is fully complete.  Additionally, a message will be sent if an error was encountered.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2.1
 orbs:
   # `oss` is a local reference to the package.  The source for Apollo Orbs can
   # be found at http://github.com/apollographql/CircleCI-Orbs.
-  oss: apollo/oss-ci-cd-tooling@0.0.3
+  oss: apollo/oss-ci-cd-tooling@0.0.4
 
 commands:
 # These are the steps used for each version of Node which we're testing
@@ -18,20 +18,7 @@ commands:
     steps:
       - oss/install_specific_npm_version
       - checkout
-      - restore_cache:
-          keys:
-            # When lock file changes, use increasingly general patterns to restore cache
-            - npm-v2-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "package-lock.json" }}
-            - npm-v2-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-
-            - npm-v2-{{ .Environment.CACHE_VERSION }}-
-      - run: npm --version
-      - run: npm ci
-      - save_cache:
-          key: npm-v2-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "package-lock.json" }}
-          paths:
-            # This should cache the npm cache instead of node_modules, which is needed because
-            # npm ci actually removes node_modules before installing to guarantee a clean slate.
-            - ~/.npm
+      - oss/npm_clean_install_with_caching
       - run: npm run circle
 
 # Important! When adding a new job to `jobs`, make sure to define when it
@@ -57,7 +44,6 @@ jobs:
           paths:
             - ./**
 
-
   VSCode:
     executor: { name: oss/node }
     steps:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1,76 +1,58 @@
...
-#
-run_install_desired_npm: &run_install_desired_npm
-  run:
-    name: Install npm@6.
-    command: sudo npm install -g npm@6
+# These "CircleCI Orbs" are reusable bits of configuration that can be shared
+# across projects.  See https://circleci.com/orbs/ for more information.
+orbs:
+  # `oss` is a local reference to the package.  The source for Apollo Orbs can
+  # be found at http://github.com/apollographql/CircleCI-Orbs.
```
abernix: This comment points to the right place, but I'm just drawing extra attention to it so the source for these can be more easily found.

.circleci/config.yml
```diff
@@ -1,76 +1,58 @@
...
       - run:
           name: Install deps for VSCode / GUI
           command: sudo apt-get update && sudo apt-get install -y nodejs libgtk3.0 libxss1 libgconf-2-4 libasound2 libnss3 x11-xserver-utils xvfb
-      - *run_install_desired_npm
+      - oss/install_specific_npm_version
       - checkout
-      - restore_cache:
-          keys:
-            # When lock file changes, use increasingly general patterns to restore cache
-            - npm-v2-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "package-lock.json" }}
```
abernix: Note, while I think it's been _ages_ since we've needed to bump the `CACHE_VERSION`, I've moved to a new pattern which leverages a `CIRCLECI_CACHE_VERSION`.  It's still set via the _environment variables_ for the project in CircleCI, and I've already updated it to `v1` (the old `v1` is very much expired by now.)

.circleci/config.yml
```diff
@@ -82,68 +64,88 @@ jobs:
             DISPLAY: :99
   # Other tests, unrelated to typical code tests.
   Linting:
-    docker: [{ image: "circleci/node:8" }]
+    executor: { name: oss/node }
```
abernix: These now use a newer version of `node`.  The "default" for our orbs is Node.js 10.

.circleci/config.yml
```diff
@@ -82,68 +64,88 @@ jobs:
...
-      - Node.js 8
-      - Node.js 10
-      - VSCode
-      - Linting
-      - Query Check
-      - Generated Types Check
+      - NodeJS 8:
+          <<: *common_non_publish_filters
+      - NodeJS 10:
+          <<: *common_non_publish_filters
```
abernix: We should probably start testing Node.js 12, but that can be put in another PR.

lerna.json
```diff
@@ -2,5 +2,10 @@
   "packages": [
     "packages/*"
   ],
-  "version": "independent"
+  "version": "independent",
+  "command": {
+    "version": {
+      "includeMergedTags": true
```
abernix: More on this addition is included in the comment for the commit that introduced it: https://github.com/apollographql/apollo-tooling/pull/1501/commits/873349fdcbfb654329f159af91013425fe7f6cce

package.json
```diff
@@ -29,7 +29,8 @@
     "lint": "prettier --list-different \"packages/*/src/**/*.{js,jsx,ts,tsx}\"",
     "lint-fix": "prettier --write \"packages/*/src/**/*.{js,jsx,ts,tsx}\"",
     "posttest": "npm run lint",
-    "release": "npm run clean && npm ci && lerna publish --exact",
+    "release:version-bump": "lerna version --exact -m 'Release'",
+    "release:start-ci-publish": "node -p '`Publish (dist-tag:${process.env.APOLLO_DIST_TAG || \"latest\"})`' | git tag -F - \"publish/$(date -u '+%Y%m%d%H%M%S')\" && git push origin \"$(git describe --match='publish/*' --tags --exact-match HEAD)\"",
```
abernix: Note the scripts are named differently, but they behave a bit differently, so it seemed worth not re-using the same name.

PR #1504, opened 2019-09-06
Document codegen requirement about SDL file extension

mrclay:


packages/apollo/README.md
```diff
@@ -642,6 +643,10 @@ When using the codegen command with Typescript or Flow, make sure to add the `__
 
 If you're using a client like `apollo-client` that does this automatically for your GraphQL operations, pass in the `--addTypename` option to `apollo codegen:generate` to make sure the generated Typescript and Flow types have the `__typename` field as well. This is required to ensure proper type generation support for `GraphQLUnionType` and `GraphQLInterfaceType` fields.
 
+## Default `--includes` parameter
```
JakeDawkins: I think this would be better under the `--includes` flag description. Same for excludes. You can find these flag descriptions in `packages/apollo/src/Command.ts`

packages/apollo/README.md
```diff
@@ -143,7 +143,8 @@ OPTIONS
 
   --key=key                                  The API key for the Apollo Engine service
 
-  --localSchemaFile=localSchemaFile          Path to your local GraphQL schema file (introspection result or SDL)
+  --localSchemaFile=localSchemaFile          Path to your local GraphQL schema file (introspection result or SDL as
```
JakeDawkins: changing the description of a flag actually doesn't happen here! This portion of the file is generated by oclif. You'd need to change this in `packages/apollo/src/commands/client/codegen.ts` along with changing any other uses of the `localSchemaFile` flag, like in the service commands as well.

packages/apollo/README.md
```diff
@@ -217,7 +218,7 @@ OPTIONS
 
   --header=header                        Additional header to send to server for introspectionQuery. May be used
                                          multiple times to add multiple headers. NOTE: The `--endpoint` flag is REQUIRED
-                                         if using the `--header` flag.
+                                         if using the `--header` flag. NOTE: `--header` flags must preceed `--endpoint`.
```
JakeDawkins: I'm not certain this is true. How are you using the multiple header flags? If so, I imagine this would be a bug, not a documentation fix.

zionts:
@mrclay thanks for opening this up! Having better documentation about how to use `localSchemaFile` would definitely be nice, and I'm also curious if this tripped you up because we were using another extension. In other words, should additional schema file extensions be supported rather than documenting the necessity of having them be `.graphql`?

PR #1506, opened 2019-09-06
Update codegen types

trevor-scheer:
Update generated types due to schema changes.

This is a result of:
`npm run client:codegen`


trevor-scheer:
Effectively a no-op, just description updates

PR #1515, opened 2019-09-09
Fix Swift codegen escaping for identifiers, types, and strings

lilyball:
The source generator now operates using a string wrapper type called `SwiftSource`. Identifiers are escaped when converting into `SwiftSource` values, with an alternative method for producing escaped string literals. An escape hatch is provided for disabling escapes, but the default behavior is to escape all identifiers in the dynamic input.

Most `SwiftSource` values are produced by a tagged template literal, which allows for easy mixing of string literals containing Swift keywords and dynamic input that needs escaping.

As part of this, rewrite string escaping such that it actually escapes string contents properly. This required fixing a test that had a bad spec enforcing broken behavior.

Fixes apollographql/apollo-ios#193.
Fixes apollographql/apollo-ios#752.

---

This PR isn't quite done yet. I haven't updated the changelog, and I also haven't added tests for apollographql/apollo-ios#193 that fail without this change. I'm submitting this now because it's almost 1AM and I'm going to bed, and I'd like to get some eyes on it.

TODO:

### DIFF
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -1167,6 +1167,7 @@ exports[`Swift code generation #structDeclarationForSelectionSet() should escape
 
   public static let selections: [GraphQLSelection] = [
     GraphQLField(\\"name\\", alias: \\"private\\", type: .nonNull(.scalar(String.self))),
+    GraphQLField(\\"friends\\", alias: \\"self\\", type: .list(.object(\`Self\`.selections))),
   ]
 
   public private(set) var resultMap: ResultMap
@@ -1175,12 +1176,12 @@ exports[`Swift code generation #structDeclarationForSelectionSet() should escape
     self.resultMap = unsafeResultMap
   }
 
-  public static func makeHuman(\`private\`: String) -> Hero {
-    return Hero(unsafeResultMap: [\\"__typename\\": \\"Human\\", \\"private\\": \`private\`])
+  public static func makeHuman(\`private\`: String, \`self\`: [\`Self\`?]? = nil) -> Hero {
+    return Hero(unsafeResultMap: [\\"__typename\\": \\"Human\\", \\"private\\": \`private\`, \\"self\\": \`self\`.flatMap { (value: [\`Self\`?]) -> [ResultMap?] in value.map { (value: \`Self\`?) -> ResultMap? in value.flatMap { (value: \`Self\`) -> ResultMap in value.resultMap } } }])
   }
 
-  public static func makeDroid(\`private\`: String) -> Hero {
-    return Hero(unsafeResultMap: [\\"__typename\\": \\"Droid\\", \\"private\\": \`private\`])
+  public static func makeDroid(\`private\`: String, \`self\`: [\`Self\`?]? = nil) -> Hero {
+    return Hero(unsafeResultMap: [\\"__typename\\": \\"Droid\\", \\"private\\": \`private\`, \\"self\\": \`self\`.flatMap { (value: [\`Self\`?]) -> [ResultMap?] in value.map { (value: \`Self\`?) -> ResultMap? in value.flatMap { (value: \`Self\`) -> ResultMap in value.resultMap } } }])
   }
 
   /// The name of the character
@@ -1190,6 +1191,48 @@ exports[`Swift code generation #structDeclarationForSelectionSet() should escape
     }
     set {
       resultMap.updateValue(newValue, forKey: \\"private\\")
+    }
+  }
+
+  /// The friends of the character, or an empty list if they have none
+  public var \`self\`: [\`Self\`?]? {
+    get {
+      return (resultMap[\\"self\\"] as? [ResultMap?]).flatMap { (value: [ResultMap?]) -> [\`Self\`?] in value.map { (value: ResultMap?) -> \`Self\`? in value.flatMap { (value: ResultMap) -> \`Self\` in \`Self\`(unsafeResultMap: value) } } }
+    }
+    set {
+      resultMap.updateValue(newValue.flatMap { (value: [\`Self\`?]) -> [ResultMap?] in value.map { (value: \`Self\`?) -> ResultMap? in value.flatMap { (value: \`Self\`) -> ResultMap in value.resultMap } } }, forKey: \\"self\\")
+    }
+  }
+
+  public struct \`Self\`: GraphQLSelectionSet {
+    public static let possibleTypes = [\\"Human\\", \\"Droid\\"]
+
+    public static let selections: [GraphQLSelection] = [
+      GraphQLField(\\"id\\", type: .nonNull(.scalar(GraphQLID.self))),
+    ]
+
+    public private(set) var resultMap: ResultMap
+
+    public init(unsafeResultMap: ResultMap) {
+      self.resultMap = unsafeResultMap
+    }
+
+    public static func makeHuman(id: GraphQLID) -> \`Self\` {
+      return \`Self\`(unsafeResultMap: [\\"__typename\\": \\"Human\\", \\"id\\": id])
+    }
+
+    public static func makeDroid(id: GraphQLID) -> \`Self\` {
+      return \`Self\`(unsafeResultMap: [\\"__typename\\": \\"Droid\\", \\"id\\": id])
+    }
+
+    /// The ID of the character
+    public var id: GraphQLID {
+      get {
+        return resultMap[\\"id\\"]! as! GraphQLID
+      }
+      set {
+        resultMap.updateValue(newValue, forKey: \\"id\\")
+      }
     }
   }
 }"
--- a/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
+++ b/packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
@@ -108,7 +108,7 @@ exports[`Swift code generation #classDeclarationForOperation() should correctly 
   ///   }
   /// }
   public let operationDefinition =
-    \\"mutation CreateReview($episode: Episode) {\\\\n  createReview(episode: $episode, review: {stars: 5, commentary: \\\\\\"\\\\\\"\\\\\\"\\\\n    Wow!\\\\n     I thought\\\\n      This movie \\\\ ROCKED!\\\\n  \\\\\\"\\\\\\"\\\\\\"}) {\\\\n    stars\\\\n    commentary\\\\n  }\\\\n}\\"
+    \\"mutation CreateReview($episode: Episode) {\\\\n  createReview(episode: $episode, review: {stars: 5, commentary: \\\\\\"\\\\\\"\\\\\\"\\\\n    Wow!\\\\n     I thought\\\\n      This movie \\\\\\\\ ROCKED!\\\\n  \\\\\\"\\\\\\"\\\\\\"}) {\\\\n    stars\\\\n    commentary\\\\n  }\\\\n}\\"
 
   public let operationName = \\"CreateReview\\"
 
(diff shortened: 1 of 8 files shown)


packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -420,15 +427,20 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
...
         this.printNewlineIfNeeded();
-        this.printOnNewline("public static let possibleTypes = [");
+        this.printOnNewline(swift`public static let possibleTypes = [`);
         this.print(
-          join(variant.possibleTypes.map(type => `"${type.name}"`), ", ")
+          join(
+            variant.possibleTypes.map(
+              type => swift`${SwiftSource.string(type.name)}`
+            ),
+            ", "
```
designatednerd: Am I correct that this doesn't need `SwiftSource.string` since as the separator parameter, it needs to be a string literal?

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -55,38 +43,263 @@ const reservedKeywords = new Set(['associatedtype', 'class', 'deinit', 'enum', '
...
+   */
+  static string(string: string, trim: boolean = false): SwiftSource {
+    if (trim) {
+      string = string
+        .split(/\n/g)
+        .map(line => line.trim())
+        .join(" ");
+    }
+    return new SwiftSource(
+      `"${string.replace(/[\0\\\t\n\r"]/g, c => {
```
designatednerd: Would you mind throwing in a comment for the regex-impared explaining what's being replaced here?

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -55,38 +43,263 @@ const reservedKeywords = new Set(['associatedtype', 'class', 'deinit', 'enum', '
   'indirect', 'lazy', 'left', 'mutating', 'none', 'nonmutating', 'optional',
   'override', 'postfix', 'precedence', 'prefix', 'Protocol', 'required', 'right',
   'set', 'Type', 'unowned', 'weak', 'willSet']);
+const reservedMemberKeywords = new Set(["self", "Type", "Protocol"]);
+
+/**
+ * A class that represents Swift source.
+ *
+ * Instances of this type will not undergo escaping when used with the `swift` template tag.
+ */
```
designatednerd: 🙏 [deity or lack thereof of choice] bless you for adding doc comments 🙏

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -55,38 +43,263 @@ const reservedKeywords = new Set(['associatedtype', 'class', 'deinit', 'enum', '
...
+    return this.source;
+  }
+
+  /**
+   * Concatenates multiple `SwiftSource`s together.
+   */
+  concat(...sources: SwiftSource[]): SwiftSource {
+    // Documentation says + is faster than String.concat, so let's use that
+    var result = this.source;
+    for (let elt of sources) {
```
designatednerd: silly question: What does `elt` stand for?

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -55,38 +43,263 @@ const reservedKeywords = new Set(['associatedtype', 'class', 'deinit', 'enum', '
...
+
+  toString(): string {
+    return this.source;
+  }
+
+  /**
+   * Concatenates multiple `SwiftSource`s together.
+   */
+  concat(...sources: SwiftSource[]): SwiftSource {
+    // Documentation says + is faster than String.concat, so let's use that
```
designatednerd: Moderately less silly question: Would this work with Typescript `reduce`, or is that only something that works in Swift?

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -55,38 +43,263 @@ const reservedKeywords = new Set(['associatedtype', 'class', 'deinit', 'enum', '
...
   constructor(context: Context) {
     super(context);
   }
 
   multilineString(string: string) {
-    this.printOnNewline(`"${escapedString(string)}"`);
+    // Disable trimming if the string contains """ as this means we're probably printing an
+    // operation definition where trimming is destructive.
+    this.printOnNewline(
+      SwiftSource.string(string, /* trim */ !string.includes('"""'))
```
designatednerd: the inline comment here is to indicate what that parameter is, correct?

packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
```diff
@@ -108,7 +108,7 @@ exports[`Swift code generation #classDeclarationForOperation() should correctly
   ///   }
   /// }
   public let operationDefinition =
-    \\"mutation CreateReview($episode: Episode) {\\\\n  createReview(episode: $episode, review: {stars: 5, commentary: \\\\\\"\\\\\\"\\\\\\"\\\\n    Wow!\\\\n     I thought\\\\n      This movie \\\\ ROCKED!\\\\n  \\\\\\"\\\\\\"\\\\\\"}) {\\\\n    stars\\\\n    commentary\\\\n  }\\\\n}\\"
+    \\"mutation CreateReview($episode: Episode) {\\\\n  createReview(episode: $episode, review: {stars: 5, commentary: \\\\\\"\\\\\\"\\\\\\"\\\\n    Wow!\\\\n     I thought\\\\n      This movie \\\\\\\\ ROCKED!\\\\n  \\\\\\"\\\\\\"\\\\\\"}) {\\\\n    stars\\\\n    commentary\\\\n  }\\\\n}\\"
```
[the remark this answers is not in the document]
lilyball: I'm not familiar with snap but it looks like everything here is double-escaped, right? So the old output ended up evaluating to `This movie \ ROCKED!` which is of course bogus because that suggests we're escaping a space.

That was actually how the old code worked since our `escapedString` function wasn't escaping backslashes, it was only escaping quotes and converting newlines into `\n`. So a backslash in the input string would stay as an unescaped backslash in the output string, and therefore be interpreted as an escape by Swift (such that the backslash isn't present in the runtime value of the string).

The new version of `escapedString` actually escapes everything that's supposed to be escaped, so the string literal now reads `This movie \\ ROCKED!`, and therefore the backslash is preserved in the runtime string value.

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -19,13 +19,14 @@ import {
   Field
 } from "apollo-codegen-core/lib/compiler";
 
-import { join, wrap } from "apollo-codegen-core/lib/utilities/printing";
+// import { join, wrap } from "apollo-codegen-core/lib/utilities/printing";
```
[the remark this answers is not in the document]
lilyball: Yeah I meant to nuke this, thanks for the reminder.

packages/apollo-codegen-swift/src/codeGeneration.ts
lilyball: Yeah. I put a note to that effect on the doc comment for this function; I left the separator as `string` for convenience as it's expected to just be something like `", "`. I originally left the start/end strings for `wrap` as that too, until I discovered a call to `wrap` that used string interpolation in the `start` parameter. Having the separator to `join` be ` string` is technically unsafe because you could interpolate something into it, but I thought the convenience outweighed that here as I can't think of a non-contrived scenario where I'd interpolate an identifier into the separator.

packages/apollo-codegen-swift/src/language.ts
lilyball: To be honest, this is a pretty simple regex. It's just matching any of 6 characters. I suppose I could say something like `// Match any character that needs escaping in Swift strings` but I thought that was inferable from context.

packages/apollo-codegen-swift/src/__tests__/__snapshots__/codeGeneration.ts.snap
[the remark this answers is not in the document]
designatednerd: Yeah this stuff with the "It needs to be escaped exactly double the number of times" thing is super-confusing. Thanks for the explanation!

packages/apollo-codegen-swift/src/language.ts
lilyball: element.

packages/apollo-codegen-swift/src/language.ts
designatednerd: It is, but I think maybe what threw me is I'm not really sure what's being escaped with `\0`, or why a single quote isn't being escaped. The other bits are pretty straightforward.

packages/apollo-codegen-swift/src/language.ts
lilyball: This *absolutely* could be `Array.reduce` and I will make that change. I'm actually not normally a JavaScript/TypeScript programmer so I don't know offhand what functional methods are in the standard library.

packages/apollo-codegen-swift/src/language.ts
lilyball: Yep. Would you prefer I remove it, or use a different style?

packages/apollo-codegen-swift/src/language.ts
designatednerd: Ah - is that a standard Typescript thing? I obviously wasn't familiar with it and probably would have just used `element` personally, but if it's idiomatic typescript it should probably stay as is.

packages/apollo-codegen-swift/src/language.ts
lilyball: Technically all I really need to escape is `\n`, `\r`, `"`, and `\\`, but I included `\0` (the NUL character) and `\t` because Swift has defined escapes for those and that makes the source easier to read. Single quote isn't escaped because it's not a special character in a Swift string literal.

packages/apollo-codegen-swift/src/language.ts
designatednerd: Nah, I just hadn't seen that particular syntax before

packages/apollo-codegen-swift/src/language.ts
designatednerd: Yeah neither am I, clearly.

packages/apollo-codegen-swift/src/language.ts
lilyball: I'm not a TypeScript programmer normally, I actually don't know what the standard conventions are. `elt` is a fairly common language-agnostic shorthand for "element", but I don't know if it's idiomatic to use abbreviated variable names or not in TypeScript code.

packages/apollo-codegen-swift/src/language.ts
designatednerd: OK - maybe if there's someplace where those are defined it might be worth throwing in a comment link to it. Otherwise, this is fine.

lilyball: Good idea, I can put in a link to the Swift language reference.

packages/apollo-codegen-swift/src/language.ts
designatednerd: FWIW I mostly do Swift / Obj-C / Java / Kotlin, and I'd personally never seen that abbreviation before. It'd probably be good to expand to `element` for clarity for others who aren't typically typescript programmers (which unfortunately is the majority of folks working with this part of the codebase). However if you'd prefer to leave it, it's at least clear enough from what it does what the intention is.

lilyball:
Force-pushed to fix the flagged issues (including renaming `elt` to `value`), though it looks like there's now a conflict.

lilyball:
Updated again, this time with a test for unescaped types (similar to what was encountered in https://github.com/apollographql/apollo-ios/issues/193), as well as the changelog. This PR should now be good to go.

designatednerd:
@JakeDawkins Can you confirm the failing circle tests are an issue on our end? It certainly looks like it with the error ` ›   "Loading schema for engine": Error: ENGINE_API_KEY not found`, but wanted to double check before merging this

designatednerd:
Failing checks are due to this coming from a fork - tooling team is on it. Merging this!

lilyball:
Oh good, now that this has been released, I can state that it did not introduce any performance regressions, which I was afraid it might (due to replacing strings with objects). But running `apollo@2.18.0` against my codebase takes approximately the same amount of time as `apollo@2.18.3` (that time being ~51 seconds, which is rather unfortunate).

PR #1518, opened 2019-09-09
Replace instanceof checks with their respective predicates

trevor-scheer:
This replaces all `instanceof GraphQLXYZ` checks with their respective predicate functions.

TODO:

### DIFF
--- a/packages/apollo-codegen-scala/src/codeGeneration.ts
+++ b/packages/apollo-codegen-scala/src/codeGeneration.ts
@@ -3,8 +3,10 @@ import {
   getNamedType,
   isCompositeType,
   GraphQLEnumType,
-  GraphQLNonNull,
-  GraphQLInputObjectType
+  GraphQLInputObjectType,
+  isNonNullType,
+  isEnumType,
+  isInputObjectType
 } from "graphql";
 
 import { isTypeProperSuperTypeOf } from "apollo-codegen-core/lib/utilities/graphql";
@@ -157,9 +159,7 @@ export function classDeclarationForOperation(
             undefined,
             true
           );
-          const isOptional = !(
-            type instanceof GraphQLNonNull || type instanceof GraphQLNonNull
-          );
+          const isOptional = !isNonNullType(type);
           return { name, propertyName, type, typeName, isOptional };
         });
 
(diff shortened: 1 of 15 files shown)


packages/apollo-codegen-scala/src/codeGeneration.ts
```diff
@@ -157,9 +159,7 @@ export function classDeclarationForOperation(
             undefined,
             true
           );
-          const isOptional = !(
-            type instanceof GraphQLNonNull || type instanceof GraphQLNonNull
```
JakeDawkins: Lol what even was this?

packages/apollo-language-server/src/languageProvider.ts
```diff
@@ -178,9 +178,7 @@ export class GraphQLLanguageProvider {
         if (!suggestedField) {
           return suggest;
         } else {
-          const requiredArgs = suggestedField.args.filter(
-            a => a.type instanceof GraphQLNonNull
-          );
+          const requiredArgs = suggestedField.args.filter(isNonNullType);
```
JakeDawkins: Is this right, or would you need to check the args' `arg.type` like the old check was doing?

packages/apollo-codegen-scala/src/codeGeneration.ts
trevor-scheer: @shadaj this is way back in history but can you recall if this was intentional or possibly a mistake?
Re: https://github.com/apollographql/apollo-tooling/commit/58f07c42be6f817b646cef002a366327683e0028#diff-508d6e2cd0e11941b1548a2ea48a909dR139

packages/apollo-language-server/src/languageProvider.ts
trevor-scheer: Good catch!

packages/apollo-codegen-scala/src/codeGeneration.ts
shadaj: Looks like a mistake! I really hope there's no situation in which double-checking a type is actually needed 😄

JakeDawkins: @shadaj Was removing the second check the mistake, or was that correct? Like this was originally

`const isOptional = !(type instanceof GraphQLNonNull || type.ofType instanceof GraphQLNonNull);`

was the `ofType` incorrectly removed, or was that fine to remove?

shadaj: It was fine to remove the extra check, I believe that `ofType` was eliminated with changes to the IR format so only the top level type needed to be checked.

PR #1525, opened 2019-09-13
Support loading env variables through .env.local

### LINKED ISSUE #1058: VSCode: ENGINE_API_KEY not found in in Vue CLI project
Apollo GraphQL Visual Studio Code plugin loads schema from Apollo Engine.

The plugin fails to find my `ENGINE_API_KEY`

In a Vue CLI project, [the `.env` file is a committed file](https://cli.vuejs.org/guide/mode-and-env.html), where common variables should be stored. `.env.local` is a git ignored file where private variables should be defined. Since the Apollo GraphQL VSCode plugin looks in the `.env` file only, it can't find the `ENGINE_API_KEY` as defined in my `.env.local` file.

I'm proposing that the code (included below) that loads the env file needs to change to make it [compatible with Vue CLI projects](https://cli.vuejs.org/guide/mode-and-env.html). Depending on how you want to change it to make it flexible, it might look something like checking all files that start with `.env`, although I can see potential issues there.

https://github.com/apollographql/apollo-tooling/blob/4a27c3323ea243706ce12b0b29a85366000e4bbb/packages/apollo-language-server/src/config/loadConfig.ts#L88-L104

For inspiration, here's how Vue CLI does it:

https://github.com/vuejs/vue-cli/blob/bd57f15a15303d7ae32eb4961d9418b76593189c/packages/%40vue/cli-service/lib/Service.js#L63-L70

https://github.com/vuejs/vue-cli/blob/bd57f15a15303d7ae32eb4961d9418b76593189c/packages/%40vue/cli-service/lib/Service.js#L92-L133

Apollo GraphQL Visual Studio Code 1.5.2

gwardwell:
Closes #1058 

Vue CLI uses a unique ENV file pattern where `.env` files are committed, and local untraced ENV variables are added to a `.env.local` file. This PR updates the env loader logic so it looks for both the `.env` and `.env.local` files.

- If the loader find a `.env` file, it will try to load that file and extract the API key
- If the loader find a `.env.local` file, it will try to load that file and extract the API key
- If the loader find both a `.env` file and a `.env.local` file, it will first try to load the `.env` file and extract the API key, then it will try to load the `.env.local` file and will prefer an API key found in `.env.local` over an API key found in the `.env` file

TODO:


JakeDawkins:
Wonderful! I pulled this down, built it and everything seemed to work for me :) Thanks for working on this and for your patience @gwardwell

coler-j:
Just a note, as I discovered this today:

The behaviour listed in this PR "If the loader find both a `.env` file and a `.env.local` file, it will first try to load the `.env` file and extract the API key, then it will try to load the `.env.local` file and will prefer an API key found in `.env.local` over an API key found in the `.env` file" isn't accurate. 

If `APOLLO_KEY` is missing in `.env.local` but is present in `.env` then it will OVERWRITE the `.env` value with `undefined` and therefore you lose the API key! 

It should fallback to the initial `.env` value.

https://github.com/apollographql/apollo-tooling/issues/2161

PR #1537, opened 2019-09-19
[swift] Use multi-line string literals for queries

designatednerd:
In this PR there are a few changes to the Swift codegen: 

- Defaulted to printing queries as multi-line string literals to make sure the queries match the SHA256 hash being generated in `operationIdentifier`
- Added `--suppressSwiftMultilineStringLiterals` flag which defaults to false if you'd rather strip out all whitespace. 
- Removed comments showing the query since this is now covered by multi-line string literals

- [x] Update CHANGELOG.md with your change (include reference to issue & this PR)

### DIFF
--- a/packages/apollo-codegen-swift/src/__tests__/language.ts
+++ b/packages/apollo-codegen-swift/src/__tests__/language.ts
@@ -313,6 +313,33 @@ describe("Swift code generation: Escaping", () => {
       expect(SwiftSource.string("foo\n  bar  ", true).source).toBe('"foo bar"');
     });
 
+    it(`should generate multiline strings`, () => {
+      expect(SwiftSource.multilineString("foobar").source).toBe(
+        '"""\nfoobar\n"""'
+      );
+      expect(SwiftSource.multilineString("foo\n  bar  ").source).toBe(
+        '"""\nfoo\n  bar  \n"""'
+      );
+      expect(SwiftSource.multilineString(`"""foo"""`).source).toBe(
+        '#"""\n"""foo"""\n"""#'
+      );
+      expect(SwiftSource.multilineString("foo\\nbar").source).toBe(
+        '#"""\nfoo\\nbar\n"""#'
+      );
+      expect(SwiftSource.multilineString(`"""\\"""#"""`).source).toBe(
+        '##"""\n"""\\"""#"""\n"""##'
+      );
+      expect(SwiftSource.multilineString(`foo\\\\#bar`).source).toBe(
+        '##"""\nfoo\\\\#bar\n"""##'
+      );
+      expect(SwiftSource.multilineString(`foo\\\\#\\##bar`).source).toBe(
+        '###"""\nfoo\\\\#\\##bar\n"""###'
+      );
+      expect(SwiftSource.multilineString("foo\\###nbar").source).toBe(
+        '####"""\nfoo\\###nbar\n"""####'
+      );
+    });
+
     it(`should support concatenation`, () => {
       expect(swift`one`.concat().source).toBe("one");
       expect(swift`one`.concat(swift`two`).source).toBe("onetwo");
--- a/packages/apollo-codegen-swift/src/__tests__/language.ts
+++ b/packages/apollo-codegen-swift/src/__tests__/language.ts
@@ -338,14 +338,20 @@ describe("Swift code generation: Escaping", () => {
       generator = new SwiftGenerator({});
     });
 
-    it(`should trim with multilineString`, () => {
-      generator.multilineString("foo\n  bar  ");
+    it(`should not trim with multiline string if multiline strings are not suppressed and there is no triple quote`, () => {
+      generator.multilineString("foo\n  bar  ", false);
+
+      expect(generator.output).toBe('"""\nfoo\n  bar  \n"""');
+    });
+
+    it(`should trim with multilineString if multiline strings are suppressed`, () => {
+      generator.multilineString("foo\n  bar  ", true);
 
       expect(generator.output).toBe('"foo bar"');
     });
 
-    it(`shouldn't trim with multilineString when using """`, () => {
-      generator.multilineString('"""\nfoo\n  bar  \n"""');
+    it(`shouldn't trim with multilineString when using """ even when multiline strings are suppressed`, () => {
+      generator.multilineString('"""\nfoo\n  bar  \n"""', true);
       expect(generator.output).toBe('"\\"\\"\\"\\nfoo\\n  bar  \\n\\"\\"\\""');
     });
   });
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -99,13 +99,21 @@ export function generateSource(
           () => {
             Object.values(context.operations).forEach(operation => {
               if (operation.filePath === inputFilePath) {
-                generator.classDeclarationForOperation(operation, true, false);
+                generator.classDeclarationForOperation(
+                  operation,
+                  true,
+                  suppressMultilineStringLiterals
+                );
               }
             });
 
             Object.values(context.fragments).forEach(fragment => {
               if (fragment.filePath === inputFilePath) {
-                generator.structDeclarationForFragment(fragment, true, false);
+                generator.structDeclarationForFragment(
+                  fragment,
+                  true,
+                  suppressMultilineStringLiterals
+                );
               }
             });
           }
@@ -121,11 +129,19 @@ export function generateSource(
       });
 
       Object.values(context.operations).forEach(operation => {
-        generator.classDeclarationForOperation(operation, false, false);
+        generator.classDeclarationForOperation(
+          operation,
+          false,
+          suppressMultilineStringLiterals
+        );
       });
 
       Object.values(context.fragments).forEach(fragment => {
-        generator.structDeclarationForFragment(fragment, false, false);
+        generator.structDeclarationForFragment(
+          fragment,
+          false,
+          suppressMultilineStringLiterals
+        );
       });
     });
   }
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -62,6 +62,7 @@ export interface Options {
 export function generateSource(
   context: CompilerContext,
   outputIndividualFiles: boolean,
+  suppressMultilineStringLiterals: boolean,
   only?: string
 ): SwiftAPIGenerator {
   const generator = new SwiftAPIGenerator(context);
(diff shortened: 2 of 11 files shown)


designatednerd:
@lilyball Would love your opinion on these changes - I think a bunch of this is what you were looking for in the original implementation

packages/apollo/src/generate.ts
```diff
@@ -72,7 +72,19 @@ export default function generate(
     const outputIndividualFiles =
       fs.existsSync(outputPath) && fs.statSync(outputPath).isDirectory();
 
-    const generator = generateSwiftSource(context, outputIndividualFiles, only);
+    var suppressSwiftMultilineStringLiterals: boolean;
+    if (options.suppressSwiftMultilineStringLiterals == true) {
+      suppressSwiftMultilineStringLiterals = true;
+    } else {
+      suppressSwiftMultilineStringLiterals = false;
+    }
```
designatednerd: Typescript frenz: I was doing this late at night - is there a better way to "if false or undefined use this default value" than this? My googling skills on Typescript aren't great.

lilyball:
I think using a multiline literal is better than having the comment. Personally, I'd prefer to strip whitespace and have no comment because if I'm looking at the generated source it's to see the types, not the query, so the existence of `--suppressSwiftMultilineStringLiterals` makes me happy.

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -311,12 +311,33 @@ export class SwiftGenerator<Context> extends CodeGenerator<
...
+    if (string.includes('"""')) {
+      this.printOnNewline(SwiftSource.string(string, /* trim */ false));
+    } else {
+      if (suppressMultilineStringLiterals) {
+        this.printOnNewline(SwiftSource.string(string, /* trim */ true));
+      } else {
+        // Use a multiline string literal
+        this.printOnNewline(SwiftSource.raw`"""`);
+        string.split("\n").forEach(line => {
+          this.printOnNewline(SwiftSource.raw`${line}`);
```
lilyball: This is wrong; if the line contains a backslash, it won't be escaped. We need a modified form of `SwiftSource.string` that omits the surrounding quotes (maybe just called something like `SwiftSource.stringContents`). Doing that directly will also escape double quotes, which is unnecessary (as long as there aren't 3 in a row), so perhaps we should instead have a specific `SwiftSource.multilineString` that emits the triple quotes and does the modified escaping.

packages/apollo-codegen-swift/src/language.ts
```diff
@@ -311,12 +311,33 @@ export class SwiftGenerator<Context> extends CodeGenerator<
...
+   *                                          the same as Swift's string literal and will break the query.
+   */
+  multilineString(string: string, suppressMultilineStringLiterals: Boolean) {
     // Disable trimming if the string contains """ as this means we're probably printing an
     // operation definition where trimming is destructive.
-    this.printOnNewline(
-      SwiftSource.string(string, /* trim */ !string.includes('"""'))
-    );
+    if (string.includes('"""')) {
+      this.printOnNewline(SwiftSource.string(string, /* trim */ false));
```
lilyball: We don't have to do this. We can instead tweak escaping on the multiline string form such that any sequence of three quotes in a row has a backslash before the first quote. The following is perfectly legal and represents 9 quotes in a row:

```
"""
\"""\"""\"""
"""
```

packages/apollo/src/generate.ts
lilyball: Since the default value is indeed `false` you can just say `suppressSwiftMultilineStringLiterals = options.suppressSwiftMultilineStringLiterals || false;`. This will work for any optional type where the default value is the same as the falsy value of the type (if there is any at all), e.g. it would work for `string?` if the default value is `""`.

I think you can also say `!!options.suppressSwiftMultilineStringLiterals` but I'm not sure offhand if TypeScript likes that idiom.

packages/apollo-codegen-swift/src/__tests__/language.ts
```diff
@@ -338,14 +338,20 @@ describe("Swift code generation: Escaping", () => {
...
+      generator.multilineString("foo\n  bar  ", true);
 
       expect(generator.output).toBe('"foo bar"');
     });
 
-    it(`shouldn't trim with multilineString when using """`, () => {
-      generator.multilineString('"""\nfoo\n  bar  \n"""');
+    it(`shouldn't trim with multilineString when using """ even when multiline strings are suppressed`, () => {
+      generator.multilineString('"""\nfoo\n  bar  \n"""', true);
       expect(generator.output).toBe('"\\"\\"\\"\\nfoo\\n  bar  \\n\\"\\"\\""');
```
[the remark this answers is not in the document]
lilyball: BTW ignore the fact that the suggested preview is deleting a space. That seems to be a GitHub bug with suggestions on context lines.

packages/apollo-codegen-swift/src/language.ts
lilyball: Or instead of escaping we can do a raw string literal instead, we just have to calculate the appropriate string literal prefix (not that we should run into a conflict with Apollo sources, but it's better to be correct instead of assuming the string `"""#` will never show up in an Apollo query).

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -62,6 +62,7 @@ export interface Options {
 export function generateSource(
   context: CompilerContext,
   outputIndividualFiles: boolean,
+  suppressMultilineStringLiterals: boolean,
```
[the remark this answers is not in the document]
designatednerd: When I move all this crap to Swift 😁

packages/apollo/src/generate.ts
designatednerd: The default value is `false` when passed in from the command line but there's a couple places where this struct is created without any properties - that's why this has to be `suppressSwiftMultilineStringLiterals?: boolean` rather than `suppressSwiftMultilineStringLiterals: boolean`

Yes I agree this is silly

packages/apollo-codegen-swift/src/__tests__/language.ts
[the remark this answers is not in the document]
designatednerd: I would prefer to avoid this change here as I'm trying to validate that I didn't break anything. Maybe after this is merged we can take a look at updating these if you really want.

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -205,10 +223,10 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
       },
       () => {
         if (source) {
-          this.commentWithoutTrimming(source);
+          this.comment("The raw GraphQL definition of this operation");
```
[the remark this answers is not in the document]
designatednerd: It's not really a full sentence, so I'm not sure it needs a period the same way `This file was automatically generated and should not be edited.` does

packages/apollo-codegen-swift/src/language.ts
designatednerd: I took a look at using raw strings, but from what I saw raw strings and multiline strings do NOT play well together.

packages/apollo-codegen-swift/src/codeGeneration.ts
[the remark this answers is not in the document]
lilyball: Doc comment summaries in general should be a sentence with punctuation though. A "this was automatically generated" is just a comment, but this is a doc comment (speaking of, the method `comment` producing a doc comment is surprising).

packages/apollo-codegen-swift/src/language.ts
lilyball: Yeah they do.

```swift
#"""
This is a multiline string
It can include """
It just can't put a hash after it.
Escapes look like \#t.
"""#
```

(Ignore GitHub's syntax highlighting, that's not correct).

My only worry about using raw strings is the logic to calculate the appropriate `#` count. I think it's simpler just to use a regular multiline string and add the appropriate escaping as needed.

packages/apollo-codegen-swift/src/language.ts
designatednerd: This sounds like a great idea for a follow-up PR.

packages/apollo-codegen-swift/src/language.ts
designatednerd: I'll take a look at a test around a line with a backslash

designatednerd: Aha, i was fucking something up with the raw strings. I think I'd probably just use `###` as the delimiter by default so we don't have to count `#`s. Plus raw strings mean we don't have to screw around with slashies too

lilyball: I'm not actually sure it's legal to get `"""#` or `\#` in a GraphQL query, so it's possible we don't have to worry. I'm currently trying to figure out if it's possible to construct something that parses like this.

I am assuming comments get stripped, otherwise anything goes in a comment.

lilyball: Ok no, it's possible to get the token `\"""#` to show up in a GraphQL multiline string.

```graphql
query Bar {
  foo(str: """wat\"""#""") {
    id
  }
}
```

I can also apparently pass `\#` through the GraphQL parse step, though that's not a known escape so I assume it should fail at runtime.

lilyball: The best thing to do here is probably to just add basic string escaping here and then I'll try and do a followup PR that cleans it up such that it does the "optimal" formatting.

lilyball:
I filed #1538 as a PR against your branch to do the fixes to multiline string literals that I talked about.

packages/apollo/src/generate.ts
JakeDawkins: @designatednerd I believe you could use the `Boolean` constructor here too. Which would play a little nicer than the `!!` with ts and would prevent you from having to define the type explicitly.

```
    const suppressSwiftMultilineStringLiterals = Boolean(options.suppressSwiftMultilineStringLiterals)
```

this would probably also work if you need to check if its explicitly `true` and not just something truthy like a string.

```
const suppressSwiftMultilineStringLiterals = options.suppressSwiftMultilineStringLiterals === true;
```

unless I'm missing something here.

JakeDawkins: Also, any reason you're using `var` instead of `const` or `let` here?

designatednerd: I think because I thought I had to because I wasn't assigning immediately

lilyball: FWIW the [MDN docs on the `Boolean` object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/boolean) explicitly describes `!!` as equivalent to using `Boolean` as a function.

PR #1539, opened 2019-09-21
Modify service:check command to use combined mutation

zionts:
Several months ago, we implemented a mutation called
`checkPartialSchema`, which was designed to encapsulate both checking
for ability to compose and validating schema changes against operations.
  This new mutation has several added benefits, like displaying a new
  webpage with better information for federation users, along with
  allowing users to see multiple status lines appear in their PRs when
  using the GitHub integration. This PR attempts to migrate to use the
  newer mutation, abandoning the old. There were some sad moments trying
  to fight against the automatic type generation, so please forgive me
  if anything is not quite right type-wise, and please take over this
    commit at will. I have yet to test it locally, but I figured that
    this could at the very least get the ball rolling, and I'm happy to
    push this to completion if there isn't enough capacity on the
    (single person) tooling team at the moment, though I think this
    change should be already mostly complete and relatively small.

TODO:

### DIFF
--- a/packages/apollo/src/commands/service/__tests__/__snapshots__/check.test.ts.snap
+++ b/packages/apollo/src/commands/service/__tests__/__snapshots__/check.test.ts.snap
@@ -186,7 +186,6 @@ exports[`service:check integration federated should report composition success c
 exports[`service:check integration federated should report composition success correctly compacts output in CI 1`] = `
 "Loading Apollo Project
 Found 0 graph composition errors for service accounts on graph engine
-Validated composed schema against tag master on graph engine
 Compared 1 schema change against 0 operations over the last 548 days
 Found 0 breaking changes and 1 compatible change
 ╔════════╤══════════════════╤═══════════════════════════════════════════════════════════════════════════════╗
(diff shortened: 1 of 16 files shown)


packages/apollo/src/commands/service/check.ts
```diff
@@ -429,6 +422,19 @@ export default class ServiceCheck extends ProjectCommand {
...
+                } else {
+                  if (!checkSchemaResult) {
+                    throw new Error(
+                      "Violated invariant. Schema should have been validated against operations if" +
+                        "there were no composition errors"
+                    );
+                  }
+
+                  // this is used for the printing
+                  taskOutput.checkSchemaResult = checkSchemaResult;
```
JakeDawkins: @zionts the only thing missing here was that the `checkSchemaResult` wasn't being attached to the task output, so the `formatHumanReadable` function below didn't have access to the data it expected. This should fix that :)

JakeDawkins:
@zionts these changes look good to me, and I think this cleans up our schema checks pretty well! So thank you!

I think we just need to fix these tests, and we'd be good to go. Do you have the time to do that?

packages/apollo/src/commands/service/__tests__/__snapshots__/check.test.ts.snap
```diff
@@ -148,7 +148,6 @@ exports[`service:check integration federated should report composition success c
 exports[`service:check integration federated should report composition success correctly compacts output in CI 1`] = `
 "Loading Apollo Project
 Found 0 graph composition errors for service accounts on graph engine
-Validated composed schema against tag master on graph engine
 Compared 1 schema change against 0 operations over the last 548 days
```
JakeDawkins: This change is because on federated services, we no longer need to run that step

PR #1555, opened 2019-10-01
Add missing `graphql` peer dependency

briancoit:
`graphql` is currently imported, but not mentioned as a package dependency. When installing via pnpm, `graphql` cannot be resolved without this addition. Even with npm/yarn, the missing peer dependency may prove to be problematic for consumers (lack of version constraint).

TODO:

### DIFF
--- a/packages/apollo-tools/package.json
+++ b/packages/apollo-tools/package.json
@@ -18,6 +18,9 @@
   "dependencies": {
     "apollo-env": "file:../apollo-env"
   },
+  "peerDependencies": {
+    "graphql": "^14.2.1"
+  },
   "jest": {
     "preset": "ts-jest",
     "testEnvironment": "node",


zionts:
@briancoit , thanks for submitting this! I think declaring a peer dependency on `graphql` sounds perfectly reasonable, and I'm curious what research led you to pick the version number that you went with. @JakeDawkins is definitely the right person to speak to what the peer dependency should be at, so I'm not equipped to approve this PR, but any research you did to reach that version number would be helpful :)

zionts:
hmm, actually... since this package is managed with `lerna`, I think that the dependencies are actually specified directly under each package (e.g. apollo-language-server's [package.json](https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-language-server/package.json#L37)). If there's a particular problem that you're running into with one of the packages, please let us know so we can resolve it, but I'm going to close this PR out for now, since it's modifying the root dependency rather than one of the lerna-managed packages.

PR #1589, opened 2019-10-17
swift: Fix codegen with empty input objects / arrays

lilyball:
The codegen was omitting the `[:]` and `[]` tokens entirely.

Fixes apollographql/apollo-ios#838

The root cause of missing `[:]` was there was some code relying on the truthy value of strings, but the conversion to `SwiftSource` caused even empty sources to become truthy. I'm actually surprised TypeScript happily accepts `someObject || someStr` and evaluates it as the object type with no warnings (I mean, it will always be the object type, but the `|| someStr` is now dead code).

Empty arrays I'm pretty sure were always broken even before `SwiftSource` due to the counterintuitive behavior of `wrap` (namely, `wrap("[", "", "]") == ""` instead of returning `"[]"`). I didn't want to deviate too much from the underlying `wrap` behavior (since the behavior is relied-upon in places) so I just made it return `undefined` instead of an empty source object, which forced us to handle that at the call sites.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - `apollo-codegen-scala`
   - <First `apollo-codegen-scala` related entry goes here>
 - `apollo-codegen-swift`
-  - <First `apollo-codegen-swift` related entry goes here>
+  - Fix code generation for empty input objects / arrays [#1589](https://github.com/apollographql/apollo-tooling/pull/1589)
 - `apollo-codegen-typescript`
   - <First `apollo-codegen-typescript` related entry goes here>
 - `apollo-env`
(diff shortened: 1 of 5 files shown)


lilyball:
Failure is legitimate. Why didn't I see any errors locally?

lilyball:
Rather, I see it now in VSCode when I scroll to the right line, but I didn't see any when running `npm run clean && npm install` in the terminal.

designatednerd:
@abernix Any idea what's going on with the failure on Node 8? Seems unrelated. 

@lilyball Thanks for this - @cltnschlosser When you have a chance can you confirm this fixes your issue please? I believe you should be able to pull the updated CLI zip from the artifacts tab of the `Build CLI` circle job (grab the one that says Darwin, that includes node).

designatednerd:
Great - kicked the build again, @abernix wasn't really sure what the hell was going on

PR #1595, opened 2019-10-17
Swift: Add flag to omit deprecated enum cases

kevinmbeaulieu:
Implement feature request described in https://github.com/apollographql/apollo-ios/issues/840

TODO:

### DIFF
--- a/__fixtures__/starwars/schema.graphql
+++ b/__fixtures__/starwars/schema.graphql
@@ -44,6 +44,8 @@ enum LengthUnit {
   METER
   # Primarily used in the United States
   FOOT
+  # Ancient unit used during the Middle Ages
+  CUBIT @deprecated(reason: "Test deprecated enum case")
 }
 # A humanoid creature from the Star Wars universe
 type Human implements Character {
--- a/packages/apollo/src/commands/client/codegen.ts
+++ b/packages/apollo/src/commands/client/codegen.ts
@@ -64,6 +64,9 @@ export default class Generate extends ClientCommand {
     namespace: flags.string({
       description: "The namespace to emit generated code into."
     }),
+    omitDeprecatedEnumCases: flags.boolean({
+      description: "Omit deprecated enum cases from generated code [Swift only]"
+    }),
     operationIdsPath: flags.string({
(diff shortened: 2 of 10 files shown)


kevinmbeaulieu:
@designatednerd Added tests for both the enabled & disabled states of the new flag

__fixtures__/starwars/schema.graphql
```diff
@@ -44,6 +44,8 @@ enum LengthUnit {
   METER
   # Primarily used in the United States
   FOOT
+  # Ancient unit used during the Middle Ages
+  CUBIT @deprecated(reason: "Test deprecated enum case")
```
designatednerd: I'm OK with this given that tests passed but I want to make sure our JS team is OK with these fixtures being updated since this could theoretically affect other things besides swift

kevinmbeaulieu: Makes sense, just let me know (or if there's a better way to test this without making a change to a schema used by other languages, happy to update)

designatednerd: (update: They seem cool with it)

packages/apollo/src/commands/client/codegen.ts
```diff
@@ -64,6 +64,9 @@ export default class Generate extends ClientCommand {
     namespace: flags.string({
       description: "The namespace to emit generated code into."
     }),
+    omitDeprecatedEnumCases: flags.boolean({
+      description: "Omit deprecated enum cases from generated code [Swift only]"
```
designatednerd: Should this default to `false` to preserve existing behavior, or would that be overkill?

kevinmbeaulieu: Yeah I intended for this to default to false to preserve existing behavior- is that not how it would behave as currently implemented? Looking at the implementation of `flags.boolean` I saw "specifying a default of false is the same not specifying a default"

designatednerd: Ah ok (not exactly a typescript expert over here). I think that should work fine, then.

PR #1596, opened 2019-10-17
swift: Fix code generation for null input fields

### LINKED ISSUE #1592: Swift: Can't specify null in query
**Intended outcome:**
Works correctly.

**Actual outcome:**
Codegen error:
```
→ Cannot read property 'kind' of null
TypeError: Cannot read property 'kind' of null
    at expressionFromValue (path/Pods/Apollo/scripts/apollo/node_modules/apollo-codegen-swift/lib/helpers.js:159:23)
    at language_1.SwiftSource.wrap.language_1.SwiftSource.join.Object.entries.map (path/Pods/Apollo/scripts/apollo/node_modules/apollo-codegen-swift/lib/helpers.js:167:87)
    at Array.map (<anonymous>)
    at expressionFromValue (path/Pods/Apollo/scripts/apollo/node_modules/apollo-codegen-swift/lib/helpers.js:166:124)
    at language_1.SwiftSource.wrap.language_1.SwiftSource.join.args.map.arg (path/Pods/Apollo/scripts/apollo/node_modules/apollo-codegen-swift/lib/helpers.js:178:84)
    at Array.map (<anonymous>)
    at Helpers.dictionaryLiteralForFieldArguments (path/Pods/Apollo/scripts/apollo/node_modules/apollo-codegen-swift/lib/helpers.js:177:99)
    at withIndent (path/Pods/Apollo/scripts/apollo/node_modules/apollo-codegen-swift/lib/codeGeneration.js:499:79)
    at GeneratedFile.withIndent (path/Pods/Apollo/scripts/apollo/node_modules/apollo-codegen-core/lib/utilities/CodeGenerator.js:54:9)
    at SwiftAPIGenerator.withIndent (path/Pods/Apollo/scripts/apollo/node_modules/apollo-codegen-core/lib/utilities/CodeGenerator.js:109:26)
```

This should help in making a testcase.
```
const schema = buildSchema(`
         schema {
           query: Query
         }
         type Query {
           foo(input: FooInput!): Int
         }
         input FooInput {
           id: ID
         }
       `);
       const document = parse(`
         query FieldArgumentsWithEmptyInputObject {
           foo(input: { id: null }) {
             id
           }
         }
       `);
```

lilyball:
Also add a test that covers all scalar types just in case.

Fixes #1592

TODO:

### DIFF
--- a/packages/apollo-codegen-swift/src/helpers.ts
+++ b/packages/apollo-codegen-swift/src/helpers.ts
@@ -255,7 +255,9 @@ export class Helpers {
 
   dictionaryLiteralForFieldArguments(args: Argument[]): SwiftSource {
     function expressionFromValue(value: any): SwiftSource {
-      if (value.kind === "Variable") {
+      if (value === null) {
+        return swift`nil`;
+      } else if (value.kind === "Variable") {
         return swift`GraphQLVariable(${SwiftSource.string(
           value.variableName
         )})`;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - `apollo-codegen-swift`
   - Fix code generation for empty input objects / arrays [#1589](https://github.com/apollographql/apollo-tooling/pull/1589)
   - Add flag to omit deprecated enum cases [#1595](https://github.com/apollographql/apollo-tooling/pull/1595)
+  - Fix code generation for input fields with the value `null` [#1596](https://github.com/apollographql/apollo-tooling/pull/1596)
 - `apollo-codegen-typescript`
   - <First `apollo-codegen-typescript` related entry goes here>
 - `apollo-env`
--- a/packages/apollo-codegen-swift/src/__tests__/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/__tests__/codeGeneration.ts
@@ -870,5 +870,45 @@ describe("Swift code generation", () => {
 
       expect(dictionaryLiteral).toBe('["input": []]');
     });
+
+    it("should handle input fields of various scalar types including null", () => {
+      // As with the previous test, we need to build our own schema.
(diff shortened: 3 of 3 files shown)


packages/apollo-codegen-swift/src/helpers.ts
```diff
@@ -255,7 +255,9 @@ export class Helpers {
 
   dictionaryLiteralForFieldArguments(args: Argument[]): SwiftSource {
     function expressionFromValue(value: any): SwiftSource {
-      if (value.kind === "Variable") {
+      if (value === null) {
+        return swift`nil`;
```
lilyball: I'm assuming here that we want a literal `nil` in the result map instead of e.g. `NSNull()`, since the `ResultMap`'s type is `[String: Any?]`. That said, I haven't tested this change in the context of a full project, just the unit test as given.

@cltnschlosser I would appreciate it if you could test this with your full project.

cltnschlosser: Yeah, I think `nil` makes sense. I didn’t actually use this in a project. I had discovered it when trying to workaround the other issue. I think we just need to verify that the resulting JSON has the null value. I can do that tomorrow.

lilyball:
Oops I didn't update the changelog after creating the PR. I'll do that tomorrow.

packages/apollo-codegen-swift/src/helpers.ts
cltnschlosser:
```
{
	"operationName": "Test",
	"query": "query Test {\n  employee(request: {id: null}) {\n    __typename\n    fullName\n  }\n}",
	"variables": null
}
```

Seems to work, and graphql-faker was fine with the request.

lilyball:
Updated changelog (and rebased for good measure).

PR #1597, opened 2019-10-17
Add variant to service list output

zionts:
We always fetch list of services with a variant, so include it in the
CLI output

TODO:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,25 @@
         "lodash.xorby": "^4.7.0"
       },
       "dependencies": {
+        "apollo-env": {
+          "version": "0.5.1",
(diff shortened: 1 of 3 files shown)


JakeDawkins:
I think the only concerning thing is that the cyan is less visible on a white bg, than the blue. But I think the cyan-white combo is _more_ visible than the blue-black combo, and since the majority of people use dark terminals, I think this is an okay tradeoff.

Just a couple screenshots to show the differences here :) 

**Before/after with black bg**

Screen Shot 2019-10-18 at 10 35 22 AM
Screen Shot 2019-10-18 at 10 35 55 AM

**Before/after with a white bg**

Screen Shot 2019-10-18 at 10 46 50 AM
Screen Shot 2019-10-18 at 10 35 33 AM

PR #1598, opened 2019-10-18
Modify styling of CLI output

zionts:
Specifically, replace all instances of chalk.blue with chalk.cyan, which
is more readable in dark mode (the green color we use is basically
unreadable in light mode anyway)

Additionally, this changes some spacing of the service:list command to
allow users to see the URL to navigate to more obviously.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - `apollo`
   - Add sdl download ability to `client:download-schema`
+  - colors: use cyan instead of blue for text highlighting
 - `apollo-codegen-flow`
   - <First `apollo-codegen-flow` related entry goes here>
 - `apollo-codegen-scala`
--- a/packages/apollo/src/commands/client/check.ts
+++ b/packages/apollo/src/commands/client/check.ts
@@ -149,7 +149,7 @@ export default class ClientCheck extends ClientCommand {
   }) => {
     const { name, locationOffset, relativePath } = operation;
     this.log(
-      `${name}: ${chalk.blue(`${relativePath}:${locationOffset.line}`)}\n`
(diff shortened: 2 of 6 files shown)


JakeDawkins:
I think the only concerning thing is that the cyan is less visible on a white bg, than the blue. But I think the cyan-white combo is _more_ visible than the blue-black combo, and since the majority of people use dark terminals, I think this is an okay tradeoff.

Just a couple screenshots to show the differences here :) 

**Before/after with black bg**

Screen Shot 2019-10-18 at 10 35 22 AM
Screen Shot 2019-10-18 at 10 35 55 AM

**Before/after with a white bg**

Screen Shot 2019-10-18 at 10 46 50 AM
Screen Shot 2019-10-18 at 10 35 33 AM

zionts:
thanks for attaching the screenshots @JakeDawkins ! Maybe @daniman wants to mess around with colors too 👀

JakeDawkins:
@zionts @daniman maybe a solution would be to use background colors in tandem with foregrounds? that way, we know the background color and the foreground color so we don't run into any theme issues with terminals? Just a thought, it could look clunky

zionts:
@JakeDawkins I found that background colors looked super clunky, unfortunately :/ Ideally, there'd be a library that takes into account the users' theme, but in lieu of that, I think this is an improvement!

JakeDawkins:
@zionts gotcha, gotcha. I think another alternative is to try to set colors as best as we can, but allow a `--noColor` or `-n` flag to disable colors and instead set the text to bold for users whose themes are still unreadable. 

I think since people can adjust their color palettes all over the place, we may not be able to _fully_ ensure readability.

PR #1606, opened 2019-10-20
Swift: Add CLI argument to set indent width

kevinmbeaulieu:
Enable the ability to set a custom indentation width
(e.g., 4 spaces instead of the default 2) on generated code, so
users can configure the generated code to match the rest of their
project's indentation.

TODO:

### DIFF
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -41,6 +41,7 @@ import { generateOperationId } from "apollo-codegen-core/lib/compiler/visitors/g
 import { collectAndMergeFields } from "apollo-codegen-core/lib/compiler/visitors/collectAndMergeFields";
 
 import "apollo-codegen-core/lib/utilities/array";
+import { GeneratedFile } from "apollo-codegen-core/lib/utilities/CodeGenerator";
 
 const { join, wrap } = SwiftSource;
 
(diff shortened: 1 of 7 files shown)


packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -156,6 +157,29 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
...
+
+      let indentWidth = this.context.options.indentWidth;
+      if (indentWidth) {
+        file.indentWidth = indentWidth;
+      }
+
+      this.generatedFiles[fileName] = file;
+    }
+    const oldCurrentFile = this.currentFile;
+    this.currentFile = file;
```
designatednerd: why does this need to be grabbed and reset?

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -156,6 +157,29 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
     super(context);
 
     this.helpers = new Helpers(context.options);
+
+    let indentWidth = context.options.indentWidth;
```
designatednerd: This is the only thing that jumps out to me as a possible source of where some of these failures are coming from: Since it's `context.options` instead of `this.context.options`, something may be getting adjusted in the `super` call that is hammering everything else.

JakeDawkins:
> The main reason I ask is because it will make things somewhat more annoying to maintain over the long term (which is almost certainly why this isn't already a feature), and I want to try to see the other side of the argument.

I think this is my biggest concern here too. I'd like to keep from adding an abundance of flags and features that only work on one language. It just makes setting up and configuring codegen more of a hassle imho

designatednerd:
@JakeDawkins What would your feeling be if we adjusted the code to make it work for all supported languages rather than just Swift? 

Again, @kevinmbeaulieu if you could share the benefits here I'd appreciate it.

kevinmbeaulieu:
@designatednerd @JakeDawkins Thanks for the feedback, I hadn't thought about how it might make long-term maintenance more difficult. The only benefit was that when I'm find myself reading generated code (e.g., to find the name of some field/type), I find it easier to read with 4 spaces. But that small benefit isn't worth making things more complicated to maintain in the future, so will close this for now

designatednerd:
Cool, thanks a lot for considering!

PR #1610, opened 2019-10-21
Remove Azure Pipelines CI work

JakeDawkins:
This PR intends to do the following:

- Remove all Azure checks
  - This will unify our CI processes and shared work for releases and generally make life easier when running manual rebuilds, debugging, etc.
- Move extension building/testing step to circleci
- Automatically publish the extension after release has been confirmed

TODO:


packages/vscode-apollo/package.json
```diff
@@ -20,6 +20,7 @@
     "update-vscode": "node ../../node_modules/vscode/bin/install",
     "create-server-symlink": "npx rimraf node_modules && mkdir node_modules && node create-server-symlink.js",
     "package-extension": "npx rimraf node_modules && yarn install --force --production --ignore-scripts && ../../node_modules/vsce/out/vsce package --yarn --baseContentUrl https://raw.githubusercontent.com/apollographql/apollo-tooling/master/packages/vscode-apollo && git clean -dfqX -- ./node_modules ./yarn.lock && npm run create-server-symlink",
+    "publish-extension": "npx rimraf node_modules && yarn install --force --production --ignore-scripts && ../../node_modules/vsce/out/vsce publish -p $VS_MARKETPLACE_TOKEN --yarn --baseContentUrl https://raw.githubusercontent.com/apollographql/apollo-tooling/master/packages/vscode-apollo",
```
JakeDawkins: publish extension _can_ just publish an already built vsix file, but since we don't know the name of it already (and aren't currently renaming it), I thought it'd probably just be easiest to build it again in this step. I'm open to renaming it something predictable and just publishing the already-built artifact from a previous step though. I'd need to figure out how to access artifacts from previous steps too :)

package.json
```diff
@@ -32,6 +32,8 @@
     "release:version-bump": "lerna version --exact",
     "release:start-ci-publish": "node -p '`Publish (dist-tag:${process.env.APOLLO_DIST_TAG || \"latest\"})`' | git tag -F - \"publish/$(date -u '+%Y%m%d%H%M%S')\" && git push origin \"$(git describe --match='publish/*' --tags --exact-match HEAD)\"",
     "package-extension": "npm run clean && npm ci && lerna run package-extension --stream --scope=vscode-apollo",
+    "package-extension:no-clean": "lerna run package-extension --stream --scope=vscode-apollo",
```
JakeDawkins: On CI, this is a fresh install, so we don't need to clean and reinstall. I want to leave it for local builds though just as a safety measure

packages/vscode-apollo/package.json
abernix: Also, is the cleanup bit that currently lives on `package-extension` important for `publish-extension`? (Assuming it's ever ran locally?)  I guess I'm worried that if `yarn.lock` gets left (which for some inexplicable reason is cleansed, rather than using `yarn --no-lockfile`, 🤷‍♂), something weird might happen for the next step.

e.g., this bit:

```
 && git clean -dfqX -- ./node_modules ./yarn.lock && npm run create-server-symlink
```

.circleci/config.yml
```diff
@@ -44,7 +46,39 @@ jobs:
           paths:
             - ./**
 
-  VSCode:
+  NodeJS 12 Windows:
+    executor:
+      name: win/default
```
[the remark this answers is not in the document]
JakeDawkins: I'm not sure what you mean. I define version 2.0.0 of the orb at the top of the file. Default is just the name of the executor

packages/apollo-language-server/src/config/__tests__/loadConfig.ts
```diff
@@ -21,8 +21,8 @@ const makeNestedDir = dir => {
 };
 
 const deleteFolderRecursive = path => {
-  // don't delete files on azure CI
-  if (process.env.AZURE_HTTP_USER_AGENT) return;
+  // don't delete files on circlci windows -- will get a resource locked error
+  if (process.env.OS && process.env.OS.includes("Windows")) return;
```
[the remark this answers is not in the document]
JakeDawkins: Unfortunately, it's a windows thing 😞 I haven't had the time to figure out what it is. So technically we don't support _developing_ the tooling project on windows very well right now. I updated it to check for windows, not just the specific env variable circleci uses

packages/vscode-apollo/package.json
JakeDawkins: I extracted out those two parts and added an `_internal:prepare-extension` and `_internal:cleanup-extension` just to be safe :) I can revisit all this when I work on bundling the extension in the mid-to-near future

PR #1611, opened 2019-10-22
Make `apollo-env` a standard TS package

trevor-scheer:
The purpose of this PR is to simplify the build step of the tooling monorepo by implementing `apollo-env` as another project reference just as the others are.

`apollo-env` has been a notoriously bad citizen of this monorepo, especially for developers who aren't familiar with the quirks of this particular setup. It's removed hours of productivity even for those who _are_ well acquainted with said quirks. For example, the need for frequent `npm run clean && npm ci` is really a _requirement_ for development, but is an unexpected quirk and serious time sink - and this really just marks the simplest case.

This cleans up all things `apollo-env` to the best of my ability:
* Special cases of inclusion or exclusion in configs (Jest, TS)
* Removal of `apollo-env` as a project dependency when it is unused
* Special build scripts which previously compiled `apollo-env` before everything else
* Removal of handwritten `fetch` types

The last bullet is a major question mark to me - as my understanding of this whole workaround is for the sake of these handwritten types. I don't see the value that they add, but I'll defer to those with more knowledge of the history here @martijnwalraven and possibly @abernix.

TODO:


packages/apollo-env/src/fetch/fetch.ts
```diff
@@ -0,0 +1,100 @@
+import { Agent as HttpAgent } from "http";
+import { Agent as HttpsAgent } from "https";
+
+// declare module "node-fetch" {
```
trevor-scheer: Everything commented out here is intended to be deleted. Left as a TODO for now.

packages/apollo-env/src/fetch/fetch.ts
```diff
@@ -0,0 +1,100 @@
...
+//   export function fetch(
+//     input?: RequestInfo,
+//     init?: RequestInit
+//   ): Promise<Response>;
+// }
+
+export type RequestAgent = HttpAgent | HttpsAgent;
```
abernix: I'm not sure we use this anywhere, but I think it's okay to leave it for now.

packages/apollo-env/src/fetch/fetch.ts
```diff
@@ -0,0 +1,100 @@
...
+//   | "default"
+//   | "no-store"
+//   | "reload"
+//   | "no-cache"
+//   | "force-cache"
+//   | "only-if-cached";
+
+// export type RequestRedirect = "follow" | "error" | "manual";
+
+export type ReferrerPolicy =
```
abernix: I also don't think we use this anywhere, but let's leave it for now.

packages/apollo-env/src/fetch/fetch.ts
```diff
@@ -0,0 +1,100 @@
...
+//   headers?: HeadersInit;
+//   status?: number;
+//   statusText?: string;
+//   // Although this isn't part of the spec, `node-fetch` accepts a `url` property
+//   url?: string;
+// }
+
+// export type BodyInit = ArrayBuffer | ArrayBufferView | string;
+
+export {
```
abernix: We could alternatively just import these directly from `node-fetch`, but this is a great improvement right now.

JakeDawkins:
I'm not sure what is going on with the azure pipelines fail here... all the jobs pass. Feel free to ignore that

PR #1612, opened 2019-10-22
Unify and simplify jest config

trevor-scheer:
Semi-dependent upon #1611, though this could be split out to be its own set of changes independent of said PR.

This PR introduces a base Jest config from which all projects extend. It also simplifies the current config, as many options that were specified within each package were actually just the defaults that Jest provides.

This approach is a boon for maintainability of the configuration of this repo, and aligns with the approach that [`apollo-server`](https://github.com/apollographql/apollo-server/blob/master/jest.config.base.js) has taken.

TODO:

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  projects: ["<rootDir>/packages/!(vscode-apollo)"]
+};
--- a/packages/apollo-tools/jest.config.js
+++ b/packages/apollo-tools/jest.config.js
@@ -2,9 +2,5 @@ const baseConfig = require("../../jest.base");
 
 module.exports = {
   ...baseConfig,
-  setupFiles: ["apollo-env"],
-  testPathIgnorePatterns: [
-    ...baseConfig.testPathIgnorePatterns,
-    "snapshotSerializers"
-  ]
+  setupFiles: ["apollo-env"]
 };
--- a/packages/apollo-tools/jest.config.js
+++ b/packages/apollo-tools/jest.config.js
@@ -1,0 +1,10 @@
+const baseConfig = require("../../jest.base");
+
+module.exports = {
+  ...baseConfig,
+  setupFiles: ["apollo-env"],
+  testPathIgnorePatterns: [
+    ...baseConfig.testPathIgnorePatterns,
+    "snapshotSerializers"
+  ]
+};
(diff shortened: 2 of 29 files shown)


jest.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
JakeDawkins: Why do this here instead of `package.json`?

packages/apollo-tools/jest.config.js
```diff
@@ -0,0 +1,9 @@
+const baseConfig = require("../../jest.base");
+
+module.exports = {
+  ...baseConfig,
+  testPathIgnorePatterns: [
+    ...baseConfig.testPathIgnorePatterns,
+    "snapshotSerializers"
```
JakeDawkins: This wasn't in the original config. since they're under `__tests__` wouldn't they be ignored anyway?

packages/apollo-tools/src/__tests__/snapshotSerializers/astSerializer.ts
```diff
@@ -12,7 +12,7 @@ export = {
     indentation: string,
     depth: number,
     refs: Refs,
-    printer: Printer
+    printer: any
```
JakeDawkins: Why not use the `Printer` type? was it removed from the exports?

packages/apollo-tools/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.base",
   "compilerOptions": {
     "rootDir": "./src",
-    "outDir": "./lib"
+    "outDir": "./lib",
+    "esModuleInterop": true
```
JakeDawkins: Any specific reason for this?

jest.config.js
trevor-scheer: I think this is just a matter of consistency - with all the other configs moving out of `package.json`, I think it would be unexpected for this config to continue living there.

packages/apollo-tools/jest.config.js
trevor-scheer: I think `__tests__` files _won't_ be ignored, however I think you're correct that this is unnecessary since they don't end in `.test.*` or `.spec.*`! Removing this worked for me locally, so trying in CI.

packages/apollo-tools/src/__tests__/snapshotSerializers/astSerializer.ts
trevor-scheer: I only vaguely recall why I did this, and it has to do with type incompatibility from `pretty-format`'s types =, however it works perfectly fine at runtime. Decided to remove these unused params anyway!

packages/apollo-tools/tsconfig.json
trevor-scheer: Nope, must be an artifact of some exploration that I backtracked on. Removed!

packages/apollo-tools/jest.config.js
JakeDawkins: ahh, I thought that was ignored in the base config 🙈

PR #1614, opened 2019-10-23
Support User Tokens

caydie-tran:
This PR supports setting `ENGINE_API_KEY` to the user's token! The reason why we need to support user tokens is so that we can eventually move to a world where the CLI doesn't need to depend on service tokens and so that we can instantiate the environment variables by running `apollo login` rather than manually copying and pasting keys into configuration files. 

TODO:

### DIFF
--- a/packages/apollo-language-server/src/config/__tests__/config.ts
+++ b/packages/apollo-language-server/src/config/__tests__/config.ts
@@ -58,18 +58,18 @@ describe("ApolloConfig", () => {
   describe("tag", () => {
     it("gets default tag when none is set", () => {
       const config = new ApolloConfig({ client: { service: "hai" } });
-      expect(config.tag).toEqual("current");
+      expect(config.serviceGraphVariant).toEqual("current");
     });
 
     it("gets tag from service specifier", () => {
       const config = new ApolloConfig({ client: { service: "hai@master" } });
-      expect(config.tag).toEqual("master");
+      expect(config.serviceGraphVariant).toEqual("master");
     });
 
     it("can set and override tags", () => {
       const config = new ApolloConfig({ client: { service: "hai@master" } });
-      config.tag = "new";
-      expect(config.tag).toEqual("new");
+      config.serviceGraphVariant = "new";
+      expect(config.serviceGraphVariant).toEqual("new");
     });
   });
 
--- a/packages/apollo-language-server/src/config/__tests__/config.ts
+++ b/packages/apollo-language-server/src/config/__tests__/config.ts
@@ -7,7 +7,7 @@ import {
 import URI from "vscode-uri";
 import { DeepPartial } from "apollo-env";
 
-function createConfig(raw: DeepPartial<ApolloConfigFormat>) {
+export function createConfig(raw: DeepPartial<ApolloConfigFormat>) {
   return loadConfigWithDefaults(
     {
       config: raw
--- a/packages/apollo-language-server/src/config/__tests__/config.ts
+++ b/packages/apollo-language-server/src/config/__tests__/config.ts
@@ -1,11 +1,29 @@
-import { ApolloConfig, ApolloConfigFormat } from "../";
+import {
+  ApolloConfig,
+  ApolloConfigFormat,
+  getGraphInfo,
+  loadConfigWithDefaults
+} from "../";
 import URI from "vscode-uri";
+import { DeepPartial } from "apollo-env";
+
+function createConfig(raw: DeepPartial<ApolloConfigFormat>) {
+  return loadConfigWithDefaults(
+    {
+      config: raw
+    },
+    getGraphInfo(raw)
+  );
+}
 
 describe("ApolloConfig", () => {
-  describe("confifDirURI", () => {
+  describe("configDirURI", () => {
     it("properly parses dir paths for configDirURI", () => {
       const uri = URI.parse("/test/dir/name");
-      const config = new ApolloConfig({ service: { name: "hai" } }, uri);
+      const config = new ApolloConfig(
+        createConfig({ service: { name: "hai" } }),
+        uri
+      );
       // can be either /test/dir/name or \\test\\dir\\name depending on platform
       // this difference is fine :)
       expect(config.configDirURI.fsPath).toMatch(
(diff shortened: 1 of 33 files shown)


packages/apollo/src/commands/service/check.ts
```diff
@@ -283,22 +283,18 @@ export default class ServiceCheck extends ProjectCommand {
     const { isCi } = envCi();
 
     let schema: GraphQLSchema | undefined;
+    let graphID: string | undefined;
+    let graphVariant: string | undefined;
     try {
       await this.runTasks<TasksOutput>(
         ({ config, flags, project }) => {
-          if (!isServiceProject(project)) {
```
caydie-tran: this should still be here

caydie-tran: actually this needs to be removed otherwise when we use a user token as the engine_api_key this will throw

zionts:
All right. For the past couple of days, I have been fooling around on this branch to refactor this repo to better support the concept of `service` and `client` project definitions in the config (in a backwards compatible way) as well as empty configs that contain defaults. I'm not sure what the next step is, but I'm going to take a step back from these changes for the new few days to focus on Summit-related activity. The current state of this branch is that all tests are passing and that all commands (client / service) can be run with a config that defines both `client` and `service` keys in the config. I have not done any manual QA with VS Code, though I _believe_ the changes should be backwards compatible — it's mostly just refactoring. This also supports user tokens. The commits are not very easy to review as they are since I spent a lot of time trying to understand all of the mechanics around what was going on within this codebase. My hope is that this refactor will make things easier to understand by separating out some of the setup steps that are shared between the VS Code workspace the CLI, and that by supporting defaults & multi-configs, we can have both a better starting experience and a better story around advanced support.

packages/apollo/src/commands/service/check.ts
```diff
@@ -351,9 +347,9 @@ export default class ServiceCheck extends ProjectCommand {
                 const {
                   compositionValidationResult,
                   checkSchemaResult
-                } = await project.engine.checkPartialSchema({
-                  id: graphName,
-                  graphVariant: tag,
+                } = await serviceProject.engine.checkPartialSchema({
+                  id: graphID!,
```
timbotnik: I think we should assert the existence of `graphID` and `graphVariant` and deal with that above instead of using non-null assertions here?  There are a few more of these assertions below, I can come back and mark them based on your thoughts here.

packages/apollo-language-server/src/config/loadConfig.ts
```diff
@@ -50,26 +52,69 @@ export interface LoadConfigSettings {
...
+  graphInfo: GraphInfo,
+  configPath?: string,
+  apiKey?: string
+): ApolloConfig {
+  const graphId = graphInfo.graphId;
+
+  const configWithDefaults = {
+    ...(loadedConfig && loadedConfig.config),
+    client: merge(
+      DefaultConfigBase,
```
timbotnik: We should put an `Object.create(null)` here in the first position to avoid mutation of `DefaultConfigBase`.

packages/apollo-language-server/src/config/loadConfig.ts
```diff
@@ -50,26 +52,69 @@ export interface LoadConfigSettings {
...
+      graphId
+        ? {
+            service: `${graphId}@${graphInfo.clientGraphVariant}`,
+            graphId
+          }
+        : {},
+      loadedConfig && loadedConfig.config.client
+    ),
+    service: merge(
+      DefaultConfigBase,
```
timbotnik: We should put an `Object.create(null)` here in the first position to avoid mutation of `DefaultConfigBase`.

packages/apollo-language-server/src/config/loadConfig.ts
```diff
@@ -50,26 +52,69 @@ export interface LoadConfigSettings {
...
+      graphId
+        ? {
+            name: `${graphId}@${graphInfo.serviceGraphVariant}`,
+            graphId
+          }
+        : {},
+      loadedConfig && loadedConfig.config.service
+    ),
+    engine: merge(
+      DefaultEngineConfig,
```
timbotnik: We should put an `Object.create(null)` here in the first position to avoid mutation of `DefaultEngineConfig `.

packages/apollo-language-server/src/config/__tests__/config.ts
```diff
@@ -1,11 +1,29 @@
...
+import {
+  ApolloConfig,
+  ApolloConfigFormat,
+  getGraphInfo,
+  loadConfigWithDefaults
+} from "../";
 import URI from "vscode-uri";
+import { DeepPartial } from "apollo-env";
+
+export function createConfig(raw: DeepPartial<ApolloConfigFormat>) {
```
timbotnik: Q: it feels strange that we now have to perform this when creating configs for tests... would it make sense to try to move this transformation into the `ApolloConfig` construction / initialization?

packages/apollo-language-server/src/config/__tests__/config.ts
```diff
@@ -24,52 +45,66 @@ describe("ApolloConfig", () => {
...
-      expect(projects).toHaveLength(1);
+      expect(projects).toHaveLength(2);
       expect(projects[0].isService).toBeTruthy();
+      expect(config.graphId).toEqual("my-service");
     });
     it("creates multiple configs when both client and service are present", () => {
-      const rawConfig: ApolloConfigFormat = {
+      const rawConfig: ApolloConfigFormat = createConfig({
         client: { service: "my-service" },
-        service: "my-service"
```
timbotnik: It feels like we are losing this test case, but maybe it's just an artefact of using this `createConfig` method.

packages/apollo-language-server/src/config/__tests__/config.ts
```diff
@@ -24,52 +45,66 @@ describe("ApolloConfig", () => {
...
+      });
       const config = new ApolloConfig(rawConfig);
       const projects = config.projects;
 
       expect(projects).toHaveLength(2);
       expect(projects.find(c => c.isClient)).toBeTruthy();
       expect(projects.find(c => c.isService)).toBeTruthy();
     });
+    it("creates multiple configs when nothing is present", () => {
+      const config = new ApolloConfig(createConfig({}));
```
timbotnik: This definitely feels like cheating since it's using a test-only function (`createConfig`) to set up conditions for success.

PR #1619, opened 2019-10-26
[Swift] Add documentation to input structs' init

kevinmbeaulieu:
Implement [feature request](https://github.com/apollographql/apollo-ios/issues/562) to add documentation to input structs' generated constructors.

Generates documentation of the form:
```
/// - Parameters:
///   - someUndocumentedField
///   - someDocumentedField: This field has a documentation comment. When it is really long,
/// it wraps like this.
```

The way the text wraps is admittedly not ideal, but I couldn't think of a good way to make it wrap like this:
```
///   - someDocumentedField: This field has a documentation comment. When it is really long,
///                          it would be nice if it wrapped like this.
```

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - `apollo-codegen-scala`
   - <First `apollo-codegen-scala` related entry goes here>
 - `apollo-codegen-swift`
-  - <First `apollo-codegen-swift` related entry goes here>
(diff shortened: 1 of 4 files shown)


zionts:
@designatednerd you might be the right person to take a look here :)

kevinmbeaulieu:
No prob, thanks! Just rebased so should be safe for someone to merge once CI finishes re-running

PR #1634, opened 2019-11-05
Fix client localSchemaFile for vscode stats command

JakeDawkins:
Currently, if you click on the Apollo logo on the status bar, stats for the project you're working with are supposed to show, to confirm the config was setup properly and the extension has loaded.

If you're setting up a client project and using a `localSchemaFile` though, the stats will fail to load unless there is a service `name` also provided, which isn't required.

This removes the check in the stats loading function that forced a client to have a `client.service.name` even if not using graph manager to load the schema.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 - `apollo-tools`
   - <First `apollo-tools` related entry goes here>
 - `vscode-apollo`
-  - <First `vscode-apollo` related entry goes here>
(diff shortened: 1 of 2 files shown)


trevor-scheer:
Can we tell if there's another configuration path where `serviceID` is required that this might affect? We might want something more like:
```js
Boolean((this.serviceID || this.localSchemaFile) && (this.schema || this.serviceSchema))
```
This is just a stab / guess.

JakeDawkins:
@trevor-scheer I can check, but since this is only used for logging output to vscode, I don't _think_ that'd matter. I tried something similar to what you dropped in but TS was v. angry 🙃

PR #1640, opened 2019-11-08
Supports User Tokens

caydie-tran:
This is a new PR that cherry picks the first several commits from https://github.com/apollographql/apollo-tooling/pull/1614

TODO:


packages/apollo-language-server/src/providers/schema/index.ts
```diff
@@ -25,19 +25,17 @@ export function schemaProviderFromConfig(
...
-  if (isServiceConfig(config)) {
-    if (config.service.localSchemaFile) {
-      const isListOfSchemaFiles = Array.isArray(config.service.localSchemaFile);
-      return new FileSchemaProvider(
-        isListOfSchemaFiles
-          ? { paths: config.service.localSchemaFile as string[] }
-          : { path: config.service.localSchemaFile as string }
-      );
-    }
+  if (config.service && config.service.endpoint) {
```
JakeDawkins: Why was the order of these flipped again? I remember there was a distinct reason, but I don't remember exactly what it was

packages/apollo/src/commands/service/check.ts
```diff
@@ -506,14 +502,14 @@ export default class ServiceCheck extends ProjectCommand {
                   : null;
 
                 task.title = `Compared ${pluralize(
-                  chalk.cyan(schemaChanges.length.toString()),
+                  chalk.blue(schemaChanges.length.toString()),
```
JakeDawkins: these should all still be `cyan` right?

packages/apollo/src/commands/service/list.ts
```diff
@@ -101,37 +101,33 @@ export default class ServiceList extends ProjectCommand {
     const taskOutput: TasksOutput = {};
 
     let schema: GraphQLSchema | undefined;
+    let graphID: string | undefined;
+    let graphVariant: string | undefined;
     try {
       await this.runTasks<TasksOutput>(({ config, flags, project }) => {
-        if (!isServiceProject(project)) {
```
JakeDawkins: This was for the sake of being able to run this command in a client project, right? I'm still not sure how I feel about allowing that from an ideological perspective, but technically, it's probably fine

JakeDawkins:
Tests are failing intermittently because of backend issues, but I've seen them all pass locally, so I'll merge anyway

PR #1683, opened 2019-11-25
Allow client and service sections to be mixed in apollo.config.js

joankaradimov:
This is useful for monorepo projects. The intention is to allow mixed configuration for client and service configuration in the top level `apollo.config.js` file. I found no problems during my initial testing.

The currently suggested approach is to have two separate config files. I just couldn't get it to work as I wanted it. But even if I did, the Apollo VSCode plugin would still have trouble handling both `apollo.config.js` files.


joankaradimov:
One of the reasons I didn't provide tests is that I assumed this might not be good enough for a final implementation.

> this will create a discrepancy between how configuration works with VSCode and with the CLI commands.

From what I've seen the VSCode extension is tricky. But I have the hunch it will work better with a single `apollo.config.js` in the repo root. I haven't delved deeper into that, though.

> If you're interested in taking on this initiative

I definitely am. And it will solve problems that I am encountering. I'll try to familiarize myself with #1614. Any other pointers?

PR #1701, opened 2019-12-05
[fix] add @types/node-fetch as a devDependency to apollo-env

### LINKED ISSUE #1686: [TypeScript] Could not find a declaration file for module 'node-fetch'
When using `apollo-env` through `apollo-server`, I get the following error during TypeScript builds:

```
node_modules/apollo-env/lib/fetch/fetch.d.ts(6,195): error TS7016: Could not find a declaration file for module 'node-fetch'
```

This can be fixed by adding `@types/node-fetch` as a dependency of `apollo-env`. Or as an alternative, update `node-fetch` to v3, as [it bundles type declarations](https://github.com/bitinn/node-fetch/issues/668) with itself.

lookfirst:
fixes #1686

I went with this instead of going to v3 of node-fetch because I didn't want to mess with changing to a major version.

One question I have is why this change isn't breaking the release build somewhere.... cause this is breaking my build now. =(

TODO:

### DIFF
--- a/packages/apollo-env/package.json
+++ b/packages/apollo-env/package.json
@@ -18,5 +18,8 @@
     "core-js": "^3.0.1",
     "node-fetch": "^2.2.0",
     "sha.js": "^2.4.11"
+  },
+  "devDependencies": {
+    "@types/node-fetch": "*"
   }
 }


trevor-scheer:
Thanks for taking a look @lookfirst, I've landed a fix here: https://github.com/apollographql/apollo-tooling/pull/1749

PR #1703, opened 2019-12-06
Support disabling literal stripping when extracting operations.

### LINKED ISSUE #1099: Extraction strips quoted operation parameters
**Intended outcome:**

Operation parameter values remain in the parsed operation string.

```
fragment Articles on Query {
  article: getArticles(
    sort: [{ field: "created", direction: ASC }],
  ){
    count
    entities {
      ...Article
    }
  }
}
```
Parses to:

```
fragment Articles on Query { article: getArticles(sort: [{ field: "created", direction: ASC }],){ count entities { ...Article } } }
```

**Note:** `sort.field` === `"created"`

**Actual outcome:**

Operation parameters values are stripped from the operation string.

```
fragment Articles on Query { article: getArticles(sort: [{ field: \"\", direction: ASC }],){ count entities { ...Article } } }
```

**Note:** `sort.field` === `""`

Run `client:extract` on a fragment such as:

```
fragment Articles on Query {
  article: getArticles(
    sort: [{ field: "created", direction: ASC }],
  ){
    count
    entities {
      entityId
    }
  }
}
```

**Versions**

`2.6.0`

damondoucet:
Fixes #1099

Adds a flag `--[no]-preserveStringAndNumericLiterals` to `client:extract`
that controls whether literals are preserved in extracted queries. Prior to
this flag, object and list literals were already preserved; that's unaffected
by this change.

As noted in the linked issue, there are a few reasons to do this, but generally
they boil down to relying on the extracted queries for some functionality, and
literals in the queries controlling important behavior (e.g. generating mock data
for queries might rely on literal args).

A flag is preferred to changing the behavior outright since older clients likely
rely on the current behavior.

TODO:

### DIFF
--- a/packages/apollo-graphql/src/operationId.ts
+++ b/packages/apollo-graphql/src/operationId.ts
@@ -72,7 +72,7 @@ export function defaultEngineReportingSignature(
 // sorting the AST in a deterministic manner, potentially hiding string and numeric
 // literals, and removing unused definitions. This is a less aggressive transform
 // than its engine reporting signature counterpart.
-export function defaultOperationRegistrySignature(
+export function operationRegistrySignature(
   ast: DocumentNode,
   operationName: string,
   options: { preserveStringAndNumericLiterals: boolean } = {
@@ -84,6 +84,15 @@ export function defaultOperationRegistrySignature(
     ? withoutUnusedDefs
     : hideStringAndNumericLiterals(withoutUnusedDefs);
   return printWithReducedWhitespace(sortAST(maybeWithLiterals));
+}
+
+export function defaultOperationRegistrySignature(
+  ast: DocumentNode,
+  operationName: string
+): string {
+  return operationRegistrySignature(ast, operationName, {
+    preserveStringAndNumericLiterals: false
+  });
 }
 
 export function operationHash(operation: string): string {
--- a/packages/apollo-graphql/src/operationId.ts
+++ b/packages/apollo-graphql/src/operationId.ts
@@ -75,12 +75,14 @@ export function defaultEngineReportingSignature(
 export function defaultOperationRegistrySignature(
   ast: DocumentNode,
   operationName: string,
-  options: { stripLiterals: boolean } = { stripLiterals: true }
+  options: { preserveStringAndNumericLiterals: boolean } = {
+    preserveStringAndNumericLiterals: false
+  }
 ): string {
   const withoutUnusedDefs = dropUnusedDefinitions(ast, operationName);
-  const maybeWithLiterals = options.stripLiterals
-    ? hideStringAndNumericLiterals(withoutUnusedDefs)
-    : withoutUnusedDefs;
+  const maybeWithLiterals = options.preserveStringAndNumericLiterals
+    ? withoutUnusedDefs
+    : hideStringAndNumericLiterals(withoutUnusedDefs);
   return printWithReducedWhitespace(sortAST(maybeWithLiterals));
 }
 
--- a/packages/apollo-graphql/src/operationId.ts
+++ b/packages/apollo-graphql/src/operationId.ts
@@ -69,25 +69,19 @@ export function defaultEngineReportingSignature(
 }
 
 // The operation registry signature function consists of removing extra whitespace,
-// sorting the AST in a deterministic manner, hiding string and numeric literals,
-// and removing unused definitions. This is a less aggressive transform than its
-// engine reporting signature counterpart.
-//
-// XXX
-// The hiding of literals is currently being discussed. This behavior
-// should not be depended on in the future, as it's likely we will choose not
-// to hide them at all. The rationale being, if queries are being shipped to
-// a client bundle, exposing PII via this signature is a very small concern,
-// relatively speaking.
+// sorting the AST in a deterministic manner, potentially hiding string and numeric
+// literals, and removing unused definitions. This is a less aggressive transform
+// than its engine reporting signature counterpart.
 export function defaultOperationRegistrySignature(
   ast: DocumentNode,
-  operationName: string
+  operationName: string,
+  options: { stripLiterals: boolean } = { stripLiterals: true }
 ): string {
-  return printWithReducedWhitespace(
-    sortAST(
-      hideStringAndNumericLiterals(dropUnusedDefinitions(ast, operationName))
-    )
-  );
+  const withoutUnusedDefs = dropUnusedDefinitions(ast, operationName);
+  const maybeWithLiterals = options.stripLiterals
+    ? hideStringAndNumericLiterals(withoutUnusedDefs)
+    : withoutUnusedDefs;
+  return printWithReducedWhitespace(sortAST(maybeWithLiterals));
 }
 
 export function operationHash(operation: string): string {
--- a/packages/apollo/src/commands/client/extract.ts
+++ b/packages/apollo/src/commands/client/extract.ts
@@ -11,11 +11,13 @@ export default class ClientExtract extends ClientCommand {
   static description = "Extract queries from a client";
   static flags = {
     ...ClientCommand.flags,
-    stripLiterals: flags.boolean({
+    preserveStringAndNumericLiterals: flags.boolean({
       description:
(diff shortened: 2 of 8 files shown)


damondoucet:
I left the `push` command alone, since the issue only references the `extract`
command, but I'd be happy to adjust that command as well if it's valuable!

damondoucet:
I also wasn't sure about the casing of the `no` prefix - it looks a little funny
to me, but I figured I'd default to OCLIF's behavior unless encouraged otherwise. :)

jhampton:
Thanks @damondoucet !  I'd like input from @jbaxleyiii and the team as to whether we should include the `push` command, but this looks clean to me.

Cheers!

jhampton:
👍 LGTM

packages/apollo/src/commands/client/extract.ts
```diff
@@ -9,7 +10,13 @@ import { ClientIdentity } from "apollo-language-server";
 export default class ClientExtract extends ClientCommand {
   static description = "Extract queries from a client";
   static flags = {
-    ...ClientCommand.flags
+    ...ClientCommand.flags,
+    stripLiterals: flags.boolean({
```
abernix: Since "literals" also applies to object and list literals (which are not stripped), perhaps we should be a bit more clear with this option name.  Even if it's a bit wordy, the intent is more clear, I think?  (And its less necessary for someone to understand the notion of negated flags and perhaps more intrinsic understanding of the default behavior?).  This change would require changing the `default` to `false` below and flipping other conditionals too, of course.)

```suggestion
    preserveStringAndNumericLiterals: flags.boolean({
```

packages/apollo/src/commands/client/extract.ts
```diff
@@ -9,7 +10,13 @@ import { ClientIdentity } from "apollo-language-server";
...
+      description:
+        "Whether to strip literals from extracted queries. DEFAULT: true",
+      default: true,
```
abernix: If heeding my suggestion above, this default would become `false`.

packages/apollo/src/commands/client/extract.ts
```diff
@@ -9,7 +10,13 @@ import { ClientIdentity } from "apollo-language-server";
 export default class ClientExtract extends ClientCommand {
   static description = "Extract queries from a client";
   static flags = {
-    ...ClientCommand.flags
+    ...ClientCommand.flags,
+    stripLiterals: flags.boolean({
+      description:
+        "Whether to strip literals from extracted queries. DEFAULT: true",
```
abernix:
```suggestion
        "Disable redaction of string and numerical literals.  Without this flag, these values will be replaced with empty strings (`''`) and zeroes (`0`) respectively.  This redaction is intended to avoid inadvertently outputting potentially personally identifiable information (e.g. embedded passwords or API keys) into operation manifests.",
```

packages/apollo/src/commands/client/extract.ts
```diff
@@ -9,7 +10,13 @@ import { ClientIdentity } from "apollo-language-server";
...
+      default: true,
+      allowNo: true
```
abernix: I believe we could get rid of this if the default was `false`.

```suggestion
```

packages/apollo/src/commands/client/extract.ts
```diff
@@ -30,7 +37,9 @@ export default class ClientExtract extends ClientCommand {
       {
         title: "Extracting operations from project",
         task: async ctx => {
-          ctx.operations = getOperationManifestFromProject(this.project);
+          ctx.operations = getOperationManifestFromProject(this.project, {
+            stripLiterals: flags.stripLiterals
```
abernix: I would also suggest renaming this to more accurately describe its scope:
```suggestion
            preserveStringAndNumericLiterals: flags.preserveStringAndNumericLiterals
```

packages/apollo/src/utils/getOperationManifestFromProject.ts
```diff
@@ -13,14 +13,16 @@ export interface ManifestEntry {
 }
 
 export function getOperationManifestFromProject(
-  project: GraphQLClientProject
+  project: GraphQLClientProject,
+  options: { stripLiterals: boolean } = { stripLiterals: true }
```
abernix:
```suggestion
  options: {
    preserveStringAndNumericLiterals: boolean,
  } = {
    preserveStringAndNumericLiterals: true,
  }
```

packages/apollo/src/utils/getOperationManifestFromProject.ts
```diff
@@ -13,14 +13,16 @@ export interface ManifestEntry {
...
 ): ManifestEntry[] {
   const manifest = Object.entries(
     project.mergedOperationsAndFragmentsForService
   ).map(([operationName, operationAST]) => {
     const printed = defaultOperationRegistrySignature(
       operationAST,
-      operationName
+      operationName,
+      { stripLiterals: options.stripLiterals }
```
abernix:
```suggestion
      {
        preserveStringAndNumericLiterals:
          options.preserveStringAndNumericLiterals
      }
```

damondoucet:
Hey Jesse, thanks for getting back to me! So as not to leave this hanging - I had a chance to skim your feedback and it generally makes sense, but I won't be able to update this PR until the new year. It'll be a high priority for us then (1/7 or so). Happy holidays!

packages/apollo/src/utils/getOperationManifestFromProject.ts
damondoucet: Presumably, the default should also flip to false? :)

damondoucet:
Thanks for the feedback @abernix. I read through and it all makes sense to me, so I've updated the PR with the corresponding changes. Let me know what you think!

damondoucet:
bump @abernix - would be great to get this merged as soon as we can since it's blocking a high priority feature for us. thanks!

damondoucet:
@abernix friendly bump - any updates on the timeline for getting this merged?

packages/apollo-graphql/src/operationId.ts
```diff
@@ -69,25 +69,21 @@ export function defaultEngineReportingSignature(
...
+// sorting the AST in a deterministic manner, potentially hiding string and numeric
+// literals, and removing unused definitions. This is a less aggressive transform
+// than its engine reporting signature counterpart.
 export function defaultOperationRegistrySignature(
   ast: DocumentNode,
-  operationName: string
+  operationName: string,
+  options: { preserveStringAndNumericLiterals: boolean } = {
+    preserveStringAndNumericLiterals: false
+  }
```
abernix: Introducing options to a method called `defaultOperationRegistrySignature` seems a bit off since a "default" should be sans options.  Could we rename this method to `operationRegistrySignature` and introduce another method using the existing name that merely calls this with an explicit `false` for `preserveStringAndNumericLiterals` in order to make the "default" intentions clear?

damondoucet: Thanks! I made that change, though I'll note this may not actually be merge-able in its current state. The main issue is that the tests generally prefer the default value for `options`, but in some cases want to override it.

To avoid the duplication between tests and `defaultOperationRegistrySignature`, I left the default value in `operationRegistrySignature`, but that sort of defeats the purpose of `defaultOperationRegistrySignature`. My personal recommendation would be to deprecate `defaultOperationRegistrySignature` (I think that's part of the external API of this library?), but I'm curious what you think!

Tangentially, I noticed you're in Finland, which may make this back-and-forth take longer than it otherwise ought to. If you think it'd be easier, I'm totally open to you/the Apollo team taking over the PR from here, but I'm also happy to do another round of changes. As I've mentioned, my main goal is to get this shipped quickly to unblock my team. :)

jhampton: I believe you're right that `defaultOperationRegistrySignature` is a part of the API.  This LGTM, thank y'all.

damondoucet:
Woo, thanks!!

abernix:
@damondoucet Can you try and see if this is working for you in the latest version of `apollo` (`apollo@2.22.0`)?

damondoucet:
it works, thanks @abernix!

PR #1745, opened 2020-01-06
Fix tabular output for unrecognized errors

sponomarev:
"table" package is strict about data and fails on inconsistent rows dimension.

For that kind of composition error

```js
[
  {
    message: 'Field "Mutation.createSomething" can only be defined once.'
  }
]
```

`service:check` ends up with an exception

```
Error: Table must have a consistent number of cells.
    at validateTableData (~/Projects/sandbox/node_modules/table/dist/validateTableData.js:47:15)
    at Object.table (~/Projects/sandbox/node_modules/table/dist/table.js:96:34)
    at ServiceCheck.run (~/Projects/sandbox/node_modules/apollo/lib/commands/service/check.js:326:30)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Upcoming
 
 - `apollo`
-  - <First `apollo` related entry goes here>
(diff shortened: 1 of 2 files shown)


sponomarev:
Closed in favor of #1806

PR #1766, opened 2020-01-19
feat: add tsUseOptionalForNullables option to ts codegen

HosseinAgha:
`tsUseOptionalForNullables` makes nullable fields optional instead of a union of field type and null.  
As discussed in #1622, this issue makes sending an Apollo client query with nullable field to standard React components typed with TypeScript very painfull. You have to add `| null` to optional props of every component you have and add a wrapper for every Library component you use (Material UI uses optional props for example).  

Avoiding the edge cases (like `typeof field === 'object'`) that comes with enabling this config (Which is not the case 99% of times) is the developer's responsibility as this is disabled by default.  

This also makes integrating Apollo into existing systems much easier.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - `apollo-codegen-swift`
   - <First `apollo-codegen-swift` related entry goes here>
 - `apollo-codegen-typescript`
-  - <First `apollo-codegen-typescript` related entry goes here>
+  - add `tsUseOptionalForNullables` config [#1766](https://github.com/apollographql/apollo-tooling/pull/1766) and [#622](https://github.com/apollographql/apollo-tooling/issues/622)
 - `apollo-env`
   - <First `apollo-env` related entry goes here>
 - `apollo-graphql`
--- a/packages/apollo-codegen-core/src/compiler/index.ts
+++ b/packages/apollo-codegen-core/src/compiler/index.ts
@@ -45,6 +45,7 @@ export interface CompilerOptions {
   // this option is only implemented in the ts codegen, so we name it
   // `ts` fileExtension for now.
   tsFileExtension?: string;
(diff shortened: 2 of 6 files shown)


marcelltoth:
I love the idea. I would really want to use this, in my projects, 99% of the time the distinction between `null` and `undefined` in apollo just makes TypeScript code more bloated for no reason.

However, this needs to be consistent, unfortunately. Generating an optional type while `apollo-client` will actually return `null` is very dangerous. I've seen your warning, but I'm sure that at some point, some WILL write `field !== undefined`. For this reason I believe this feature would only work if apollo implemented actual support for `undefined` beforehand. Which is - IMO - something that'd benefit the TypeScript community a lot, but unfortunately it is not there yet, and will probably never get there either. (https://github.com/apollographql/apollo-client/issues/2412)

Stupidism:
Can't wait to use this option. What a savior it will be!

Stupidism:
I really love this idea! @HosseinAgha   Why is this not pushed forward?

HosseinAgha:
@Stupidism  
I don't know! I'm waiting for a maintainer.  
As this is actually a wrong type from JS perspective I added an option. But looks like Apollo ecosystem maintenance is almost stopped.  
Meanwhile, I've switched to [Graphql Codegen](https://www.graphql-code-generator.com/docs/getting-started/index) and it rules!

HosseinAgha:
This PR is open without a single maintainer comment for 4 years.  
I think Apollo may resemble the current state of the GraphQL community.  
I don't think GraphQL.js will survive considering the investment that the general JS community are making.  Unfortunately advancements are happening elsewhere in the JS ecosystem (mainly center stack data fetching).  I still believe in the promise and technical relevancy of GraphQL (and Apollo) but considering the current situation I don't have much hope for it.

PR #1806, opened 2020-02-20
Fix display of unformatted errors in composition

JakeDawkins:
Right now, `apollo service:check` for federated services expects a specifically shaped error message from composition errors, but not all errors are the expected shape. When that happens, we are not formatting the errors properly for display, and users are seeing a cryptic error about an inconsistent number of cells in the display table. This PR aims to fix that by displaying unformatted (unexpected) errors separately from the ones we can recognize and format.

before:

Screen Shot 2020-02-20 at 11 49 35 AM

after: 

Screen Shot 2020-02-20 at 11 49 11 AM

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 - `apollo`
   - Update shortlinks to use go.apollo.dev instead of bitly [#1790](https://github.com/apollographql/apollo-tooling/pull/1790)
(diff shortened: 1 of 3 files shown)


sponomarev:
@JakeDawkins The same problem has been adressed in #1745 early and was ignored 😞

PR #1812, opened 2020-02-24
Make duplicate operations error a failing error

JakeDawkins:
This PR aims to fix an issue where some users are pushing client code that errors because the error notifying users of duplicate client operation names doesn't exit appropriately.

the previous error, just told users that duplicate operations aren't permitted in client projects, but never actually _enforced_ that as it should have. This PR should fix that behavior.

TODO:


JakeDawkins:
I believe this is good to go, so I'm gonna give it a merge and minor release. 

I tested manually with client:push and client:check and verified it failed both before trying to actually run the operation (which is what we want).

I think this error could be cleaned up (as right now it's a very unfriendly format when it's output), potentially even as a step on `client:*` commands, but as for right now, I think that work would be best pushed to a major, since it'd require reworking the project loading as well as VS Code errors a bit.

PR #1847, opened 2020-03-25
Update "no service found to link..." error

zionts:
This updates the error message that occurs when a config.service.name is
required in order to execute a command. Before this commit, there were
two distinct messages that were used in commands to communicate this
error to the user, which were slightly different in their wording. This
updates the wording to provide an explanation as well as next steps on
how to remedy the issue. Additionally, it uses the `CLIError` class in
order to print out an error in a "friendly" way without a stack trace,
as recommended by the @oclif/errors package we are using: https://oclif.io/docs/commands#thiserrormessage-string--error-options-code-string-exit-number

For changelog entry, please see https://github.com/apollographql/apollo-tooling/pull/1849

TODO:

### DIFF
--- a/packages/apollo/src/commands/service/list.ts
+++ b/packages/apollo/src/commands/service/list.ts
@@ -4,17 +4,12 @@ import { GraphQLSchema } from "graphql";
 import sortBy from "lodash.sortby";
 import { table } from "table";
 import moment from "moment";
-import {
-  ApolloConfig,
-  isServiceProject,
-  DefaultEngineConfig
-} from "apollo-language-server";
+import { ApolloConfig, DefaultEngineConfig } from "apollo-language-server";
 import chalk from "chalk";
 import {
   ListServices_service_implementingServices,
   ListServices_service_implementingServices_FederatedImplementingServices_services
 } from "apollo-language-server/lib/graphqlTypes";
-import { CLIError } from "@oclif/errors";
 import { graphUndefinedError } from "../../utils/errors";
 
 interface TasksOutput {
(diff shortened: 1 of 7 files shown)


zionts:
Before: 
image

After: 
image

zionts:
Update -- Added a commit (that we can revert if we'd like) that makes the error message split with newlines instead of spaces. This makes it more likely to fit in your shell:
image

packages/apollo/src/utils/errors.ts
```diff
@@ -0,0 +1,9 @@
+import { CLIError } from "@oclif/errors";
+
+const errorMessage = [
+  "No graph (i.e. service) found to link to Apollo Graph Manager.",
+  "In order to run this command, please provide a graph ID using the 'apollo.config.js' file.",
+  "\n\nFor more information on configuring the Apollo CLI, please go to",
+  "https://www.apollographql.com/docs/devtools/apollo-config/"
```
JakeDawkins:
```suggestion
  "https://go.apollo.dev/t/config"
```

Can you use the short link here instead just in case we change that in the future? 😊

packages/apollo/src/commands/service/list.ts
```diff
@@ -14,6 +14,8 @@ import {
   ListServices_service_implementingServices,
   ListServices_service_implementingServices_FederatedImplementingServices_services
 } from "apollo-language-server/lib/graphqlTypes";
+import { CLIError } from "@oclif/errors";
```
JakeDawkins: Do you need to import this?

PR #1849, opened 2020-03-25
Create --variant flag, deprecate --tag flag

zionts:
This PR changes all commands that previously used a `--tag` flag to additionally include a `--variant` (alias -v) flag that can be used instead. These two different flags are exclusive, so no command can pass both a --tag and a --variant. **The PR may look daunting at a glance, but please go commit-by-commit to see the progression, and I hope that will make it really manageable!** These changes were QA'ed manually, and all tests were updated and inspected to ensure that they continued to work as expected. Some other small changes were added, which are noted in commit messages.

This is rebased off of #1847 , so check that one out first :) 
See https://github.com/apollographql/apollo/pull/819 for the corresponding documentation update 🎉 

Notably, they include:
* Changing messages in `service:check` to include `@variant` where appropriate
* Changing (buggy) logic in several places that were using `flags.tag` directly rather than `config.tag`, in GraphQL commands. This could have caused issues for people that were not using a `--tag` flag but instead were using `service { name: 'blah@foo' }`, in that their command would fail unexpectedly.

Error message when both --tag and --variant are set:
image

`client:push` with --tag (note deprecation warning):
image

`client:push` with --variant: 
image

`client:check` with --tag: 
image

`client:check` with --variant:
image

`client:codegen` with --tag:
image

`client:codegen` with --variant:
image

`service:push` (federated) with --tag:
image

`service:push` (un-federated) with --variant:
image

`service:check` (un-federated) with --tag:
image

`service:check` (federated) with --variant:
image

`service:list` with --tag:
image

`service:list` with --variant:
image

`service:list` with neither --tag nor --variant, just to show that defaults work :)
image

Also throwback to #1398 🎉 Credit to former me & @jgzuke for inspiration, along with whomever made the `config` magic (which I think may have been @trevor-scheer?), which I relied on heavily here :)

TODO:


packages/apollo-language-server/src/config/config.ts
```diff
@@ -162,12 +162,12 @@ export class ApolloConfig {
     return configs;
   }
 
-  set tag(tag: string) {
-    this._tag = tag;
+  set variant(tag: string) {
```
JakeDawkins: My only concern is if this breaks anything in vs code. But I don’t think it should

CHANGELOG.md
```diff
@@ -3,7 +3,14 @@
 ## Upcoming
 
 - `apollo`
-  - <First `apollo` related entry goes here>
+  - https://github.com/apollographql/apollo-tooling/pull/1849
```
JakeDawkins: Formatting of this entry isn’t consistent with what we have but I’m not concerned too much there, and that’s non blocking. I can amend before releasing since I’ll have to do that anyway

packages/apollo-language-server/src/config/config.ts
zionts: I admit I didn't test with VS code, but mainly because I don't know how to test a local build! Is there a QA practice we go through before release?

JakeDawkins: Not in a consistent way! I’ll run through all this before releasing, but I don’t expect any issue here since it only interfaces with the language server anyway :)

PR #1855, opened 2020-03-27
Improve error message for failing to provide service endpoint

StephenBarlow:
Currently if you run `apollo service:push` failing to provide an endpoint or localSchemaFile, you get an unhelpful error:

```
TypeError: Cannot read property 'startsWith' of undefined
```

### DIFF
--- a/packages/apollo-language-server/src/providers/schema/endpoint.ts
+++ b/packages/apollo-language-server/src/providers/schema/endpoint.ts
@@ -29,6 +29,17 @@ export class EndpointSchemaProvider implements GraphQLSchemaProvider {
       uri: url,
       fetch
     };
+
+    if (!url) {
+      throw new Error(
+        "This command requires either the --endpoint option or the" +
+          "\n--localSchemaFile option to obtain your service's schema." +
+          "\nYou can also provide one of these values in your apollo.config.js file.\n" +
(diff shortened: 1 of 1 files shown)


packages/apollo-language-server/src/providers/schema/endpoint.ts
```diff
@@ -29,6 +29,17 @@ export class EndpointSchemaProvider implements GraphQLSchemaProvider {
       uri: url,
       fetch
     };
+
+    if (!url) {
+      throw new Error(
```
zionts: Could we throw a CLIError here to get proper formatting? :)

zionts:
Additionally, there might be some way to handle a `oneof` style requirement of flags with oclif

packages/apollo-language-server/src/providers/schema/endpoint.ts
zionts: Resolving this as `CLIError` is only available in the CLI code. I do think this error message will never trigger, though, because the default is to make `--endpoint` localhost:4000

zionts:
@StephenBarlow , looking more into this, it seems that `url` must be defined. Can you provide a reproduction that will reach this error message? Seems like something should be failing compilation if this is possible!

PR #1856, opened 2020-03-27
service:check Add ignoreFailures flag

### LINKED ISSUE #1852: Allow exit 0 on errors reported by schema check
It'd be great to have a `apollo schema:check` option to exit 0 if the check finished, even if it returned errors, so we can fail in CI when the checks are broken, but not when they report errors. The build should be green in GitHub, the Apollo checks are where the red goes.

OTOH, ignoring the command exit code means we don't catch checks being broken in other ways (invalid command, bad API key, service is down, etc.).

zionts:
When using the GitHub integration, it's often recommended to run
service:check && echo "ignoring output" because GH status will be
updated by the command. Adding this option makes using the command with
GH integration cleaner by exiting with 0 on all schema errors.

Fixes #1852 

TODO:

### DIFF
--- a/packages/apollo-language-server/src/graphqlTypes.ts
+++ b/packages/apollo-language-server/src/graphqlTypes.ts
@@ -69,14 +69,14 @@ export interface CheckPartialSchema_service_checkPartialSchema_checkSchemaResult
   /**
    * delta in seconds from current time that determines the start of the window
    * for reported metrics included in a schema diff. A day window from the present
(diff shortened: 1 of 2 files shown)


JakeDawkins:
@zionts I think the naming of this flag could be a little more clear. What do you think about something like `--ignoreFailures`? I'm just concerned the name of this flag has nothing to do with errors

PR #1865, opened 2020-04-01
Remove "hidden: true" from variant flag in service:push

sachindshinde:
When running `npx apollo service:push --help`, there is no entry for `--variant`, due to `hidden: true` being used in the options for the flag (#1849). This PR removes the `hidden: true`.

TODO:


zionts:
FWIW I don't think this needs a changelog entry.

PR #1866, opened 2020-04-01
Update deprecation message of ENGINE_API_KEY to include alternative of APOLLO_KEY

zionts:
The message that appeared for users using `ENGINE_API_KEY` was not actionable before this change. This updates the message to suggest using `APOLLO_KEY` in lieu of `ENGINE_API_KEY` :)

I don't _think_ this kind of change really necessitates a CHANGELOG entry, but happy to add one if requested.

TODO:


JakeDawkins:
Thanks @zionts!

PR #1892, opened 2020-04-13
Fix .env api key taking precedence over localSchemaFile

JakeDawkins:
Previously, if a client project was using a `localSchemaFile` to write operations against but _also_ had a `.env` file with an `APOLLO_KEY` defined, the language server would try to load the schema from apollo graph manager, without regard to the local schema file.

The reason this was happening was in `loadConfig`, the service override at the end where the final config gets built was overwriting the existing `client.service` if there was a `serviceName`. This change should check for a `client.service` to prevent that accidental overwrite.

TODO:

### DIFF
--- a/packages/apollo-language-server/src/config/loadConfig.ts
+++ b/packages/apollo-language-server/src/config/loadConfig.ts
@@ -189,7 +189,13 @@ export async function loadConfig({
               client: {
                 ...DefaultConfigBase,
                 ...(loadedConfig && loadedConfig.config.client),
-                service: serviceName
+                // if there was a `client.service` defined in the loaded config file,
+                // prefer that over the service name defined by the api key in .env
+                ...(loadedConfig &&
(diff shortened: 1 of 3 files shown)


joemckie:
Would this also fix the env var taking precedence over the `url` field? I have the same issue as you described, however our codebase has a code-first schema so it uses a locally running service as opposed to a local schema file.

JakeDawkins:
@joemckie Ah, I don't think I accounted for that in this yet. Good thought. I'll make sure to update it :)

joemckie:
@JakeDawkins thanks so much!

joemckie:
FWIW I think it may work as it is, as the `url` property is part of `client.service`

Nicoowr:
Hi is there any news on this PR? the issue it solves is quite problematic

macrozone:
any update on this? this is seriously affecting developing in vscode.

PR #1894, opened 2020-04-14
Ignore non-file documents when checking operation duplicates

shroudedcode:
I'm using the Apollo GraphQL extension for VS Code and I regularly run into `️️There are multiple definitions for the '...' operation.` errors when viewing Git diffs in the side bar.

That's because whenever a new editor is opened [`GraphQLProject#documentDidChange`](https://github.com/shroudedcode/apollo-tooling/blob/cffba044d2fd5057ad6a762fbfc917d09e1e68fa/packages/apollo-language-server/src/project/base.ts#L210) is called with the editor's `TextDocument` object. The document's URI and contents are then stored and later used for checking whether there are duplicate operations in the project. This happens regardless of whether the document is an actual file (URI starting with `file://`) or something else like a Git diff (URI starting with `git:`).

This PR adds a check to ensure only actual files are scanned for duplicate operation names.

### DIFF
--- a/packages/apollo-language-server/src/project/base.ts
+++ b/packages/apollo-language-server/src/project/base.ts
@@ -208,6 +208,8 @@ export abstract class GraphQLProject implements GraphQLSchemaProvider {
   }
 
   documentDidChange(document: TextDocument) {
+    if (!document.uri.startsWith("file://")) return;
+
     const documents = extractGraphQLDocuments(
       document,
       this.config.client && this.config.client.tagName


nibtime:
This is a very important fix! It actually renders the extension kind of unusable if you use the diff view (which probably every developer using VS Code does). Maybe that's even the reason it has such a low rating on the VS Code Marketplace because apart from this showstopper it's a great extension.

I applied this fix manually in the `node_modules` folder of the extension and haven't had any problems since.

chinanderm:
That is the status of this? This bug is rendering the extension unusable for me.

shroudedcode:
At this point I don't even remember whether I already signed the CLA or not, but it seems like either way this project is pretty much abandoned. It took me like 10 minutes to run into this issue after installing the extension, so clearly even the developers don't seem to care about it (anymore?).

I'm not using any Apollo tooling anymore and am not really interested in worrying about this PR, so I'm closing it and unsubscribing. Feel free to "steal" my changes and open a new PR or properly fix the issue as @mattlubner suggested.

Issue #1061, opened 2019-02-28
VS Code (Go to Definition) is slow when Apollo GraphQL is enabled


tho-graf:
Hi there,

Since yesterday (I have Apollo GraphQL v1.5.2 and I think it is released since yesterday) my VS Code is really slow, like 10-20 seconds, when I use "Go to Definition" by ctrl + clicking, F12 or right-click -> "Go to Definition". I am using ubuntu, but my coworker with osx has the same problem.

If I disable Apollo GraphQL it's fast again, so I am pretty sure the performance leads back to this extension. 

Can you reproduce the bug? If you need more details I am happy to help.

Version:

Apollo GraphQL 1.5.2

My VS Code:
Version: 1.31.1
Commit: 1b8e8302e405050205e69b59abb3559592bb9e60
Date: 2019-02-12T02:19:29.629Z
Electron: 3.1.2
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Linux x64 4.15.0-45-generic

thearrow:
Noticed this as well over the past couple days - more broadly: all intellisense suggestions in typescript files, all go to definition commands, even find references are all excruciatingly slow. Disabling the Apollo GraphQL extension restores performance. 

Tried profiling extensions before I realized this one was the culprit - here is a profile trying multiple times to get an intellisense suggestion to show up (with this extension enabled, version 1.5.2):

[profile.cpuprofile.zip](https://github.com/apollographql/apollo-tooling/files/2916478/profile.cpuprofile.zip)

JakeDawkins:
Interesting... I can't seem to reproduce this in 1.5.2. We also have some people with a huge number of types/fields in their projects that aren't seeing this, so I'm curious as to what could be causing it. Have either of you seen this happening in specific projects, or is it in any project of any size?

thomaskempel:
Actually our graph is not too big maybe 10 types and 50 fields (have to look at it tomorrow). Maybe it's because we have a lerna project and the apollo config file (for vs code) is in the root? @thearrow do you have lerna as well?

gund:
Happens to me as well. All autocomplete results are slow when this plugin is active.
Have to keep it disabled and only enable when actually working with graphql queries...

thearrow:
No lerna, single project, not a particularly large or complex graph. `apollo.config.js` is in the root of the project and it looks like this:

```ts
module.exports = {
  client: {
    name: "project_name",
[...]

PR #1938, opened 2020-05-08
[apollo-language-server] Improve performance for FileSet#includesFile

threehams:
Fixes: https://github.com/apollographql/apollo-tooling/issues/1061.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+### 1.19.4
+
+- Fix VS Code extension performance issues for larger projects [#1938](https://github.com/apollographql/apollo-tooling/pull/1938)
+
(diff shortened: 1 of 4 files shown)


threehams:
This code does work in Windows, within the VS Code extension, but the test environment is just slightly different from what VS Code does, and `minimatch` doesn't match `C:/` vs. `c:/`.

It'll take some time to figure this out, but the fix will be the same.

jgzuke:
@threehams thanks a ton for looking into and fixing this 🎉 I dont know this code super well yet but this change makes sense, and works great 😄 We're currently in the process of moving this code over to https://github.com/apollographql/vscode-graphql, I'm going to merge this into a transition branch and push it over there. Will update on the issue when this gets released, which will hopefully be tomorrow.

> the test environment is just slightly different from what VS Code does, and minimatch doesn't match C:/ vs. c:/.

I'm going to go through and get tests working in the new repo anyways, so will fix that then, no worries 😄

Issue #230, opened 2017-08-17
Unescaped " in deprecation message


victorrr:
In version 0.16.5 the deprecation messages contains "..." in the message. This generates errors in API.swift 

![skarmavbild 2017-08-17 kl 12 20 33](https://user-images.githubusercontent.com/116061/29408036-f9db8f42-8346-11e7-9da2-db5c8616b52a.png)

PR #1950, opened 2020-05-15
Swift: Make selections a var instead of a let

designatednerd:
A workaround for [iOS #230](https://github.com/apollographql/apollo-ios/issues/230) - the way that Swift allocates `static lets` was basically eating through memory that it shouldn't been. This transitions to using a computed `static var` instead, which should alleviate the problem. 

~I did look into just making it a computed var but there were some...complications... with the formatting that I wasted far too much time on trying to make work, and this achieves the same effect with a more minimal change.~ Got it working as a computed `var` after stepping away from the code for a weekend.

### DIFF
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -471,14 +471,17 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
 
         this.printNewlineIfNeeded();
         this.printOnNewline(
-          swift`public private(set) static var selections: [GraphQLSelection] = `
+          swift`public static var selections: [GraphQLSelection] {`
         );
-        if (typeCase) {
-          this.typeCaseInitialization(typeCase);
-        } else {
-          this.selectionSetInitialization(variant);
-        }
-
+        this.withIndent(() => {
+          this.printOnNewline(swift`return `);
+          if (typeCase) {
+            this.typeCaseInitialization(typeCase);
+          } else {
+            this.selectionSetInitialization(variant);
+          }
+        });
+        this.printOnNewline(swift`}`);
         this.printNewlineIfNeeded();
 
         this.printOnNewline(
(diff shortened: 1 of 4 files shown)


designatednerd:
Validated that the .zip produced by CI works with the current iOS and all tests still pass 🎉- will have to bump the iOS version once we have a new release for tooling but hey, at least it works in a vacuum 😆

packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -471,7 +471,7 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
 
         this.printNewlineIfNeeded();
         this.printOnNewline(
-          swift`public static let selections: [GraphQLSelection] = `
+          swift`public private(set) static var selections: [GraphQLSelection] = `
```
gsabran, line 474: I think that as long as the variable is static this will be retained in memory. What about
```suggestion
          swift`public var selections: [GraphQLSelection] { `
```

or maybe for old swift versions that don't support implicit return, which is a bit ugly as the line is wrong
```suggestion
          swift`public var selections: [GraphQLSelection] { return`
```

and close the `}` on line 481?

designatednerd: Oof, yeah turning it into an instance var won't work because we need to be able to access it without an instance.

back to the drawing board...

gsabran: Also, feel free to ask on the issue how important this is for folks to have the right priority against the codegen rewrite work you're doing. As this looks to turn out to be a decent amount of work.

designatednerd: Yeah...this was my procrastination project that somehow got away from me. 

However, I took a different approach on take 2, and it was WAY easier and still works. Yet another reminder to come back to stuff the next day when it's being dumb instead of just beating your head on the wall 🙃

designatednerd:
@gsabran updated to computed var in manner I can't believe I didn't think of the other day. LMK what you think here.

PR #1969, opened 2020-05-27
Support Union types in the exported IR typesUsed property

hwillson:
When using `codegen`'s `json-modern` export target, Union types will now show in the exported IR `typesUsed` property. E.g.

```json
"typesUsed": [
  {
    "kind": "UnionType",
    "name": "SearchResult",
    "types": [
      "Human",
      "Droid",
      "Starship"
    ]
  }
]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,16 +3,16 @@
 ## Upcoming
 
 - `apollo`
-  - Fix CLI crashing when trying to open a directory that is named like a file [#1948](https://github.com/apollographql/apollo-tooling/pull/1948).
+  - <First `apollo` related entry goes here>
 - `apollo-codegen-flow`
   - <First `apollo-codegen-flow` related entry goes here>
 - `apollo-codegen-scala`
   - <First `apollo-codegen-scala` related entry goes here>
 - `apollo-codegen-swift`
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,13 @@
 - `vscode-apollo`
   - <First `vscode-apollo` related entry goes here>
 
+## `apollo@2.28.0`
```
hwillson, line 28: It looks like the `CHANGELOG` was a bit out of date, so I moved the items that were sitting in `Upcoming` down into a section for the latest version.

packages/apollo-codegen-core/src/serializeToJSON.ts
```diff
@@ -51,13 +53,15 @@ export function serializeAST(ast: any, space?: string) {
   );
 }
 
-function serializeType(type: GraphQLType, options: serializeOptions) {
+function serializeType(type: GraphQLType, options?: serializeOptions) {
```
hwillson, line 56: Making `options` optional here isn't really related to the work in this PR, but TS was throwing a warning in the existing test file I had to modify. Looking over the code, `options` is truly optional, so let's make that evident.

designatednerd:
@hwillson This looks reasonable to me

hwillson:
Well, TIL - @jbaxleyiii has helped me understand that `typesUsed` is reserved for input values, which of course Unions can't be. Closing - we'll find another solution for this.

PR #2015, opened 2020-06-24
Update generated types using `npm run client:codegen`.

abernix:
Time for another update!


abernix:
@AlexanderMann Can you confirm that I should remove `ChangeSeverity.WARNING` from the (Node.js; current) `apollo` CLI? i.e.:

https://github.com/apollographql/apollo-tooling/blob/513037355e37d8d7d3e841a1968d6f970333f266/packages/apollo/src/commands/service/check.ts#L33-L39

AlexanderMann:
@abernix yes, this is safe to do. Mind you, we may add something like this back in the future, but we were unable to detect any usage of this going back months. As such, we updated our backend code to only have the binary decision of info/break

PR #2017, opened 2020-06-26
Add tests for "swift: Fix --operationIdsPath option" #1431

manicmaniac:
I took over the rest tasks (adding tests) of #1431.
Could you make sure that it is enough?

- Rebased onto the current HEAD of `master` branch
- Modified existing tests and updated snapshots
- Confirmed the tests passed in my local env
    - However it looks flaky so I ran tests with `node node_modules/.bin/jest -t generate -i --no-cache` to exclude unchanged.

TODO:

### DIFF
--- a/packages/apollo-codegen-swift/src/codeGeneration.ts
+++ b/packages/apollo-codegen-swift/src/codeGeneration.ts
@@ -245,12 +245,13 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
         );
 
         if (this.context.options.generateOperationIds) {
-          const { operationId } = generateOperationId(
+          const { operationId, sourceWithFragments } = generateOperationId(
             operation,
             fragments,
             fragmentsReferenced
           );
           operation.operationId = operationId;
+          operation.sourceWithFragments = sourceWithFragments;
           this.printNewlineIfNeeded();
           this.printOnNewline(
             swift`public let operationIdentifier: String? = ${SwiftSource.string(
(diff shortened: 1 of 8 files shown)


packages/apollo-codegen-swift/src/codeGeneration.ts
```diff
@@ -267,7 +267,7 @@ export class SwiftAPIGenerator extends SwiftGenerator<CompilerContext> {
           );
           fragmentsReferenced.forEach(fragmentName => {
             this.print(
-              swift`.appending(${this.helpers.structNameForFragmentName(
+              swift`.appending("\\n" + ${this.helpers.structNameForFragmentName(
```
manicmaniac, line 270: @designatednerd Sorry for the additional commit after receiving LGTM.

I just found that the current generated code for Swift doesn't have a newline between query and fragment.
Because of missing newlines, the generated operation ID and sent query's sha256 hash always differs.
After this change, operation ID and sha256 will be same so we can treat Persisted Query more easily.

manicmaniac: The current implementation outputs the following query document as `source` property of `operationIds.json`.
sha256 hash calculated from this source is put into the same `operationIds.json`.

```graphql
query SomeQuery {
    ...SomeFragment
}
fragment SomeFragment {
    id
}
```

However, Apollo iOS sends the following query document (missing newlines).
sha256 hash is calculated from this source at runtime so there's difference between 2 hashes.

```graphql
query SomeQuery {
    ...SomeFragment
}fragment SomeFragment {
    id
}
```

This commit fixes this problem with appending missing newlines.

manicmaniac:
@JakeDawkins Could you take a look at this one when you have time?

Issue #1760, opened 2020-01-15
Enums should not be sorted alphabetically by default


lschomann:
### Problem
When you use the codegen tool to generate your types, enums are by default ordered alphabetically. I  think this might be ok for the most cases, but for example if you have an enum of months like this:

``` JS
enum Month {
    JANUARY
    FEBRUARY
    MARCH
    APRIL
    MAY
    JUNE
    JULY
    AUGUST
    SEPTEMBER
    OCTOBER
    NOVEMBER
    DECEMBER
}
```
And you want to use them in a dropdown or address them by index, it is pretty useless to have it sorted by codegen like this;

``` JS
export enum Month {
  APRIL = "APRIL",
  AUGUST = "AUGUST",
  DECEMBER = "DECEMBER",
  FEBRUARY = "FEBRUARY",
  JULY = "JULY",
  JUNE = "JUNE",
  JANUARY = "JANUARY",
  MARCH = "MARCH",
  MAY = "MAY",
  NOVEMBER = "NOVEMBER",
  OCTOBER = "OCTOBER",
  SEPTEMBER = "SEPTEMBER",
}
```

### Expectation
My suggestion would be to reflect the type definition in the schema by default and implement a possibility to set custom sorting via the config file. So that the outcome will be by default:

``` JS
export enum Month {
  JANUARY = "JANUARY",
  FEBRUARY = "FEBRUARY",
  MARCH = "MARCH",
  APRIL = "APRIL",
  MAY = "MAY",
  JUNE = "JUNE",
  JULY = "JULY",
  AUGUST = "AUGUST",
  SEPTEMBER = "SEPTEMBER",
  OCTOBER = "OCTOBER",
  NOVEMBER = "NOVEMBER",
  DECEMBER = "DECEMBER",
}
```

### Workaround
As a (dirty) workaround for anyone who is also experiencing this issue you can open **_npm/node_modules/apollo/node_modules/apollo-codegen-core/lib/utilities/graphql.js_** from your global npm directory and perform the following replacement:

``` JS
4: function sortEnumValues(values) {
5:    //return values.sort((a, b) => a.value < b.value ? -1 : a.value > b.value ? 1 : 0);
6:    return values;
7: }
```

But keep in mind that **this will only work until you make the next update**, or this is solved ;-)

maxfrigge:
I was surprised to see this too. Having enum keys generated in the same order as they are defined in the schema would seem more intuitive to me personally.

[...]

PR #2043, opened 2020-07-17
Add option to keep schema enum order for code generating

viico:
Hi,

This is a PR related to issue #1760. We want to can keep schema enum order when generating code and not sort them alphabetically. This PR keep sort by default but add an option to don't.

I added a test and I update only for typescript generation. I don't know if it could be interesting for another generation.

TODO:

### DIFF
--- a/packages/apollo-codegen-core/src/compiler/index.ts
+++ b/packages/apollo-codegen-core/src/compiler/index.ts
@@ -56,6 +56,7 @@ export interface CompilerOptions {
   suppressSwiftMultilineStringLiterals?: boolean;
   omitDeprecatedEnumCases?: boolean;
   exposeTypeNodes?: boolean;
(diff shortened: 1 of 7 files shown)


viico:
I fixed unit tests but "ci/circleci: NodeJS 10" still failing, I don't see why.

cdaz5:
this bug has been a thorn in our side for a while now ... anything i can do to help push this merge along? happy to help just LMK

RebMeir:
We also faced the same problem.. anything we can do?

StasKeiss:
Can this be pushed to the main branch? This bug is really a big issue for us.

Shirane85:
This is a major issue. We now need to create a duplicate of the enums in order to support the existing code of the client. This is terrible.

SamuelEls:
Is this completely forgotten feature?

PR #2050, opened 2020-07-24
Add `unionTypes` and `interfaceTypes` to the `json-modern` IR

hwillson:
This commit adds new top level properties called `unionTypes` and `interfaceTypes`, to the `json-modern` exported IR. These properties list all unions and their types, as well as all interfaces and their implementing types.

### DIFF
--- a/packages/apollo-codegen-core/src/__tests__/__snapshots__/jsonOutput.ts.snap
+++ b/packages/apollo-codegen-core/src/__tests__/__snapshots__/jsonOutput.ts.snap
@@ -151,7 +151,9 @@ exports[`JSON output should generate JSON output for a mutation with an enum and
 				}
 			]
 		}
-	]
+	],
+	\\"unionTypes\\": [],
+	\\"interfaceTypes\\": []
 }"
 `;
 
@@ -300,7 +302,17 @@ exports[`JSON output should generate JSON output for a query with a fragment spr
 			]
 		}
 	],
-	\\"typesUsed\\": []
+	\\"typesUsed\\": [],
+	\\"unionTypes\\": [],
+	\\"interfaceTypes\\": [
+		{
+			\\"name\\": \\"Character\\",
+			\\"types\\": [
+				\\"Human\\",
+				\\"Droid\\"
+			]
+		}
+	]
 }"
 `;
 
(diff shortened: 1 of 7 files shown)


packages/apollo-codegen-core/src/compiler/index.ts
```diff
@@ -222,6 +244,8 @@ class Compiler {
     this.options = options;
 
     this.typesUsedSet = new Set();
+    this.unionTypesSet = new Set();
+    this.interfaceTypesMap = new Map();
```
designatednerd, line 248: why are the union types a set and the interface types a map?

packages/apollo-codegen-core/src/__tests__/__snapshots__/jsonOutput.ts.snap
```diff
@@ -658,6 +692,265 @@ exports[`JSON output should generate JSON output for an input object type with d
...
+	],
+	\\"fragments\\": [],
+	\\"typesUsed\\": [],
+	\\"unionTypes\\": [],
+	\\"interfaceTypes\\": [
+		{
+			\\"name\\": \\"Character\\",
+			\\"types\\": [
+				\\"Human\\",
+				\\"Droid\\"
```
designatednerd, line 794: Isn't there also an `Alien` type? Or is this going to only have the possible types that are actually used?

packages/apollo-codegen-core/src/compiler/index.ts
hwillson: @designatednerd There's a slight difference in the details that union types and interface types give us access to. For the union types, we want to list the union type name, and the types that make up that union. `graphql-js` supports this out of the box by giving `GraphQLUnionType`'s a `types` property, which holds the array of types that make up the union. Because of this, we can just store the `GraphQLUnionType`'s in a `Set`. When the `serializeToJSON` function is then called to write out the IR as JSON, we can access the union types we want to show by calling `getTypes()` on each `GraphQLUnionType` itself.

`GraphQLInterfaceType`'s are handled a bit differently though. `graphql-js` doesn't provide a `types` property (or something similar) that gives us easy access to every type using a specific interface. We have to get this ourselves. There was already a helper method setup in the `Compiler` class (in `index.ts`) to do this, called `possibleTypesForType`. Since it's already there in the compiler, I think it makes sense to leverage it, and pass both the interface type, and the types using that interface, through to the `serializeToJSON` function using a `Map`.

I'm happy to handle this differently of course. Thanks!

packages/apollo-codegen-core/src/__tests__/__snapshots__/jsonOutput.ts.snap
hwillson: @designatednerd The schema the tests are using does't have an `Alien` type. You can see it [here](https://github.com/apollographql/apollo-tooling/blob/f2c13e5e9d5a113369c5c236dc038986f4cb0872/__fixtures__/starwars/schema.json).

packages/apollo-codegen-core/src/compiler/index.ts
designatednerd: was mostly curious, thanks for the explanation

PR #2049, opened 2020-07-23
fix(apollo-tools): add undeclared peer dependency graphql

merceyz:
`@apollographql/apollo-tools` is using `graphql` without declaring it as a dependency, this causes issues under strict dependency environments (Yarn PnP) and causes it to rely on hoisting to hopefully place it in a accessible location which is not guaranteed.

Fixes https://github.com/apollographql/apollo-tooling/issues/1680

TODO:

- [x] Update CHANGELOG.md\* with your change (include reference to issue & this PR)
- [x] Make sure all of the significant new logic is covered by tests
- [x] Rebase your changes on master so that they can be merged easily
- [x] Make sure all tests and linter rules pass

PR #2062, opened 2020-08-04
Add missing graphql dependency

oliversalzburg:
Because the dependency is currently not declared, users will see the following error in a yarn2/PnP installation:

```shell
(node:4310) [MODULE_NOT_FOUND] Error: @apollographql/apollo-tools tried to access graphql, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

As the error indicates, the module code does import code from `graphql`, but the module manifest does not declare the dependency. This change fixes that.

### DIFF
--- a/packages/apollo-tools/package.json
+++ b/packages/apollo-tools/package.json
@@ -18,6 +18,9 @@
   "dependencies": {
     "apollo-env": "file:../apollo-env"
   },
+  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


merceyz:
Already done in https://github.com/apollographql/apollo-tooling/pull/2049 but it hasn't been merged yet

Issue #890, opened 2019-01-15
buildServiceDefinition ignores resolvers for unions


checkmatez:
When instantiating apollo-server with _modules_ config option, where typedefs contain union type, __resolveType function never gets called, resulting in error: 

> Abstract type Hi must resolve to an Object type at runtime for field Query.hello with value { there: \"Hey!\" }, received \"undefined\". Either the Hi type should provide a \"resolveType\" function or each possible type should provide an \"isTypeOf\" function.

Here is reproduction on codesandbox:

[![Edit Union resolver bug with modules](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/72mz5lok21)

JakeDawkins:
I don't believe this is related to apollo-tooling. I'd suggest asking questions about this on [spectrum](https://spectrum.chat/apollo) and seeing if it's an issue before filing an issue on apollo-server

checkmatez:
I feel it is related to apollo-tooloing, and probably fixed in https://github.com/apollographql/apollo-tooling/pull/1330. Can't confirm right now, will wait for the next release with this fix.

JakeDawkins:
Ah, I see. Let me know if we need to reopen this

justinmchase:
I just came to report this, the bug still exists in the latest version.

In fact after debugging, the error is right here:
https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-tools/src/buildServiceDefinition.ts#L216

It is trying to see if the type is an object type before supporting adding the `__` functions (which includes __resolveType) but the union is not reported as an object type. It should probably have a special check for union types, or just remove this limitation and always allow it?

I'm happy to make a PR for this as its blocking me right now. I'm trying to upgrade to use `modules` but this is affecting several union types I have.

PR #2067, opened 2020-08-06
Support resolveType for unions

justinmchase:
This fixes issue #890 

Union types are not considered object types therefore when building their schema the function `__resolveType` is not added properly even though it should be supported for union types.

The associated bug was closed but it was not actually fixed previously, this change actually fixes it.


justinmchase:
@hwillson Is there an alternate process to get these legitimate issues looked at? Or how can we get our PR's merged?

PR #2114, opened 2020-09-07
fix(apollo): codegen fails on graphql error

### LINKED ISSUE #1980: Apollo codegen doesn't return non-zero exit code when errors have been found
**Intended outcome:**
Running `apollo client:codegen ...` should exit with a number different than 0, when there are errors.

**Actual outcome:**
Running `apollo client:codegen ...` exits by returning 0, even when there are errors. Such as **Error: ️️There are multiple definitions for the AccountSelectorInputQuery operation. All operations in a project must have unique names. If generating types, only the types for the first definition found will be generated.**. This turns out to be problematic on a pipeline where the next command shouldn't be executed, if there are errors on the previous one.

Create two queries with the same name and run the `apollo client:codegen ...` command. Then run `echo $?` to get its return value. It is 0 but should be different, since there was an error.

**Versions**
2.21.3

royra:
The client:codegen command was exiting with code 0 even if the queries had errors.
Fixes #1980

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Upcoming
 
 - `apollo`
-  - <First `apollo` related entry goes here>
+  - Fixed [#1980](https://github.com/apollographql/apollo-tooling/issues/1980): codegen now returns non-zero code when errors have been found.
 - `apollo-codegen-flow`
   - <First `apollo-codegen-flow` related entry goes here>
 - `apollo-codegen-scala`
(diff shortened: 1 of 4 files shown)


russellporter:
We also keeping running into this issue, would be great to get it fixed! Anything I can do to help @royra?

royra:
I signed the CLA on the same day, not sure why this wasn't reviewed yet.

russellporter:
I contacted Apollo support through their website to request a code review for this. Hopefully it helps 🤞

Ry-DS:
really hope this gets looked at soon; our own CI is incomplete without this. As a workaround I plan to just use `grep` and a bash script to check for keywords such as `failed` to throw a non-zero exit code which definitely isn't ideal.

PR #2116, opened 2020-09-09
fix(vscode-apollo): never automatically open or focus output channel tab

acao:
Users on twitter were reporting that this extension was focusing the output channel on each file save. I had the same issue with `vscode-graphql`, and it seems to have to do with changes in `vscode` itself. 

This is the approach that suited us, but possibly you will want to handle it differently, such as only open on `Error`, or using a different setting based on the environment.

TODO:


klemenoslaj:
Can we please get this one trough? This plugin is the single most annoying thing installed to VSCode due to switching constantly to the output when switching branches etc.

PR #2122, opened 2020-09-18
Remove references to Engine and Graph Manager

zionts:
I generally try to replace with Apollo Studio when referring to the UI,
the Apollo registry when referring to schema storage, and Apollo when
referring to the platform as a whole.

This also updates the URLs the CLI will use to point to
`graphql.api.apollographql.com` instead of `engine-graphql...`

We are not bothering to change the `engine` property of the config
format (`apollo.config.js`) because that option is for internal use.

TODO:


glasser (approved):
Hooray!

PR #2127, opened 2020-09-22
Expose `didExist` in removeServiceAndCompose

michael-watson:
When trying to remove a federated service from Apollo Studio Managed Federation config using `apollo service:delete --serviceName`, the CLI does not return an error if the provided `serviceName` does not exist in the defined graph in Apollo Studio. This PR exposes `didExist` and provides an error when the service being deleted does not exist.


zionts (approved):
lgtm! Please update the changelog

PR #2146, opened 2020-10-23
Fix `removeDirectiveAnnotatedFields` to remove variables from definition

vivster7:
TODO:

# Description
This PR adds functionality to handle the edge case when all usages of a variables are removed by `removeDirectiveAnnotatedFields`. 

Apollo tooling, like `apollo client:extract` will strip specified directives (e.g. `@client`) from GraphQL queries. The function `removeDirectiveAnnotatedFields` is used to remove these fields. Fields that use variables may also be removed. If all usages of a variable are removed, then the variable should be removed from the operation definition because [a definition with an unused variable is invalid](https://spec.graphql.org/June2018/#sec-All-Variables-Used).

The tests include a more comprehensive example.

## Notes
This logic exists in apollo-utilities/src/transform.ts:removeDirectivesFromDocument. Another implementation would be to add a dependency on apollo-utilities and use that function.

### DIFF
--- a/packages/apollo-language-server/src/utilities/graphql.ts
+++ b/packages/apollo-language-server/src/utilities/graphql.ts
@@ -105,8 +105,8 @@ export function removeDirectives<AST extends ASTNode>(
 }
 
 /**
- * Recursively remove orphaned fragment definitions that have their names included in
- * `fragmentNamesEligibleForRemoval`
+ * Recursively remove orphaned fragment definitions and variables that have their names included in
+ * `fragmentNamesEligibleForRemoval` and `variablesEligibleForRemoval`, respectively.
  *
  * We expclitily require the fragments to be listed in `fragmentNamesEligibleForRemoval` so we only strip
  * fragments that were orphaned by an operation, not fragments that started as oprhans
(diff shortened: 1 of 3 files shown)


packages/apollo-language-server/src/utilities/graphql.ts
```diff
@@ -105,18 +105,19 @@ export function removeDirectives<AST extends ASTNode>(
...
+ * `fragmentNamesEligibleForRemoval` and `variablesEligibleForRemoval`, respectively.
  *
  * We expclitily require the fragments to be listed in `fragmentNamesEligibleForRemoval` so we only strip
  * fragments that were orphaned by an operation, not fragments that started as oprhans
  *
  * The `ast` param must extend ASTNode. We use a generic to indicate that this function returns the same type
  * of it's first parameter.
  */
-function removeOrphanedFragmentDefinitions<AST extends ASTNode>(
+function removeOrphanedFragmentDefinitionsAndVariables<AST extends ASTNode>(
```
vivster7, line 117: It might be reasonable to split this function into two. I was trying to avoid running through the AST too many times.

vivster7:
cc @JakeDawkins would appreciate a review or if you could point me in the right direction to find someone to review this.

vivster7:
cc @zionts perhaps you could review or help get 👀  on this PR. Thanks!

PR #2148, opened 2020-10-24
Use async file write and do a content check before actually writing a file

fforres:
Been using apollo extensions and tooling for a while now and on large codebases we've seen big performance issues when running type-generation while VSCode is open.

There's 2 main things I've done.

1. we've noticed it will actually write every file, independent on if the content for it is the same or not. In response, file watcher will see this as the file having changed and re-run processes. For a project with a bunch of files, this means generating the dependency tree practically from scratch, (This essentially means TS/VSCode will hang for a couple minutes, we've been needing to close/open it every time we do some type generation) So I've added some content check before that to avoid writing files unnecesary.
2. Made file writing async with a small concurrency limit to improve on previous point.

(Ideally for point 1 we'd want some content-hashing beforehand to check and create/write generated files based on that, but left that for a different PR)

Caveats: - While this might add some time to the type-generation process itself, since we have to do a content-checking, it will improve DX by not having to kill/restart an editor every time :/ 

BTW/DISCLAIMER: I have some trouble making a build and testing this as packaged solution on my env to gather some numbers. Any help in that area would be great.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9436,9 +9436,9 @@
       }
     },
     "husky": {
-      "version": "5.1.2",
-      "resolved": "https://registry.npmjs.org/husky/-/husky-5.1.2.tgz",
-      "integrity": "sha512-lilaRYeDXcAOj8DuRnN9IxUyEMVbYg9rK7yVNkPB5V4hCvxIUxpMeiv9K2h77CE0HzjCnk1Br0oWe1IghXngDQ==",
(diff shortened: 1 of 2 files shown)


fforres:
cc @JakeDawkins 

I have some trouble making a build and testing this as packaged solution on my env to gather some numbers. Any help in that area would be great.

Also... some tests are failing, and unsure why, I thought TS would've taken care of that 😢

Issue #1179, opened 2019-04-09
Any way to disable globalTypes generation?


tushar-singh:
I'm trying to use codegen in an existing codebase and can't run it on all files as that fails due to numerous issues (imported fragments from other modules, anonymous operations, etc)

What I've instead been doing is running codegen on new code. It works but whenever globalTypes is generated it overwrites previous data – this leaves older operation types without enums

Is there a way to disable globalTypes generation and keep enums local?

dsebastien:
I'm also interested by this, but for different reasons: I'm generating code for a few GraphQL queries using the Apollo CLI codegen feature, but in my case the globalTypes.ts file is empty, so I assume that I simply don't need it.

The issue is that I'm using the Create React App scripts in my project, and it enforces the `isolatedModules` option to `true`. Having empty files is not allowed when that option is enabled.

For now my workaround is to instruct the CLI to generate the globalTypes file outside the compilation scope, but this is a dirt hack that I'd like to avoid :)

Any idea of another (cleaner) solution? I feel like disabling the generation of that empty file would be ideal in my case.

ansoncen:
I simply exclude the `globalTypes.ts` in my `tsconfig.json` and it workes as promised.

macrozone:
i think apollo should not generate the global-types if its empty. it results in errors also for next.js because that forces isolatedModules to be true as well

wonjerry:
+1

macrozone:
damn, just run into that again. Does anyone know a workaround?

Manubi:
any news on this one? as excluding it in my tsconfig.json doesn't seem to work for me.

natterstefan:
Hi, I am also curious if there's a fix for that.

FTR: there seems to be an open PR related to this, but it's still open: https://github.com/apollographql/apollo-tooling/pull/2163.

ryan-rushton:
I got around this issue with the setting `--globalTypesFile src/__generated-globals__/globalTypes.d.ts`.

Manubi:
> I got around this issue with the setting `--globalTypesFile src/__generated-globals__/globalTypes.d.ts`.

You are putting it outside of the compilation scope which works put is not really "clean". A simple `--globalTypesFile false` would be enough put somehow no one seems to care...

ryan-rushton:
@Manubi I totally agree, just adding another workaround for people that need to get on with it in the mean time.

mhmoritz:
[...]

PR #2163, opened 2020-11-13
Do not create globalTypes.ts when it is empty

### LINKED ISSUE #2080: code generator should not generate empty global-types
continuation of https://github.com/apollographql/apollo-tooling/issues/1179#issuecomment-616496523

**Intended outcome:**

`global-types.ts` cannot be empty as this breaks typescript compiler when `--isolatedModules` is set. This option can not be disabled for nextjs users

**Actual outcome:**

if there is nothing to put into `global-types.ts`, apollo generates an empty file. This throws an error on build `Type error: All files must be modules when the '--isolatedModules' flag is provided.`

set `isolatedModules` and run a build

**Versions**

apollo 2.28.3

paulmelnikow:
Fix #2080

Alternative to #2055

TODO:


paulmelnikow:
The Node 8 and 10 builds are failing with TypeErrors that are generated somewhere in the config system. I can't reproduce these locally (even when using Node 8 or 10). Any idea what might be causing these?

```
    TypeError [ERR_INVALID_ARG_TYPE]: The "warning" argument must be one of type Error or string

      at Config.warn (../../node_modules/@oclif/config/lib/config.js:310:21)
      at Promise.all.map (../../node_modules/@oclif/config/lib/config.js:135:26)
```

natterstefan:
Hi 👋,

is there any chance (@maintainers) that we can take a look at this issue?

It would also help if the file just has an empty export `export {}` in that case (would fix https://github.com/apollographql/apollo-tooling/issues/1179).

EyMaddis:
I tried it on my project and it did not work.
Custom scalars will end up in `context.typesUsed`, but won't be in the globals file but instead will be handled by `--customScalarsPrefix`, for me there is a Guid scalar I define as `string`

packages/apollo/src/generate.ts
```diff
@@ -156,25 +156,27 @@ export default function generate(
...
         path.join(
           outputPath,
           `globalTypes.${options.tsFileExtension ||
             TYPESCRIPT_DEFAULT_FILE_EXTENSION}`
         );
-      outFiles[globalSourcePath] = {
-        output: generatedGlobalFile.fileContents
-      };
+
+      if (context.typesUsed.length > 0) {
```
EyMaddis, line 167: `context.typesUsed.filter(t => !isScalarType(t)).length > 0`
to avoid custom scalar types

paulmelnikow: Happy to make this change, though it should probably have a test. What's the behavior you're looking for, exactly?

EyMaddis: Custom scalars like (in schema terms): `scalar MyCustomScalar` are inside this array of typesUsed. However, these do not generate anything in the globals files since the generator does can not know what the data structure is actually. This would still leading to an empty file.

https://www.apollographql.com/docs/apollo-server/schema/custom-scalars/

The test scenario would require a custom scalar type and a query that does not use any enums or other globals, but returns the custom scalar.

macrozone:
its important that this gets merged, it leads to completily pointless problems that could easily be avoided.

packages/apollo/src/generate.ts
paulmelnikow: Thanks. I added the check in 6d42e81.

PR #2170, opened 2020-11-20
Update codegen to work for AbcCTum (Was converting to AbcCtum)

cltnschlosser:
Key is AbcCTA, which get's converted by singularize to structName: AbcCTum (this is already not great, but not the bug causing compilation issues)
The key is used to create the structName.
In this place the property (abcCtum) was being used instead which made: AbcCtum (AbcCTum and AbcCtum don't match, notice the T vs t)

CC @designatednerd

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - `apollo-codegen-scala`
   - <First `apollo-codegen-scala` related entry goes here>
 - `apollo-codegen-swift`
-   - <First `apollo-codegen-swift` related entry goes here>
+   - Fix bug in generated compound type names that lead to structName being inconsistent throughout codegen [#2170](https://github.com/apollographql/apollo-tooling/pull/2170)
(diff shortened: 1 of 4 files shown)


designatednerd:
@cltnschlosser Yeah there's some real messy stuff in here around pluralization but it should at least compile. 

Would you mind adding a test for this so we don't accidentally break it in the future?

cltnschlosser:
Yeah, I can try.

cltnschlosser:
Well this test isn't perfect, but it breaks if you revert my change.

designatednerd (approved):

designatednerd:
I'll ping the team tomorrow about getting a release out with this but the test is helpful. For what it's worth this won't be a problem in the new codegen because you'll be able to specify pluralizations if things aren't coming out the way you want from the default singularizer/pluralizer. [Here's the pull where I've set that up](https://github.com/apollographql/apollo-ios/pull/1552).

PR #2182, opened 2020-12-11
De dup default directives if included in schema already

jsegaran:
If the default directives are provided in the input schema do not duplicate them in the output schema.

### DIFF
--- a/packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
+++ b/packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
@@ -12,6 +12,7 @@ import {
   DirectiveDefinitionNode,
   SchemaDefinitionNode,
   SchemaExtensionNode,
+  specifiedDirectives,
   OperationTypeNode,
   GraphQLObjectType,
   GraphQLEnumType,
@@ -103,6 +104,7 @@ export function buildSchemaFromSDL(
   const schemaDefinitions: SchemaDefinitionNode[] = [];
   const schemaExtensions: SchemaExtensionNode[] = [];
 
+  let directiveNames: Set<string> = new Set<string>();
   for (const definition of documentAST.definitions) {
     if (isTypeDefinitionNode(definition)) {
       const typeName = definition.name.value;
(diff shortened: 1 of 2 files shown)


sachindshinde (approved):
LGTM, nice catch!

packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts
```diff
@@ -129,10 +132,14 @@ export function buildSchemaFromSDL(
     }
   }
 
+  const defaultDirectivesNotInSchema = specifiedDirectives.filter(
+    directive => !directiveNames.has(directive.name)
+  );
```
trevor-scheer, line 137: If the user provides a `@deprecated` directive that doesn't match spec, is it expected behavior that we take the one provided by user?

A part of me wants to suggest filtering theirs from the input document, providing the one from `specifiedDirectives`, and warning against that.

JakeDawkins: I feel like we should _allow_ people to do things that violate spec, but I wouldn't be opposed to warning about it every time

jsegaran: The spec isn't explicit about whether the specified directives may/should/must be applied to the doc. I think giving a warning seems reasonable if the directive definition is different from unexpected but I don't want to give a warning if the directives are present and match expected since that feels like a valid workflow.

trevor-scheer: In the case that the definition is different, should we take the directive specified by the user then? I agree with what you said @jsegaran - warn if different, silent if they're the same.

jsegaran: I think we should still take what the user defined since their server might be overriding or using a different definition so we want to match their source of truth.

jsegaran: Hmm looking into this more it seems like there are differences between graphql-spec and graphql-js.
Namely 15.4.0 allows deprecation of input fields and arguments, [code](https://github.com/graphql/graphql-js/blob/975c150541b990a6b160afea6a8c52d990013f31/src/type/directives.js#L179-L196), which leads to a different directive definition between version 15.4.0 and 15.2.0 in graphql-js. Note the [spec](https://spec.graphql.org/draft/#sec--deprecated) hasn't update the directive locations either.

Also graphql-js includes descriptions for the directives but neither the spec nor our docs include the description (which is more minor).

I think that verifying the directive definitions match the current version of `graphql-js` being used might not be the way best way to go since this would lead to false positives at least for deprecated based on package versions.

Thoughts @trevor-scheer @JakeDawkins?

dariuszkuc:
We are hitting the same issue. By any chance this is going to be merged?

PR #2198, opened 2021-01-06
Swift Codegen: Split up appending of fragments to query document

andybest:
Splits the appending of fragments to the query document into separate
sub expressions to fix an issue where a query that references lots of fragments
causes the Swift type checker to fail with the following message:
```
error: the compiler is unable to type-check this expression in reasonable time; try breaking up the expression into distinct sub-expressions
```

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - `apollo-codegen-scala`
   - <First `apollo-codegen-scala` related entry goes here>
 - `apollo-codegen-swift`
-  - <First `apollo-codegen-scala` related entry goes here>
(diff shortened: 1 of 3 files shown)


martijnwalraven (approved):
Thanks for the fix, this looks good to me. I wish we had tests that referenced more than one fragment, but since we're replacing codegen it may not be worth investing in that.

designatednerd (approved):
Apologies for the delay - I just tested the artifact from this against iOS. It does make some small changes in terms of how we concatenate fragment definitions, but I believe those should have a negligible impact on performance. I'll wait to merge this until Monday for "Seems like a bad weekend to deploy" reasons and then I'll cut a new version, add the new version to the iOS SDK, and cut a new version of that. Whee!

PR #2248, opened 2021-03-24
deps: Stop using fork of `graphql-language-service-*` packages.

abernix:
Originally, we forked the `graphql-language-service-*` packages to avoid a
dependency on `graphql-config` and `graphql-import` which were included as
direct dependencies of those packages (and problematic because of their OWN
dependency trees, at the time).

That is no longer the case in the upstream packages.  Furthermore, the
TypeScript types that we wrote to utilize those packages (which were at the
time written in Flow) are no longer necessary since the project has been
ported to TypeScript.

This is an attempt to switch them out for our own packages.


100terres:
It seems that this PR might fixes a couples of issues like mentioned by @samuela. 

I've also encountered this issue :point_right: #2422. This happens because `apollo-language-server` is using the old `@apollographql/graphql-language-service-interface` deps which some of its own depencencies have a peer dependencie on `graphql` that does not go higher than v14.

`apollo-tooling` seems to support support `v15`, so when installing with `npm`/`yarn` both versions of `graphql` are installed which throws error when running some of the `apollo` cli commands.

Because of that there are some projects I'm working on where we cannot upgrade a bunch of apollo and graphql dependencies :disappointed: 

Let me know if I can help in any way!

Yiin:
Any update on this?

PR #2308, opened 2021-05-06
Deprecate apollo CLI service:* commands

hwillson:
This PR deprecates the `apollo` CLI `service:*` commands by:

- Printing (to `stderr`) deprecation messages for all runtime `apollo service:*` commands (pointing people to the [Rover migration guide](https://go.apollo.dev/t/migration))
- Printing deprecation messages when using `apollo service:*` CLI `help` commands
- Labelling all `service:*` commands as deprecated in the `README`

`service:*` functionality has not been removed since this PR will be rolled out in a patch release. After Rover is officially released, we’ll push out a new major version of `apollo` that fully removes the `service:*` commands.

Reference: https://github.com/apollographql/rover/issues/471

### DIFF
--- a/packages/apollo/README.md
+++ b/packages/apollo/README.md
@@ -23,7 +23,7 @@ $ npm install -g apollo
 $ apollo COMMAND
 running command...
 $ apollo (-v|--version|version)
-apollo/2.33.1 linux-x64 node-v14.16.1
+apollo/2.33.1 darwin-x64 node-v14.15.3
 $ apollo --help [COMMAND]
 USAGE
   $ apollo COMMAND
@@ -642,10 +642,16 @@ _See code: [@oclif/plugin-plugins](https://github.com/oclif/plugin-plugins/blob/
 
 ## `apollo service:check`
 
-Check a service against known operation workloads to find breaking changes
+[DEPRECATED] Check a service against known operation workloads to find breaking changes
 
 ```
-Check a service against known operation workloads to find breaking changes
+[DEPRECATED] Check a service against known operation workloads to find breaking changes
+-----------------------------------------------------------------
+DEPRECATED: This command will be removed from the `apollo` CLI in 
+its next major version. Replacement functionality is available in 
+the new Apollo Rover CLI: https://go.apollo.dev/t/migration
+-----------------------------------------------------------------
+
 
 USAGE
   $ apollo service:check
@@ -716,6 +722,13 @@ OPTIONS
       for more granularity (see: 
       https://en.wikipedia.org/wiki/ISO_8601#Durations)
 
+DESCRIPTION
+  -----------------------------------------------------------------
+  DEPRECATED: This command will be removed from the `apollo` CLI in 
+  its next major version. Replacement functionality is available in 
+  the new Apollo Rover CLI: https://go.apollo.dev/t/migration
+  -----------------------------------------------------------------
+
 ALIASES
   $ apollo schema:check
 ```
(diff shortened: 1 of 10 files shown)


packages/apollo/README.md
```diff
@@ -46,11 +46,11 @@ USAGE
...
-* [`apollo service:check`](#apollo-servicecheck)
-* [`apollo service:delete`](#apollo-servicedelete)
-* [`apollo service:download OUTPUT`](#apollo-servicedownload-output)
-* [`apollo service:list`](#apollo-servicelist)
-* [`apollo service:push`](#apollo-servicepush)
+* [`apollo service:check`](#apollo-servicecheck) [DEPRECATED]
+* [`apollo service:delete`](#apollo-servicedelete) [DEPRECATED]
+* [`apollo service:download OUTPUT`](#apollo-servicedownload-output) [DEPRECATED]
+* [`apollo service:list`](#apollo-servicelist) [DEPRECATED]
+* [`apollo service:push`](#apollo-servicepush) [DEPRECATED]
```
abernix, line 53: I'm afraid this entire `README.md` is generated. 😬 

https://github.com/apollographql/apollo-tooling/blob/4f32ccc3083d9adfb7849a1427d696027b6259af/packages/apollo/package.json#L27

I _suppose_ that could go away? 🤭

packages/apollo/test-utils/setup.js
```diff
@@ -0,0 +1,7 @@
+global.console = {
+  log: jest.fn(),
+  error: jest.fn(),
+  warn: console.warn,
+  info: console.info,
+  debug: console.debug,
+};
```
abernix, line 7: I suspect that I want to say _thanks_ for this? 😄

abernix (approved):
LGTM, though I'm not sure what to do about the `README.md`.  🤔 

It's plausible to believe we could augment that `oclif readme` command with some `apollo service:` searchin' and replacing?

packages/apollo/README.md
hwillson: Haha - I had a feeling it was, but didn't see any references to it being auto-generated 🤦. I'll take a closer look. Thanks @abernix!

hwillson: @abernix I auto-generated the `README.md` and while the formatting isn't quite what I would have wanted, it's not that bad. It duplicates the deprecated messaging a bit extra (which isn't the end of the world), and it removes the deprecated label from the list of commands (since that list is generated from the command names). All in all though, it's not horrible, and the deprecated messaging is definitely visible. I've rebased and modified https://github.com/apollographql/apollo-tooling/pull/2308/commits/602c3a9ad1167d8e7045df456878d4cf13dd1c9c directly if you want to see the changes. Let me know if this is okay - thanks!

lrlna (approved):

packages/apollo/README.md
abernix: Looks great to me.

mindaugasnakrosis:
@hwillson if I previously used apollo service:codegen, and Rover does not have it. What do I do? Is this functionality being removed in the next major release without any replacement in the Rover?

abernix:
@mindaugasnakrosis `apollo service:codegen` is not a command I'm familiar with.  Are you perhaps referring to **`apollo client:codegen`**?

The `client:*` commands are not being deprecated.

mindaugasnakrosis:
npx apollo codegen:generate --localSchemaFile=graphql-schema.json --target=typescript --tagName=gql

Generates typescript from a schema

I get an error:

DEPRECATED: This command will be removed from the apollo CLI in
its next major version. Replacement functionality is available in
the new Apollo Rover CLI: https://go.apollo.dev/t/migration

abernix:
@mindaugasnakrosis I believe that `apollo service:download` would download the introspection result in JSON format.  I believe `rover graph introspect <endpoint>` would get you the same result.  Could you try that?

PR #2335, opened 2021-05-24
fix(apollo-graphql): support complex directive args

### LINKED ISSUE #2162: Schema composition with directives that use complex argument types
Greetings, Apollo team!

Consider the following type definitions that will be used to build an Apollo Federation schema:

```graphql
directive @something(rules: [Rule!]!) on OBJECT | FIELD_DEFINITION

enum Number {
  one
  two
}

input Rule {
  numbers: [Number!]!
}

type Something {
  id: ID!
}

extend type Query {
  getSomething(id: ID!): Something!
    @something(rules: [{ numbers: [one, two] }])
}
```

A single directive called `@something` is declared, that takes in a non-nullable list of GraphQL input object types. This directive can then be used on objects and field definitions to perform some task. An example of a similar directive can be found here: https://docs.amplify.aws/cli/graphql-transformer/auth#auth.

The problem arises during composition of the schema, namely that input type `Rule` is being registered twice, leading to this error: https://github.com/graphql/graphql-js/blob/00eab30fb269b6e2a4e8e61d097d3e249319420e/src/type/schema.js#L223.

I originally posted the issue in `graphql-js` repository: https://github.com/graphql/graphql-js/issues/2845

Upon further investigation, it appears that the actual problem occurs during schema transformation: https://github.com/apollographql/apollo-tooling/blob/dc4fe35134f3108f5c2c0e442dad54a82969398e/packages/apollo-graphql/src/schema/transformSchema.ts#L46

The problem is that `recreateNamedType` creates a new instance of a named type and changes its internal JS object reference. However, `graphql` package insists on that JS references are kept intact during schema composition, since it uses a `Set.has` check to ensure types are not registered twice.

As far as I can see, there is no need to use `recreateNamedType` if type remains the same in https://github.com/apollographql/apollo-tooling/blob/dc4fe35134f3108f5c2c0e442dad54a82969398e/packages/apollo-graphql/src/schema/transformSchema.ts#L46

Perhaps replacing the following code:

```typescript
// Returning `undefined` keeps the old type.
const newType = result || oldType;
typeMap[newType.name] = recreateNamedType(newType);
```

with the following can fix the problem?

```typescript
// Returning `undefined` keeps the old type.
typeMap[oldType.name] = oldType || recreateNamedType(result);
```

At least it seems to work when I change it in my local `node_modules`.

I published an example repository to illustrate the problem: https://github.com/azaxarov/graphql-service-with-complex-directive-parameter

Hope to hear from you guys!

lennyburdette:
This change fixes `transformSchema`, which previously left directives as-in after transforming types. This meant that Input types used in directive arguments would appear duplicated in the schema. It now recreates schema directives and their arguments like it does with named schema types.

Fixes #2162

TODO:

### DIFF
--- a/packages/apollo-graphql/src/schema/__tests__/transformSchema.test.ts
+++ b/packages/apollo-graphql/src/schema/__tests__/transformSchema.test.ts
@@ -1,0 +1,23 @@
+import { buildSchema, printSchema } from "graphql";
+import { transformSchema } from "../transformSchema";
+
+describe("transformSchema", () => {
+  test("no-op transform leaves types untouched", () => {
+    const schema = buildSchema(`#graphql
+    type Query {
+      foo: String @test(baz: { bar: "hello" })
+    }
+
+    input DirectiveArg {
+      bar: String
+    }
+
+    # https://github.com/apollographql/apollo-tooling/issues/2162
+    directive @test(baz: DirectiveArg) on FIELD_DEFINITION
+    `);
+
+    const newSchema = transformSchema(schema, namedType => namedType);
+
+    expect(printSchema(newSchema)).toEqual(printSchema(schema));
+  });
+});
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - `apollo-env`
   - <First `apollo-env` related entry goes here>
 - `apollo-graphql`
-  - <First `apollo-graphql` related entry goes here>
+  - Complex directive arguments don't break `transformSchema` (Fixes #2162)
 - `apollo-language-server`
   - <First `apollo-language-server` related entry goes here>
 - `apollo-tools`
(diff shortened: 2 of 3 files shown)


packages/apollo-graphql/src/schema/__tests__/transformSchema.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    input DirectiveArg {
+      bar: String
+    }
+
+    # https://github.com/apollographql/apollo-tooling/issues/2162
+    directive @test(baz: DirectiveArg) on FIELD_DEFINITION
+    `);
+
+    debugger;
```
trevor-scheer, line 19:
```suggestion
```

trevor-scheer (approved):

Issue #2352, opened 2021-06-24
client:check only prints the filename of the first merged definition


ford-outreach:
**Intended outcome:**

We use the client:check utility in our pull requests to verify that the code in the pull request is compatible with the version of our graphql server that is currently deployed in different environments.
After running it, we check the list of failures against the files changed in the PR and only report a failure on that PR if one of the files changed in the PR was flagged as having an incompatibility.

When an operation that uses fragments has an incompatibility, it only prints the file path of the first included fragment. 

**Actual outcome:**

The file names of all definitions that were merged to form the document are printed.

**How to reproduce the issue:**

Have an operation that uses at least 1 fragment. Introduce a change to the operation that is incompatible with a published service, run the client check. Observe that the fragment file path is printed

**Versions**

2.33.4

PR #2353, opened 2021-06-24
Print all relevant file names on client:check failures

ford-outreach:
This is one possibility to fix https://github.com/apollographql/apollo-tooling/issues/2352
All the tests for this utility are currently commented out - not sure how to proceed there

TODO:

### DIFF
--- a/packages/apollo/src/commands/client/check.ts
+++ b/packages/apollo/src/commands/client/check.ts
@@ -16,12 +16,7 @@ type ValidationResult = graphqlTypes.ValidateOperations_service_validateOperatio
 interface Operation {
   body: string;
   name: string;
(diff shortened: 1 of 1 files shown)


packages/apollo/src/commands/client/check.ts
```diff
@@ -16,12 +16,7 @@ type ValidationResult = graphqlTypes.ValidateOperations_service_validateOperatio
 interface Operation {
   body: string;
   name: string;
-  relativePath: string;
-  locationOffset: LocationOffset;
```
ford-outreach, line 20: locationOffset didn't seem useful, as far as I could tell it was always `1`

Issue #1871, opened 2020-04-06
Unused imports


kazatca:
Hi!
I faced ts checking issue. Generated ts file have unused imports in my case. In my frontend app i use component that can handle union of types (declared by fragment). And my GQL server have query that returns single one of this types. 
I wondered my single type query with fragment on union type as result is valid and working. But generated type file is not valid any more. Ts checker reports about unused imports from globalTypes. They are needed for other types in union, but generated file does not contain it.

PR #2363, opened 2021-07-07
`apollo-codegen-typescript` - Filter out unused imports for Union types

jdmoody:
TODO:

This uses the tests added by @kazatca in https://github.com/apollographql/apollo-tooling/pull/1872 and addresses https://github.com/apollographql/apollo-tooling/issues/1871.

### DIFF
--- a/packages/apollo-codegen-typescript/src/codeGeneration.ts
+++ b/packages/apollo-codegen-typescript/src/codeGeneration.ts
@@ -379,21 +379,53 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
     }
 
     if (selection.selectionSet) {
-      return selection.selectionSet.selections
-        .filter(childSelection => {
+      return (
+        selection.selectionSet.selections
+          .map(childSelection => {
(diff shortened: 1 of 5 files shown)


packages/apollo-codegen-typescript/src/codeGeneration.ts
```diff
@@ -379,10 +379,21 @@ export class TypescriptAPIGenerator extends TypescriptGenerator {
...
+            childSelection.kind === "TypeCondition" &&
+            selection.selectionSet &&
+            !selection.selectionSet.possibleTypes.some(
+              possibleType => possibleType.name === childSelection.type.name
+            )
+          ) {
+            return false;
+          }
+          return true;
+        })
```
jdmoody, line 395: This feels a bit hacky to me and I'm pretty sure it doesn't fix other cases even though it fixes the test case. I'm working on fixing the other cases.

jdmoody: I've updated the PR to fix the case in which there are nested fragments. I added a test case for it.

PR #2385, opened 2021-07-19
Remove the GraphQL VSCode extension

hwillson:
Remove all traces of the GraphQL VSCode extension (excluding historical `CHANGELOG` entries), pointing people towards [its new home](https://github.com/apollographql/vscode-graphql).


jgzuke (approved):
Makes sense to me!

I think `.vscode/launch.json` can also be removed, the task was just being used in the 'Launch VS Code Extension' configuration. The gitignore still references these as well

hwillson:
Thanks very much @jgzuke!

PR #2425, opened 2021-08-22
generate `import type` instead of `import`

jimisaacs:
Hello, this is my attempt to fix codegen to be able to use `"importsNotUsedAsValues": "error"` in tsconfig.json.
It definitely fixes my codebase, which is a pretty gigantic codebase.

I thought of maybe doing a more robust solution, but then I thought, I don't think it needs it. I don't think the codegen actually needs another other type of import other than `import type`.

TODO:


jimisaacs:
I guess `@babel/types` needs to be upgraded, which I didn't see in our codebase because it's a monorepo, and it just decided to use a newer version.

jakubzitny:
+1 for this change. If there is any help needed, please ping me.

Also, if this breaking change seems like a large step forward, an intermediate solution could be to introduce a config setting that would indicate if the generation would create `import type` or just `import`.

the21st:
@hwillson any update on the internal discussions? :) We really need this, we depend on `codegen` in our continuous integration flows. Thanks!

PR #2473, opened 2021-11-09
Align codegen and language server inline fragment __typename use

hwillson:
The `withTypenameFieldAddedWhereNeeded` utility in `apollo-language-server` was updated in #1286 to make sure `__typename`'s are added to inline fragments, when used by commands like `client:push`. The `apollo-codegen-core` package currently uses its own copy of `withTypenameFieldAddedWhereNeeded` however, which was not updated. This means that operations can be registered (in Apollo's operation registry) via `client:push` that don't line up with the operations output by codegen.

This commit adjusts the `apollo-codegen-core` package to use the same `withTypenameFieldAddedWhereNeeded` function as `apollo-language-server` (which is already a dependency of codegen), to make sure the operations output from codegen line up with the operations pushed via `client:push`.

### DIFF
--- a/packages/apollo-codegen-core/src/compiler/index.ts
+++ b/packages/apollo-codegen-core/src/compiler/index.ts
@@ -36,9 +36,10 @@ import {
   getFieldDef,
   valueFromValueNode,
   filePathForNode,
-  withTypenameFieldAddedWhereNeeded,
   isMetaFieldName
 } from "../utilities/graphql";
+
+import { withTypenameFieldAddedWhereNeeded } from "apollo-language-server";
 
 export interface CompilerOptions {
   addTypename?: boolean;
--- a/packages/apollo-language-server/src/graphqlTypes.ts
+++ b/packages/apollo-language-server/src/graphqlTypes.ts
@@ -1702,8 +1702,8 @@ export enum ValidationErrorType {
  * Filter options to exclude by client reference ID, client name, and client version.
  */
 export interface ClientInfoFilter {
-  name?: string | null;
   referenceID?: string | null;
+  name?: string | null;
   version?: string | null;
 }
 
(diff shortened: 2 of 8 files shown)


packages/apollo-language-server/src/graphqlTypes.ts
```diff
@@ -1702,108 +1702,108 @@ export enum ValidationErrorType {
  * Filter options to exclude by client reference ID, client name, and client version.
  */
 export interface ClientInfoFilter {
-  name?: string | null;
   referenceID?: string | null;
+  name?: string | null;
```
trevor-scheer, line 1706: Aside from running codegen, I don't understand why all the shuffling is happening. It looks benign but a bit surprising. Is it possible the newly imported `withTypenameFieldAddedWhereNeeded` has this effect?

trevor-scheer (approved):

packages/apollo-codegen-core/src/compiler/index.ts
```diff
@@ -36,10 +36,11 @@ import {
   getFieldDef,
   valueFromValueNode,
   filePathForNode,
-  withTypenameFieldAddedWhereNeeded,
   isMetaFieldName
 } from "../utilities/graphql";
 
+import { withTypenameFieldAddedWhereNeeded } from "apollo-language-server";
```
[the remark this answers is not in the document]
hwillson: I agree @trevor-scheer - I was surprised to find out codegen already had a dep on it 🤔. Okay, I'll leave this as is for now (knowing that this codegen will be deprecated in the not too distant future). Thanks!

packages/apollo-language-server/src/graphqlTypes.ts
hwillson: @trevor-scheer this was the result of running `client:codegen` which was required due to failing CI tests (caused by changes to our API). I just committed the new resulting `graphqlTypes.ts` file as is, but I agree - it's very annoying that things get re-arranged like this. 😞

hwillson: I just re-ran `npm run client:codegen` again as I was worried property order wasn't being preserved between invocations; luckily the output was the same as the committed file. Maybe we'll leave this as is since it is benign, and it's now lined up if someone runs this command again?

calvincestari (approved):
Thanks for the help in getting this fixed Hugh! 🥇

PR #2545, opened 2022-01-14
support-scalar-jsdoc-annotations

nalchevanidze:
scalar-annotation-tests

TODO:


nalchevanidze:
There are some cases and tests that need to be covered, and that will be addressed. But the question is if you like the idea too? We can also implement a more general approach with a plugin interface where I could put this code in the plugin and combine it with the base implementation?

PR #2590, opened 2022-03-30
Update git-parse to v2

trevor-scheer:


trevor-scheer:
Landed but not released: https://github.com/wayfair/git-parse/pull/55

PR #2611, opened 2022-05-04
Apollo CLI v3

trevor-scheer:
WIP PR for CLI v3 (and majors for other related packages as well).

The primary purpose of these versions is to get these packages on to current versions of node and graphql. I don't intend to be adding any features, this is mostly for project health and compatibility. Particularly the lack of graphql 16 support is a specific pain point for users which I intend to address.

Please read! This project is probably no longer interesting to you. We fully intend to sunset it once we have a viable replacement in place for each of the features.
* If you are using codegen, we highly recommend migrating to `graphql-code-generator`. It's proven to be a superior tool to the codegen offering here. We use it ourselves at Apollo and have had a great experience with it. If codegen is the only reason you're using the `apollo` CLI, consider this the kick you needed to abandon this tool!
* For interactions with Apollo Studio, the `rover` CLI is the successor to this project. @EverlastingBugstopper and others have put a lot of love into that tool and it shows. Please use it instead! I am aware there may be command(s) which haven't made it over there yet, and for that reason this CLI continues to exist.
* I hope to do away with `apollo-env`. If the `fetch`y bits are interesting to you for some reason, it will be helpful for me to hear about that on this PR. The rest is (was?) just node polyfills which shouldn't be applicable anymore.
* ...more to come

TODO:


danilobuerger:
@trevor-scheer what are the alternatives for codegen with swift? Afaik graphql-code-generator doesn't support that use case.

PowerKiKi:
Is this still being worked on ?

trevor-scheer:
@PowerKiKi unfortunately it is not. I know this project is not in a good state, but I don't believe Apollo intends to prioritize any additional work on this repo.

At this point, if you absolutely still need functionality from this repository (which isn't offered by `rover`) I'd recommend forking it and making the necessary changes. As you can tell by this (almost certainly bitrotted) PR, this isn't a trivial task though certainly doable and there's plenty of help from TypeScript. This PR is probably a good jumping off point but I hardly recall the contents and it's quite dated at this point.

Issue #2614, opened 2022-05-07
Seeking Codegen feedback


trevor-scheer:
I'd like to remove codegen completely in an upcoming major release of the CLI. For awhile, [`graphql-code-generator`](https://www.graphql-code-generator.com/) has been a superior tool which offers compatibility with all of Apollo's tools. We highly recommend migrating to it.

For anyone who is still using Apollo's codegen feature, is there anything in particular that's prevented you from making the switch?

gabrielcsapo:
I just used https://www.graphql-code-generator.com/ lots of configuration options and really easy to use. I really enjoyed it.

dotansimha:
Also related: https://github.com/apollographql/apollo-tooling/issues/2053 :)

trevor-scheer:
@dotansimha thanks for that fantastic writeup, I'm sorry we never went forward with any of the suggestions you laid out. It certainly wasn't for a lack of effort on your part - as I'm sure you can tell this repo has gone unloved for some time now. 2 years ago, the idea of wrapping `graphql-code-generator` might've been appealing, but at this point we're actively trying to push users away from the CLI (in most cases, when possible, excluding a few `client:*` commands which aren't yet supported by `rover`).

I'm going to leave #2053 open for now since there's a lot of great information there for people looking to migrate, in hopes that it's slightly more discoverable open instead of closed. You're more than welcome to close it if you choose, I'll probably close whenever I finally get to cut the last major for these packages.

markhalonen:
I considered switching to [graphql-code-generator](https://www.graphql-code-generator.com/) but I stayed with Apollo's version because it generates intermediate types, which I use extensively. The solution proposed in https://github.com/apollographql/apollo-tooling/issues/2053#issue-666281081 is to use Fragments, which makes sense if the Fragment is used multiple times, but often I want to use the intermediate type(s) on regular Queries with no Fragments. 

Having to insert Fragments into my Query just to get the intermediate types to generate is not ideal, it makes the Query more complex, and I have to name the Fragment.

n1ru4l:
@markhalonen  an you share your use-case for accessing indermediate types? We sis this in codegen V1, but decided to deop it as it bloated auto-complete suggestions. As an alternative you can use [lookup types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html).

dotansimha:
[...]

PR #2619, opened 2022-05-20
Deprecate `client:codegen` command

trevor-scheer:
Related: https://github.com/apollographql/apollo-tooling/issues/2614

This change officially deprecates the codegen utility in the Apollo CLI. The recommendation going forward is to use the `graphql-code-generator` utility.

For iOS (for which there is no existing replacement): patch fixes will still be entertained and released on the v2.x line, but no further development will be done unless it's necessary. Apollo iOS intends to release their own codegen in the relatively near future.

While we do encourage moving away from the CLI as quickly as possible, I'm sympathetic to the fact that a change might not be feasible right away. I've included the `--suppressDeprecationWarning` flag in order to silence the console output when running the `client:codegen` command.

I haven't added tests for the printing of this command, but I did test locally to ensure the notice is printed (and is also suppressible with the flag).

### DIFF
--- a/.changeset/chilled-yaks-refuse.md
+++ b/.changeset/chilled-yaks-refuse.md
@@ -1,0 +1,5 @@
+---
+"apollo": minor
+---
+
+Deprecate the client:codegen command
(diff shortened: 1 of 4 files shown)


hwillson (approved):
Looks great @trevor-scheer - thanks!

PR #2664, opened 2022-10-11
Add sourceName to anonymous operation error output

### LINKED ISSUE #184: Apollo does not support anonymous operations
version: apollo-codegen 0.15.2

i execute query in graphiql
`query { getAllCustomers {id} }`

i gets data.getAllCustomers[ids...]

````
{
  "data": {
    "getAllCustomers": [
      {
        "id": "34"
````
then i want to execute query under apollo-client/angular-apollo

so i continue with codegen

- apollo-codegen download-schema http://localhost:3000/graphql --output schema.json
- apollo-codegen generate src/**/*.gql --schema schema.json --target typescript --output src/app/graphql/schema.ts

[schema/json/js](https://gist.github.com/agborkowski/b1c4f05b32fab354933f917625745142)

i have only one .gql file with the same query like i executed in graphiql:
`query { getAllCustomers {id} }`

but schema.ts wont genrate with error

```
.../main/customers/_graphql/getCustomers.gql: Apollo does not support anonymous operations
error: Validation of GraphQL query document failed
````
# Update

if found problem query withink works in graphql-tag and gprahiql wont generate type script to solve above problem is add

````
query SOMESTRING { getAllCustomers {id name symbol} }
````

then apolo codegen will generate 

````
export type SOMESTINGQuery = {
  getAllCustomers:  Array< {
`````

kdawgwilk:
Should help resolve #184 issues that many run into when introducing this tooling to their app for the first time

### DIFF
--- a/packages/apollo-language-server/src/project/client.ts
+++ b/packages/apollo-language-server/src/project/client.ts
@@ -419,8 +419,9 @@ export class GraphQLClientProject extends GraphQLProject {
       for (const definition of document.ast.definitions) {
         if (definition.kind === Kind.OPERATION_DEFINITION) {
           if (!definition.name) {
+            sourceName = document.source.name
             throw new GraphQLError(
-              "Apollo does not support anonymous operations",
+              `Apollo does not support anonymous operations: ${sourceName}`,
               [definition]
             );
           }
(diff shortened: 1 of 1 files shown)


kdawgwilk:
I guess this repo is deprecated and won't be accepting any more PRs

PR #2681, opened 2023-04-21
Bug bounty testing... Please do not approve. rmeroq

malshinim:
Bug bounty testing... Please do not approve. rmeroq


peakematt:
@malshinim Thanks for the PR! From what we can tell, it looks like this PR was checking for misconfigurations in our build pipelines that would allow for exfiltration of sensitive info. Based on what we can see, we believe our configuration to be safe. However, if you did find something, please let us know at security@apollographql.com 

Thank you!

